PR #10, opened 2017-05-29
Stage 015 (CDS Documents)

mgolosova:
* New command line parameters:
```
USAGE
  ./getCDSPapers.py <options> [file]

ARGUMENTS
  file                    Input file name (default: Input/list_of_papers.json)

OPTIONS
  -o, --output-dir DIR    Output directory name
```

### DIFF
--- a/Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
+++ b/Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
@@ -21,6 +21,7 @@ output metadata:
 import json
 from urlparse import urlparse
 import sys, getopt
+import os
 
 sys.path.append("../")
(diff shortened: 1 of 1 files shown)


mgolosova:
If accepted, the #11 is supposed to be merged earlier as the first commit (20f586a) is from that branch. I believe it must disappear from the request when #11 is merged.

maria-grigorieva:
Commit 20f586a is still here. Probably it should be removed from branch so as it to be merged correctly?

mgolosova:
Thank you, now removed (by rebasing on 'dataflow-with-...').
I think it might be merged successfully even with that excessive commits (due to mergeinfo in the destination branch), but definitely wouldn't look nice.

Seems like it happened because the branch was based on the '*-fix' branch, and thus all the *-fix commits remained in the request. Now it is based on 'dataflow-with-...', so the extra commits are gone.

PR #12, opened 2017-05-31
pyDKB fix

mgolosova:
1. Code formatting changed according to PEP 8

1. Bug fix:
* broken dependency in 'messages' module (Message() function uses sys)
* TTLMessage parse() function (addressing private variable)
* processor stage output() implementation
* TTLProcessor output() function: fall back to parent's output()


mgolosova:
Code formatting and bug fixes would possibly look better in different pull requests, but as there are no changes in pyDKB overlapping with this request, it can be merged with rebase+merge, and thus, I believe, can go in one request.

Still can be splitted if the mix looks ugly :)

PR #15, opened 2017-06-02
pyDKB: --input-dir parameter enhancing

mgolosova:
Add possibility to specify only the directory without listing all the files.

Now if --input-dir is specified but no input FILE presented, the Stage will try to get list of files in the directory and use them as input.
If the directory does not exist, error message appears.
If the directory is empty, nothing happens.

### DIFF
--- a/Utils/Dataflow/pyDKB/common/__init__.py
+++ b/Utils/Dataflow/pyDKB/common/__init__.py
@@ -2,5 +2,7 @@
 Common modules.
 """
 
+from exceptions import *
+import hdfs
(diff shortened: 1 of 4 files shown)


mgolosova:
This pull request is based on b18ab7b (which is also presented in pyDKB-fix request), so before merging this one we need to do something with pyDKB-fix

mgolosova:
There were conflicts after merging #12.
Now they are resolved, and b18ab7b is moved from the history.

Please, review.

PR #18, opened 2017-06-15
Installation: script for elasticsearch+kibana+nginx

IATkachenko:
This script will install:

1. elasticsearch, because we need it;
1. kibana, because it is the simplest way to see and analyze data stored in elasticsearch;
1. nginx, because we need access to elasticsearch and kibana from outside.

Installation script will install all necessary rpms, except java.
Elasticsearch and kibana configured for local access only, without passwords. 
Nginx used in proxy mode to provide password protected access to elasticsearch and kibana from outside.


IATkachenko:
Updated based on @mgolosova comments.

PR #20, opened 2017-06-17
pyDKB refinement

mgolosova:
Some improvements of pyDKB:

1. Bug fixes:
* excessive error message when Processing Stage run without command line parameters
* undefined instance variable of JSON2TTL Processor Stage due to the wrong parent`s __init__ call

2. Refactoring&Codestyle:
* move HDFS-related functionality (get file from HDFS) to the corresponding module (common.hdfs)
* get rid of "type" variable and submodule name (reserved for Python built-in class)
* use os.path instead of working with filenames and paths directly

3. Typo fix


mgolosova:
Please review the changes:

1. Commit "`pyDKB: improve codestyle: use os.path to work with paths.`" is rewritten
    I found a bug that must have been fixed within this commit but somehow it hadn`t -- I missed one place where os.path should replace direct transformation of path as a string.

1. Commits 5e23f76..74e6f8e are added
    They also are about pyDKB refinement and it`s logical to put them to the same PR.

mgolosova:
Sorry for inconvenience, could it be reviewed again, please? 
Found out that some debug prints are left in code. Fixed -- and rewrote the history again :(

mgolosova:
Merged without rebase to keep the commits "grouped" in the graph representation of the commit history (see `git log --graph`)

PR #21, opened 2017-06-20
pyDKB: Processor Stage output full implementation

mgolosova:
The implementation of output logic according to the command line parameters was implemented.

**Major changes**
1. AbstractProcessorStage.process() method changed:
- has became a static method with two parameters: `stage, message`
  ...to allow usage of the Processor Stage output() method from within the process() implementation
- must return True or False
  ...to indicate if the processing was successful or not. If the
  processing has failed, all the output messages are dropped

2. AbstractProcessorStage.stop() method added
- must be called after run() to finalize internal processes

3. Test stage `Dataflow/test/pyDKB` added
- to demonstrate how to use Processor Stage and test its functionality

**The logic**

Output messages are buffered in `__output_buffer` parameter, and pushed forward only when the input message processing finished successfully.

Output messages separator and end-of-process (EOP) marker for (s)treaming mode are hardcoded for now (`\n` and `\0` respectively), _but should be made confugurable later_. The separator for input messages is the same as for output messages for now _(but must be configured independently in the future)_.

In (f)ile mode output goes to files with the same basename as the original one but with another extention (it is determined by the output message type). The files will appear in the same directory, as the original ones, or in the directory specified with `--output-dir`.

### DIFF
--- a/Utils/Dataflow/pyDKB/common/hdfs.py
+++ b/Utils/Dataflow/pyDKB/common/hdfs.py
@@ -28,6 +28,20 @@ def check_stderr(proc, timeout=None):
             proc.wait()
     return proc.poll()
 
+def putfile(fname, dest):
+    """ Upload file to HDFS. """
+    cmd = ["hadoop", "fs", "-put", fname, dest]
+    try:
+        proc = subprocess.Popen(cmd,
+                                stdin =subprocess.PIPE,
+                                stderr=subprocess.PIPE,
+                                stdout=DEVNULL)
+        if check_stderr(proc):
+            raise(subprocess.CalledProcessError(proc.returncode, cmd))
+    except (subprocess.CalledProcessError, OSError, HDFSException), err:
+        raise RuntimeError("(ERROR) Failed to put file to HDFS: %s\n"
+                           "Error message: %s\n" % (fname, err))
+
 def getfile(fname):
     """ Download file from HDFS.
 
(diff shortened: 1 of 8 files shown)


mgolosova:
The base branch is specified as `pyDKB-refinement` (#20) as the changes are based on the changes made in that branch.
As soon as #20 is merged, the base branch is to be changed to the `pyDKB`.

Hope it`ll work ;)

maria-grigorieva:
in README: 
staticmethod -> static method

Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
[254]  -- double (( )) -> single () ?  
+                failures.append((p, e))

mgolosova:
Maria, thank you.

> in README:
> staticmethod -> static method

Right you are. It\`s a keyword in Python so I left it unsplitted while actually shouldn\`t.
PR comment updated, commit log -- on its way.

> Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
> [254] -- double (( )) -> single () ?
> `         failures.append((p, e))`

No, it\`s meant to be a list of tuples: [(process1, error1), (process2, error2), ...].
Here I\`m trying to `close()` every process that must be closed in spite of any possible error, but keep those errors somewhere to output them later. Wasn\`t sure if I should throw the exception (then which one?) or just print the message, but for now decided just to print. We\`ll see later if I was wrong :)

mgolosova:
As the #20 was merged without rebase, just changing the base branch to pyDKB wasn`t enough. So I rebased it on the current pyDKB version and force pushed the new commit history in the PR.

The code wasn`t changed, please review.

mgolosova:
Modified test stage README according to the new output behaviour (last two commits rewritten)

mgolosova:
Yes, it`s a time, or I will never finish it.
Thank you!

PR #22, opened 2017-06-21
pyDKB v.0.1

mgolosova:
**New functionality:**
* Abstract Processor Stage can output data according to the command line parameters
* data can be output() not only at the end of process(), but during the whole process()
* --input-dir can be used without explicit FILE specification

**Bug fixes:**
* broken dependency
* direct addressing of private variable
* wrong iter() call (without `sentinel` for callable object)
* undefined instance variable due to wrong parent`s __init__() call

**Other changes (structural, behaviour):**
* common.hdfs submodule added
* AbstractProcessorStage.stop() method added
* local files are no longer removed when downloading from HDFS (exception is raised)


Utils/Dataflow/pyDKB/dataflow/stage/processors.py
```diff
@@ -11,61 +11,65 @@
...
+
+    def input(self):
+        """Override: Falls back to JSONProcessorStage.input"""
+        # Pick the method of the first parent
+        return super(JSON2TTLProcessorStage, self).input()
+
+    def output(self, message):
+        """Override: Falls back to TTLProcessorStage.output"""
+        # Skip first parent, pick the second
+        super(JSONProcessorStage, self).output(message)
```
mgolosova: Python\`s super() construction takes next ancestor after specified, so here we skip JSONProcessor\`s `output()` and take the one of the next ancestor (namely, TTLProcessor).

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -0,0 +1,118 @@
...
+        line = line.split(None, 7)
+        if len(line) != 8:
+            continue
+
+        # We need to return only the name of the file or subdir
+        line[7] = os.path.basename(line[7])
+        if line[0][0] == 'd':
+            subdirs.append(line[7])
+        elif line[0][0] == '-':
+            files.append(line[7])
```
maria-grigorieva: The lines number 7 is constant? It'll be better to set it as the constant, explaining which field in 'ls' output does it returns.

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -0,0 +1,118 @@
...
+        if check_stderr(proc):
+            raise(subprocess.CalledProcessError(proc.returncode, cmd))
+    except (subprocess.CalledProcessError, OSError, HDFSException), err:
+        raise RuntimeError("(ERROR) Failed to put file to HDFS: %s\n"
+                           "Error message: %s\n" % (fname, err))
+
+def getfile(fname):
+    """ Download file from HDFS.
+
+    Check if there already is a local version of the file and remove it.
```
maria-grigorieva: It should be done manually, or it should be implemented or it's been already implemented? 
getfile() rewrite the existing file, doesn't it?

mgolosova: Got to fix the comment, thank you.
First it was like "Function goal: download file; check the local and remove it."
But after 37c71c9 (`pyDKB/hdfs: prevent getfile() from removing local file
`) it no longer removes the local file.

PR #25, opened 2017-06-23
Stage 015 (CDS documents) with pyDKB

mgolosova:
Stage 015 now works through pyDKB Stage interface.

---
The PR waits the completion of #22, and will be rebased on `dataflow-with-document-content` branch after #22 is merged.

### DIFF
--- a/Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
+++ b/Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
@@ -1,10 +1,8 @@
 #!/usr/bin/env python
 """
-author: Maria Grigorieva
-maria.grigorieva@cern.ch
-
-refactored by: Golosova Marina
-golosova.marina@gmail.com
+authors:
+ Maria Grigorieva (maria.grigorieva@cern.ch)
+ Marina Golosova (golosova.marina@gmail.com)
 
 Execute on local machine with installed invenio-client;
 invenio-client also requires phantomjs (>=1.9.8 due to default SSL protocol)
@@ -20,46 +18,19 @@ output metadata:
 
 import json
 from urlparse import urlparse
-import sys, getopt
+import sys
 import os
-
-sys.path.append("../")
-from pyDKB.dataflow import CDSInvenioConnector, KerberizedCDSInvenioConnector
-from pyDKB.dataflow import dkbID, dataType
 
 import warnings
 from requests.packages.urllib3.exceptions import InsecurePlatformWarning
 
+sys.path.append("../")
+from pyDKB.dataflow import CDSInvenioConnector, KerberizedCDSInvenioConnector
+from pyDKB.dataflow import dkbID, dataType
+from pyDKB.dataflow.stage import JSONProcessorStage
+from pyDKB.dataflow import DataflowException
 
 counter = 0
-
-def usage():
-    """ Output usage string. """
-    msg = """
-USAGE
-  ./getCDSPapers.py <options> [file]
-
-ARGUMENTS
-  file                      Input file name (default: Input/list_of_papers.json)
-
-OPTIONS
-  -l, --login       LOGIN   CERN account login
-  -p, --password    PASSWD  CERN account password
-  -k, --kerberos            Use kerberos authorization
-
-  -P, --pretty              Pretty print output
-
-  -m, --mode        MODE    operating mode:
-                            (f)ile  -- default mode: read from file,
-                                       output to files
-                            (s)tream -- stream mode: read from STDIN,
-                                        output to STDOUT
-
-  -o, --output-dir DIR      Output directory name
-
-  -h, --help                Show this message and exit
-"""
-    sys.stderr.write(msg)
 
 
 def collection_verification(collection):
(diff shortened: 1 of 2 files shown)


Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
```diff
@@ -34,355 +34,393 @@
...
+                                 "(WARN) Will be taken the first of the"
+                                 " list\n" % str(len(res)))
+            r = None
+            for item in res:
+                if collection_verification(item.get("collection")):
+                    r = item
+                    break
+            res = r
+        if type(res) == dict:
+            res['glance_id'] = paper_info["id"]
```
maria-grigorieva: Line 142: Change quotes from single to double

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
```diff
@@ -34,355 +34,393 @@
...
-                                                         % (list, type(sdGLANCEdata)))
+    """ Combine input and found metadata; generate and add dkbID.
+
+    Parameters:
+      GLANCEdata -- main input JSON data
+      ppCDSdata  -- paper information from CDS
+      sdCDSdata  -- supporting documents information from CDS:
+                    { $sdGlanceID : [ ... ], ... }
+    """
+    if type(ppCDSdata) != type(GLANCEdata) != type(sdCDSdata) != dict:
```
maria-grigorieva: Probably it will be better to substitute type() with isinstance() as here we are checking non built-in types. 

`if not (isinstance(ppCDSdata,dict) and isinstance(GLANCEdata,dict) and isinstance(sdCDSdata,dict)):`

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
```diff
@@ -34,355 +34,393 @@
...
+    ppGLANCEdata = GLANCEdata.copy()
+    if ppGLANCEdata.get("supporting_notes"):
+        sdGLANCEdata = ppGLANCEdata.pop("supporting_notes")
+    else:
+        sdGLANCEdata = []
+
+    result["GLANCE"] = ppGLANCEdata
+    result["CDS"] = ppCDSdata
+    result["dkbID"] = dkbID(result, dataType.DOCUMENT)
+
```
maria-grigorieva: We should unify the method of getting key value from dictionary. 
*.get(key) will be better?

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
```diff
@@ -34,355 +34,393 @@
...
+    sd_results = []
+    sd_result = {}
+    result["supporting_notes"] = sd_result
+
+    if type(sdGLANCEdata) != list:
```
maria-grigorieva: type -> isinstance

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
```diff
@@ -34,355 +34,393 @@
...
+    for glance_ind in range(len(sdGLANCEdata)):
+        glance_item = sdGLANCEdata[glance_ind]
+        sd_result["GLANCE"] = glance_item
+
+        cds_item = sdCDSdata.get(glance_item.get("id"))
+        if not cds_item:
+            sys.stderr.write("(WARN) No CDS data for GLANCE id: %s\n"
+                             % glance_item.get("id"))
+            continue
+        if type(cds_item) != dict:
```
maria-grigorieva: type -> isinstance

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
maria-grigorieva: Single quotes or double quotes? We should unify

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
mgolosova: But `dict` seems to _be_ the built-in type: https://docs.python.org/2/library/stdtypes.html#dict
We don\`t want to check the inheritance here, so why not to use `type()` ?

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
mgolosova: Here\`s the values assignment -- `get()` is to safely read data from the dict, not to put it there.

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
maria-grigorieva: Sure, it's my mistake.

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
mgolosova: The `list` is also a built-in type, and we don\`t care about the inheritance here. I don\`t actually see why to use `isinstance()`, if the `list` is exactely what I expect to see there (except that we could agree to unify the way of typechecking):
https://stackoverflow.com/questions/21894575/isinstancefoo-bar-vs-typefoo-is-bar

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
mgolosova: Ok, thank you.

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
maria-grigorieva: Totally agree.

Utils/Dataflow/015_CDSDocuments/getCDSDocuments.py
mgolosova: By the way, it\`s not only about this line, I always mix up single and double quotes.
Shall we use only double, or only single, or double for everything but a single character? Or single for identifiers/code values and double for strings (like error messages etc?

maria-grigorieva: I've just read PEP8 https://www.python.org/dev/peps/pep-0008/#string-quotes. It doesn't provide recommendations. So, I think we can use double-quotes for natural language strings and signle-quotes for technical strings.

mgolosova:
Changed base to #28, as the commits to be merged after the #28.

PR #26, opened 2017-06-26
Stage 055_documents2TTL with pyDKB integration

maria-grigorieva:
- implemented process() converting JSON to TTL message
- removed EOPmaker from main arguments
- removed unused function write_ttl2file()
- in main removed condition of choosing the output,
 as it's been implemented in pyDKB

 documents2ttl script was checked with the following modes:

 1) <filename>.json -m f -s f -i /input_dir -d f -o /output_dir
 2) -s f -i /input_dir -d f -o /output_dir -m f
 3) cat <filename>.json | python documents2ttl.py -s s -d s -m s | less

 in these 3 cases it works correctly.


Utils/Dataflow/055_documents2TTL/documents2ttl.py
```diff
@@ -399,86 +401,66 @@ def fix_list_values(list_vals):
...
+        for note in data.get('supporting_notes'):
+            note_id = note.get('dkbID')
+            note_iri = get_document_iri(note_id)
+            doc_ttl += '{noteIRI} a <{ontology}#SupportingDocument> .\n' \
+                .format(noteIRI=note_iri, ontology=ONTOLOGY)
+            doc_ttl += document_glance(note.get('GLANCE'), note_iri, NOTE_GLANCE_ATTRS)
+            doc_ttl += document_cds(note.get('CDS'), note_iri, NOTE_CDS_ATTRS)
+
+    doc_ttl += documents_links(data)
+    ttl_message = messages.TTLMessage(doc_ttl)
```
mgolosova: It was supposed that single TTLMessage becomes a single TTL statement.

For now it makes no difference if we put there a number of statements, but later can become significant. For example, when we
* choose to use another message delimiter (not the `\n`) to allow multi-line messages
* add rules to check the TTLMessage content (so that not every string can become a TTLMessage, but only a triple-like one)

It might be better to do this way now, to avoid getting back to this later.

Utils/Dataflow/055_documents2TTL/documents2ttl.py
```diff
@@ -399,86 +401,66 @@ def fix_list_values(list_vals):
...
+    output_message = pyDKB.dataflow.Message(pyDKB.dataflow.messageType.TTL)(ttl_message.content())
```
mgolosova: `pyDKB.dataflow.Message(messageType)` is a function that returns a corresponding class `*Message` (here -- `TTLMessage`), so this can be avoided: `ttl_message` and `output_message` are completely the same.

Utils/Dataflow/055_documents2TTL/documents2ttl.py
```diff
@@ -399,86 +401,66 @@ def fix_list_values(list_vals):
...
-                output = paper_id + '.ttl'
-            write_ttl2file(output, doc_ttl)
-        else:
-            sys.stdout.write(doc_ttl + processor.ARGS.EOPmarker)
-            sys.stdout.flush()
+    stage.parse_args(argv)
+    define_globals(stage.ARGS)
+    stage.process = process
+    stage.run()
+    stage.stop()
```
mgolosova: Please check here: https://github.com/PanDAWMS/dkb/blob/pyDKB-refinement/Utils/Dataflow/test/pyDKB/json2TTL.py#L28

We want to run stop() even if the run() throws an exception, so try...finally is vital here.
Sorry I`ve misled everyone with example currently existing in pyDKB branch :(

PR #27, opened 2017-06-27
Stage 054 with pyDKB

anastasiakaida:
The script were checked with the last version of pyDKB from pyDKB branch.
File samples (TTL and JSON) were added.


anastasiakaida:
054 is fixed according to all comments.

PR #33, opened 2017-07-10
PDF Analyzer: pyDKB integration

mgolosova:
Original vresion of the PR #32

flow.py contains process() function which allows PDFAnalyzer to be
called as a pyDKB module. Removed obsolete kafkamode.py, its
functions are now performed either by flow.py(PDF processing in non
interactive mode) or pyDKB(input and output handling).
It should be noticed that new workflow mode implemented in flow.py
accepts JSON files instead of strings and passes dkbID from input
to output (see previous stage output).

---

PR is created on behalf of @Evildoor


mgolosova:
Closed due to the need of improvements in absence of the author (see #32)

mgolosova:
The request was reopened and closed back to update commit history.
Commits  c980c4f...c563acd show the (more or less) original history of the changes as it might be needed by @Evildoor, while the #33 has some of the changes encapsulated into the first commit for the "official" (merge) version.

PR #39, opened 2017-08-03
Stage 030 PEP-8

Evildoor:
PDF Analyzer update to meet the PEP-8 standard requirements.


Utils/Dataflow/030_PDFAnalyzer/pdfwork.py
```diff
@@ -96,49 +114,15 @@ def mine_text(infname, page_numbers = False, outtype = "text", rotated_pages = [
...
+    #            outf.writelines(lines)
             n += 1
 #    inf.close()
     device.close()
     tmp.close()
     return [n - 1, rotated_pages]
 
 if __name__ == "__main__":
-    [pages, rotated_pages] = mine_text("C:/Work/papers_analysis/ATL-COM-PHYS-2014-1357.pdf", [201], "xml")
+    fname = "C:/Work/papers_analysis/ATL-COM-PHYS-2014-1357.pdf"
```
maria-grigorieva: Try to avoid absolute paths to files from local machines.

Utils/Dataflow/030_PDFAnalyzer/pdfwork.py
```diff
@@ -96,49 +114,15 @@ def mine_text(infname, page_numbers = False, outtype = "text", rotated_pages = [
...
+    [pages, rotated_pages] = mine_text(fname, [201], "xml")
```
maria-grigorieva: What is [201] ?

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
```diff
@@ -380,7 +446,8 @@ def analyze_page(text):
 ##            print l.text
 
 if __name__ == "__main__":
-    f = open("C:/Work/papers_analysis/manager/papers/ATL-COM-PHYS-2015-170/xml/9.xml", "r")
+    fname = "papers/ATL-COM-PHYS-2015-170/xml/9.xml"
```
maria-grigorieva: Please, remove path to local files

maria-grigorieva: Replace test from main() fuction

Utils/Dataflow/030_PDFAnalyzer/pdfwork.py
Evildoor: Test will be separated from main.

Utils/Dataflow/030_PDFAnalyzer/pdfwork.py
Evildoor: Page number which i was performing tests on, IIRC.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
Evildoor: Test will be separated from main.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
```diff
@@ -133,7 +138,8 @@ def row_centery(row):
     return c
 
 class Table:
-    re_month = re.compile("(january|february|march|april|may|june|july|august|september|october|november|december)")
+    re_month = re.compile("(january|february|march|april|may|june|july|august\
+                          |september|october|november|december)")
```
mgolosova: 1. Those `^M` in the end of lines -- can they be removed? I believe it must be `\r` from `\r\n` (CRNL), the Windows-style new line delimiter. They might not affect the program itself, but are a bit distracting while reading the code or commit diff.

2. Does the search works properly after this change?
In my it won\`t find the "august":
```
Python 2.7.12 (default, Nov 19 2016, 06:48:10) 
[GCC 5.4.0 20160609] on linux2
>>> import re
>>> m1 = re.compile("(january|february|march|april|may|june|july|august|september|october|november|december)")
>>> m2 = re.compile("(january|february|march|april|may|june|july|august\
...                 |september|october|november|december)")
>>> m2.search("Some of my text with august mentioned")
>>> m.search("Some of my text with august mentioned")
<_sre.SRE_Match object at 0x7f4fe0bd65d0>
```
You can use this variant:
```
>>> m3 = re.compile("(january|february|march|april|may|june|july|august"
...                 "|september|october|november|december)")
>>> m3.search("Some of my text with august mentioned")
<_sre.SRE_Match object at 0x7f4fe0bd6648>
```
Strings followed each other are simply concatenated in Python.

mgolosova:
And one more comment to the e0aff0a -- about spaces around `+` (string concatenation etc) within brackets.
PEP-8 states only about spaces around `=`:
```
Don't use spaces around the = sign when used to indicate a keyword argument
or a default parameter value.
```

It says nothing specific about `+` in concatenation and within the brackets, yet shows following example as a good one:
```
income = (gross_wages
          + taxable_interest
          + (dividends - qualified_dividends)
          - ira_deduction
          - student_loan_interest)
```
It is not within the brackets surrounding the function arguments, though... but as for me, spaces make code much more readable.
I think it is OK to leave spaces where they were, unless you do prefer the variant without them.

Thank you!

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -30,7 +30,8 @@ def load_config(default_cfg):
             loaded_cfg = json.load(f)
     except Exception as e:
         sys.stderr.write("Exception while loading config: %s\n"%e)
-        sys.stderr.write("No config file loaded, using default values\n")
+        sys.stderr.write("No config file loaded, using default values\
```
mgolosova: `^M` again, and it seems to be added into the log message.
Use 
```
sys.stderr.write("No config file loaded, using default values"
                 "\n")
```
instead, or else you will have the two-string message: with text and with spaces.
The newline within quotes is just the same as `\n`.

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -123,76 +156,75 @@ def find(self, text, intervals, datasets):
...
-                                           [a-zA-Z\d\-_\.:!]+   #
-                                           \n*\.\n*             # Field separator
-                                           [t0proge]+\d\d_v?\d\d# version.
-                                           (\n*_\n*[a-z]\d+)*   # Possible AMITag or several.
+group = DatasetCategory("group", r"""
+group              # Indicates group dataset.
+\n*\.\n*           # Field separator
+[a-zA-Z\d\-:]+     # Group name. Examples: phys-higgs, phys-beauty.
+(\n*[._]\n*[a-zA-Z\d\-:!]+)+
+                                     """)
```
mgolosova: Seems like you have the trailing spaces here and in line [183](https://github.com/PanDAWMS/dkb/pull/39/commits/a99989c2cd6402df9d03a22c12ca22b3185d3dd9#diff-ac4be2bc1e072348e27b33ba39ff0b31R183)

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
mgolosova: And please, check other regular expressions for both `\r` and newline+spaces within them.

mgolosova:
General comment: when you use the `\` symbol to break a string value in two or more strings, you get all the indent spaces within the string.
Use string concatenation instead:
```
>>> def f(a):
...   print "Input string: '" + a + "'"
... 
>>> f("My string starts here"
...   " and goes on here")
Input string: 'My string starts here and goes on here'
```

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
Evildoor: 1. Spaces around % - i made this change following overall logic of removing spaces. Also, "..."%(...) without spaces indicates that it's a single string, just defined in such way. But i'm not adamant about this and we can discuss this point. Inline comments - my apologies, i missed that, will fix.
2. Aha. I see. Missed this as well. I'll try to update my descriptions according to this, then i'll be able to say something about 72 or 100.
3. Makes sense. I'll see what can i do about changing the commit message.
4. I can't say that i'm thrilled with all this history rewriting for the sake of perfection. Especially when it requires additional work. This does not refers to the previous point, which i approve.
5. Same as 3.

Commented debug messages in the code - messages like this one are left because they are used to debug frequently performed operations, so if i'll delete them every time then i'll most likely have to write them again just as often. If logging will be introduced and it would be possible to replace commented debug messages with it - great. But right now we have no better option, right?

^M - uh, no idea what it is. I can't see them in IDLE. This and search - will check.

Utils/Dataflow/030_PDFAnalyzer/manager.py
Evildoor: Spaces around "+" - same as above, i followed the overall logic which removed quite a bit of spaces (and in some places i was wrong, like with inline comments). However, in this case i think space removal makes sense - inside of brackets, spaces are only left after commas. Something like f(a=1, b=2+3, c=4) looks more appealing to me than f(a=1, b=2 + 3, c=4).

Trailing spaces - will check.

Thanks for comments.

Evildoor:
That's useful. Thanks.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
mgolosova: @Evildoor,
> Commented debug messages in the code - messages like this one are left because they are used to debug frequently performed operations, so if i'll delete them every time then i'll most likely have to write them again just as often. If logging will be introduced and it would be possible to replace commented debug messages with it - great. But right now we have no better option, right?

We have an option of stashed changes in Git.
The workflow is as follows:
* make changes to the code, with all the debug messages you need
* use `git add --patch <files>` instead of `git add <files>` to avoid adding debug messages to the staged changes
* `git commit`
* `git stash` (hide local changes -- the debug messages indeed)
* `git push`
* `git stash pop` (restore local changes)

After that you have the code without debug messages in the repository, and with debug messages -- in a local copy.

mgolosova: @Evildoor,

> I can't say that i'm thrilled with all this history rewriting for the sake of perfection. Especially when it requires additional work. This does not refers to the previous point, which i approve.

The repository itself is an additional work ;)

Why do I think it is better to join these two commits:
1. Functionality.
Every commit is a fixed state of the repository. It is better to have a **working state**, then a broken one, so that one can checkout the repository at every point and check how\`s it going.
However here we have 4 repository states in a row where the code can\`t be compiled, _and we have a possibility to fix it._
Yes, this is a development branch, and since we use the **merge** policy (as a contrary to **squash** or fast-forward **rebase and merge**) it means in the "general" branch we still have the working state.
But, by default, when you watch the commit history (say, via the GitHub web-interface, or with `git log`) you still see the linear history with all the commits from the development branch alongside with all the others, so why threat them differently?

2. Readability.
As it is said above, the development commints are not "hidden" behind the merge commit, they are treated as a valuable points of the history. Thus it looks like a good idea to have it more compact, not splitting one logical change ("package import style changed") into a number of commits with the same idea ("import style changed", "import style changed pt.2: fix pt.1", ...)
It gives no new information (except that the first commit is not a thing-in-itself and should not be used individually for any purposes), yet distracts attention from the main course of changes, making one to re-read history backwards to understand which commit was fixed.

I mean **yes, it is a perfectionism. But it is not only for better looks.**

Evildoor:
Fixed/updated everything apart from:
1) History rewriting concerning two commits merging - tried several times to do it, no success so far. I'm sure i'll deal with it eventually, but i still have my doubts this. Even with 2 points above, which are reasonable and make sense.
2) Inline comments - it turned out i've not forgotten to fix them, i just removed them and placed them before/after corresponding lines.
3) Spaces around "%" - i don't see this as a problem, more like a point for discussion, if needed.
4) Spaces around "+" - same.
5) Debug messages - suggested solution with stash in local repository does not looks appealing to me, so i think we should leave them be for now, and wait for logging system.

Evildoor:
Will do, thank you.

Evildoor:
Done. I'll be grateful if you'll do the rebasing and joining (if you have time/energy/etc to do so, of course), since this PEP-8 work already took enough time and we are required to proceed with more pressing tasks.

Side note - thanks for tool recommendation (the second one), it works fine. It also answered one of our questions - "%" should have spaces around it.

mgolosova:
@Evildoor,
rebase done, please check in #40.

I\`ll re-read the commits as well, to ensure I haven\`t mixed up changes from different commits into one.

Doing the rebase I found a couple more forgotten fixes, please see the diff (`git diff origin/PDFAnalyzer-PEP8 origin/PDFAnalyzer-PEP8-rebased`)

Thank you!

Evildoor:
Checked, seems fine apart from some minor things (see #40). I think we can merge. Thank you.

mgolosova:
Now #40 is merged, so I close this PR.

PR #40, opened 2017-08-25
Stage 030 PEP-8 (rebased version)

mgolosova:
It is a rebased version of #39.

The original request is to be used to continue the work on PEP-8.
This request is to be updated when all the requested changes are done and commited to the original one and then merged.

### DIFF
--- a/Utils/Dataflow/030_PDFAnalyzer/flow.py
+++ b/Utils/Dataflow/030_PDFAnalyzer/flow.py
@@ -19,6 +19,7 @@ base_dir = os.path.dirname(os.path.abspath(__file__))
 sys.path.append(os.path.dirname(base_dir))
 import pyDKB
 
+
 def process(stage, msg):
     """ Obtain the PDF name from the input.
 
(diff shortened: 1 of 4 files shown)


Evildoor:
397b5ed - 3rd point in commit should be "added ..." to match the style.

mgolosova:
@Evildoor,
right you are, fixed.

Also fixed couple of consequences of an inaccurate rebase, when line was changed in one commit and then changed back in the next one. Now it looks all right.

You can see, I added one more commit: found a fix of this kind
> `str % (i)` -> `str % i`

in the whitespace commit and desided to move it into a separate one with all similar fixes -- as we fix something once, it is better to fix it everywhere to avoid similar commits in the future.
Please see if it is needed., or you better prefer to have the brackets even around a single argument to the easier format changes in the future.

As you say what you think about the last commit, I will finally merge the request.

Thank you!

Evildoor:
I'm ok with no brackets around single "%" argument. Will try to keep this in mind in the future.

PR #41, opened 2017-09-01
Stage 030 dataset IDs update

Evildoor:
Update contains:

1. Option to extract dataset IDs only from tables.
2. Removal of "possible year" cluttering exported papers.
3. Export error fix.
4. ~~New script for adding hashtags to papers.~~

### DIFF
--- a/Utils/Dataflow/030_PDFAnalyzer/manager.py
+++ b/Utils/Dataflow/030_PDFAnalyzer/manager.py
@@ -19,6 +19,7 @@ default_cfg = {
         "DETERMINE_TITLE":       False,
         "OPEN_INTERVALS_TEXT":   False,
         "OPEN_INTERVALS_TABLES": False,
+        "TABLES_IDS_ONLY":       False,
         "HDFS_PDF_DIR":          "",
         "HDFS_DOWNLOAD_COMMAND": "hadoop fs -get"
     }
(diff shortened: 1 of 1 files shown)


mgolosova:
@Evildoor,
First three commits looks alright, thank you.

But I\`m not sure about the last commit, I mean the functionality.

1. It takes all the IDs it meets, with no regard for the dataset type. Of course, now we have hashtags only for MC16 campaign, but in general -- the script checks _any_ ID it meets against the information in CSV file. Thus it can eventually categorize some realdata dataset with the MC16 hashtags.

2. It creates common hashtags list for the paper, not for datasets.
@maria-grigorieva, I\`m sorry I missed the detailed discussion, is that what we need?

3. I would suggest to create another Dataflow stage for this script -- as I understand, it takes PDFAnalyzer output and transforms it with some external metadata source (currently the CSV file, later -- some database, I believe). It would be reasonable not to encapsulate this two actions (PDF analysis and hashtags mining) into one stage, but divide in two. It will allow us to run this processes independently -- maybe even on different servers.

4. And, talking about Dataflow -- this version doesn\`t work in a streaming mode, so it can not be added to the processing topology.

So I\`d recommend to create another branch, independent from Stage 030, move the last commit there and keep on working on the new stage there.

Evildoor:
Points 3 and 4 - it was agreed that i'll make independent script for now, without dataflow integration, and then we'll see. However, your suggestion to separate the script into its own branch looks reasonable to me and i'll follow it. Removed last commit. Can't say anything on other points.

PR #43, opened 2017-09-14
Changed ontology: reworked keywords.

gubinm:
Added class Keyword, object property hasKeyword and data property hasKeywordText.
Removed data property hasKeyword from Document class.
Reason for change: it was decided on a weekly DKB meeting that keyword-based search will be a feature of DKB, and also that keywords should not be limited in scope to documents. This necessitated separating Keyword into an entity, creating corresponding properties for Keyword, and removing hasKeyword data property from Document class.

### DIFF
--- a/DB/Virtuoso/ATLAS.owl
+++ b/DB/Virtuoso/ATLAS.owl
@@ -44,6 +44,36 @@
     
 
 
+    <!-- http://nosql.tpu.ru/ontology/ATLAS#ContainerObjectAttribute -->
+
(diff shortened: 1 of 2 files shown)


gubinm:
Commit description updated with explanation of changes.

mgolosova:
And, one general comment: the PR title states "reworked keywords", but as far as I see it is more like "Ontology update" with a short list of updates in description, like "added X", "fixed Y", etc., where every X and Y is described in details in its own commit ;)

PR #51, opened 2017-10-19
Merge ora2es to master

IATkachenko:
es-dev was turned to nice look in ora2es.

Added tools and queries that allow get tasks information from Oracle and put it to Elasticsearch.

- process.sh is a main script that do all work;
- Oracle2JSON.py get data from Oracle and convert it to NDJSON format;
- oracle2es.php prepare NDJSON from Oracle2JSON.py for loading to Elasticsearch;
- Elasticsearch/tools - is a set of tools that helps work with Elasticsearch while data processing.


Utils/Elasticsearch/raw/oracle2es.php
```diff
@@ -0,0 +1,102 @@
...
+            }
+          }
+        } else {
+          if ($row[$addon] != '') {
+            $row['extended_tags'][] = $row[$addon];
+          }
+        }
+      }
+    }
+    foreach( $row['physics_short'] as $ps) {
```
mgolosova: As we lay on the taskname nomenclature we can of course be sure that the second field of the taskname exists and thus this key `physics_short` will be defined, I just wanted to point out to the potential of the crush.
If we define this key within an `if` clause, we can never actualy be sure ;)

mgolosova:
In the description of dd26bdc there last two hashes of the `Added ElasticSearch and Oracle Helpers` commit.
Maybe a typo?

Utils/Elasticsearch/tools/load_mapping.sh
```diff
@@ -1,20 +1,25 @@
...
+ES_PORT='9200'
+
+while getopts c: opt; do
   case $opt in		
-    e) ES="$OPTARG";;
+    c) [ -f "$OPTARG" ] && ES_CONFIG="$OPTARG" || echo "Config $OPTARG not found! Will use default values!" ;;
     *) usage;
   esac
 done
+[ -f "$ES_CONFIG" ] && . $ES_CONFIG || echo "Could not found es_confg file $ES_CONFIG! Will use default values" >&2
```
mgolosova: The first check (in the `case` clause) might be excessive; also, it tells about error to STDOUT, while the second check -- to STDERR.

Utils/Elasticsearch/raw/oracle2es.php
IATkachenko: fixed

Utils/Elasticsearch/tools/load_mapping.sh
IATkachenko: We should generate warning if user configuration file not exists and we must not tries to load default config if it not exists. 
So we should have two checks.

STD{OUT,ERR} issue fixed.

IATkachenko:
Updated.

Utils/Elasticsearch/tools/load_mapping.sh
mgolosova: I see, so it is to have possibility to fall back to the default config file even if user has specified its own.
Not sure if I would do same, still it makes sense :) Thank you!

mgolosova:
Thank you!
Merging.

PR #52, opened 2017-10-20
pyDKB/enhance: add configurable EOM/EOP markers

anastasiakaida:
AbstractStage.py
*New arguments
-e - end of message
-E - end of process
*Default EOP/EOM values for each mode

AbstractProcessorStage.py:
*Hard-coded symbols were deleted
*__stream_EOMessage and __stream_EOProcess are excess variables
*__EOProcess --> self.ARGS.eop
*__EOMessage --> self.ARGS.eom


Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
```diff
@@ -303,13 +303,10 @@ def stream_input(self, fd):
         Split stream into messages;
         Yield Message object.
         """
-        if self.ARGS.eom == '\n':
+        if self.ARGS.eom:
             iterator = iter(fd.readline, "")
             for line in iterator:
                 yield self.parseMessage(line)
```
mgolosova: I\`m afraid it won\`t work correctly.
Say we have `EOM == '\n\n'`, two subsequent `newline` symbols. In theory, it would allow us to have the `newline` symbol within the message, but as we still use `readline` -- each line is taken as a separate message.
Same if the EOM is not about `newline` at all...

mgolosova: And, byu the way: what will happen, if there`s no EOM marker specified?
I think it still needs to raise some exception, like "Hey, how do you expect me to say one message from another, if there`s no EOM marker?!"
Maybe it should be addressed at the very beginning, when the arguments are parsed, like "if None, set '\n'; if empty, throw an exception."

anastasiakaida: Got it. I'll make some changes.

PR #54, opened 2017-10-23
Dataflow chain from Oracle to ES

mgolosova:
Tested on bamboo.

ToDo:
- [x] Stage 09 (Oracle connector)
- [x] Stage 19 (prepare data for upload)
- [x] Stage 69 (upload data 2 ES)
- [x] run-script for the chain...
- [ ] ~~...OR make things Kafka-styled (which means we need to finaly merge `dataflow-with-doc-content` to master)~~ [1]

To fix:
- [x] chaining script: add tmp storage for the data between Stage 19 and Stage 69 (to avoid data loss)
- [x] chaining script: add `\x00` in the end of the processing to make sure the last records will be uploaded no matter if they fil the `BATCH_SIZE` or not
- [x] Stage 69: fix DELIMITER and DELIMETER variables issue. We are just lucky that `read -d` takes empty line as `'\0'`.
- [x] Stage 09: duplicate output (unjoined and joined records)
- [ ] ~~Stage 09: fix Elasticsearch/process.sh rto work with the new version of Oarcel2JSON~~ [2]
- [ ] Stage 69: do smth with `load_data.sh: line 52: echo: write error: Interrupted system call` [3]

---
[1] run-script is enough for now
[2] the one who added this item never confessed
[3] it does not affect things heavily, but it must be investigated and addressed one day


Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -0,0 +1,307 @@
...
+except:
+    print "****ERROR : DButils. Cannot import cx_Oracle"
+    pass
+
+# Policies
+PLAIN_POLICY = 'PLAIN'
+SQUASH_POLICY = 'SQUASH'
+
+def connectDEFT_DSN(dsn):
+    connect = cx_Oracle.connect(dsn)
```
IATkachenko: We should catch exceptions for failed connections.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -0,0 +1,307 @@
...
+            records = squash(conn, queries, offset_date, end_date)
+        elif mode == PLAIN_POLICY:
+            records = plain(conn, queries, offset_date, end_date)
+        for r in records:
+            r['phys_category'] = get_category(r)
+            sys.stdout.write(json.dumps(r) + '\n')
+        update_offset(end_date)
+        offset_date = end_date
+        if not final_date_cfg:
+            final_date = date2str(datetime.now())
```
IATkachenko: final_date is determined at 133-136. Should it be redetermined at every loop iteration?

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -0,0 +1,307 @@
...
+
+def get_initial_date():
```
IATkachenko: Unused function.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -0,0 +1,307 @@
...
+    """
+    Convert human-readable interval into seconds.
+
+    If no suffix present, take step as seconds.
+    """
+    try:
+        return int(step)
+    except ValueError:
+        pass
+    if len(step) < 2:
```
IATkachenko: We will fail here if **step** less than 10 (default seconds, without suffix).

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -0,0 +1,307 @@
...
+        raise ValueError("Failed to decode interval: %s" % step)
+    suffix = { 'd': 86400, 'h': 3600, 'm': 60, 's': 1}
+    val = step[:-1]
```
IATkachenko: If step just 10 (without suffix, to pass length check), then **val** will be just 1.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -0,0 +1,307 @@
...
+    try:
+        mul = suffix[step[-1]]
```
IATkachenko: All code of this function assumes that suffix MUST exists, but comment said that it is not necessary.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -0,0 +1,307 @@
...
+            if re.search('_tt_', phys_short) is not None: categories.append("TTbar")
+            if re.search('upsilon', phys_short) is not None: categories.append("BPhysics")
+            if re.search('tanb', phys_short) is not None: categories.append("SUSY")
+            if re.search('4topci', phys_short) is not None: categories.append("Exotic")
+            if re.search('xhh', phys_short) is not None: categories.append("Higgs")
+            if re.search('3top', phys_short) is not None: categories.append("TTbarX")
+            if re.search('_wt', phys_short) is not None: categories.append("SingleTop")
+            if re.search('_wwbb', phys_short) is not None: categories.append("SingleTop")
+            if re.search('_wenu_', phys_short) is not None: categories.append("Wjets")
+        return categories
```
IATkachenko: Missed default value if taskname is not None and all re.search returns None.

We have to check that categories length > 0

Utils/Elasticsearch/mapping/prodsys.mapping
```diff
@@ -145,11 +145,11 @@
           "type" : "keyword",
           "ignore_above" : 256
         },
-        "input_datasets" : {
+        "input" : {
```
IATkachenko: input_datasets -> input in mapping, but not in any place in code.

Utils/Elasticsearch/mapping/prodsys.mapping
```diff
@@ -145,11 +145,11 @@
...
           "type" : "keyword",
           "ignore_above" : 256
         },
-        "output_datasets" : {
+        "output" : {
```
IATkachenko: output_datasets -> output in mapping, but not in any place in code

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: Yes.
For the initial data load the loop can interate for quite a long time.
If we asked the script to load all the data from `$then` till `$now`, we want to have this `$now` up to date -- however long it took to reach it.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: Will not, due to 201-204:
```
>>> interval_seconds('10')
10
>>> interval_seconds('1')                                                                                                                                                                                                                      
1
>>> interval_seconds('1s')
1
>>> interval_seconds('1d')                                                                                                                                                                                                                     
86400
>>> interval_seconds('10s')                                                                                                                                                                                                                    
10
```

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: It seems to me that it will return empty list in this case (defined at [267](https://github.com/PanDAWMS/dkb/pull/54/files/8fc1fe9e75c92877c4bdb16c043d9d85ce7fed69#diff-4c221825c8bb3980b06a244fe9499716R267)).
Have I missed something?

Utils/Elasticsearch/mapping/prodsys.mapping
mgolosova: Right, that change was "hidden".
Here: https://github.com/PanDAWMS/dkb/pull/54/commits/5f5a1932edb474431d0703b1ac8b18cd63b20741 (DF/ora2es/oc: update ES mapping according to 5ea3d9d)

First there were implicitly specified field names for "input_datasets" and "output_datasets". But they had to be hardcoded into Python code, as there were no formal rule to get these names from original data. So they were shortened to "input" and "output" -- values of Oracle `dataset_type` field.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: Addressed in PR #80

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: Addressed in PR #80

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
IATkachenko: Yes, empty list will be returned, but for unknown categories returns "Uncategorized".

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
IATkachenko: OK, now I see the logic.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
IATkachenko: If it's OK - fine.
But why for tasks without taskname returns "Uncategorized"? It seems be same for any uncategorized tasks.

PR #56, opened 2017-10-28
[WIP] pyDKB dev

mgolosova:
This PR is replaced with #70.
Left open for a while to track ToDo list for pyDKB till it is moved to some more appropriate place.

ToDo:
- [ ] rethink the logic of output file naming and store directory. It is necessary as when we run json2JSON processor, it is easy to get the collision of input and output files.
- [ ] more accurate exception handling (so that app using pyDKB had no need in `try...except` around stage methods)
- [ ] add `--config FILE` as default parameter, with config parsing


mgolosova:
ToDo list moved to #99.

PR #62, opened 2017-11-02
Stage 055 (document2TTL) fix

mgolosova:
Fix issue with abstract extraction.


mgolosova:
Changes of the first commit (523a837) should actually go to pyDKB library, but for a time being let them stay here (and in multiple other stages).

PR #65, opened 2017-11-07
PEP8 fix

mgolosova:
PEP 8 fixes, mainly performed by autopep8 v.1.3.2 (https://pypi.python.org/pypi/autopep8).

Please check your modules to work properly.


anastasiakaida:
Last time I could run scripts as:
`$ python documents2ttl.py -i input -o output`
or
`$ python docContent2TTL.py -i input -o output`

I'm not sure what is the cause of an error, but now it looks like this:
`$ python docContent2TTL.py -i input -o output
(ERROR) Failed to import pyDKB library: name 'os' is not defined`

But os is an stdlib, isn't it? :/

mgolosova:
@anastasiakaida, thank you.
`os` is a standard library indeed, but it must be imported before using ;) Now fixed.

mgolosova:
Rebased to join latter fixes with their "parent" commints.

Issue #46, opened 2017-10-10
DSID column


Evildoor:
Come papers label the column with dataset IDs as "DID" or "DCID" instead of "DSID". Regular expression should be adjusted.

Evildoor:
Fixed in #73.

Issue #14, opened 2017-06-01
Incorrect underlines scanning


Evildoor:
In some documents PDFMiner incorrectly scans underlines as different symbols. This interferes with regular expression searches of dataset names and requires modifications for them. Solved/unsolved cases:

- [x] Spaces

- [ ] Dashes

PR #73, opened 2017-11-17
Stage 030 update

Evildoor:
Small update with changes serving for:

- Catching up with dataflow-with-document-content -> master migration.
- ~~Creating a base for extension development.~~ (Removed per conversation below.)

### DIFF
--- a/Utils/Dataflow/030_PDFAnalyzer/manager.py
+++ b/Utils/Dataflow/030_PDFAnalyzer/manager.py
@@ -256,7 +256,8 @@ re_pdfname = re.compile("/([^./]+)\.pdf$")
 re_table_header = re.compile("Table \d+:.*?\n\n", re.DOTALL)
 re_table_header_short = re.compile("Table (\d+):")
 re_table_datasets = re.compile("(?:sample|dataset|run)")
-re_column_with_datasets = re.compile("^(?:ds[-_ ]?|mc[-_ ]?|data ?|dataset ?"
+re_column_with_datasets = re.compile("^(?:d[cs]?[-_ ]?|mc[-_ ]?|data ?"
+                                     "|dataset ?"
                                      "|period|request ?|run ?|sample ?)(?:id"
                                      "|number|period|range|sample|set)")
 re_dsid = re.compile("^\d{4,8}$")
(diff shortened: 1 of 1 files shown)


Evildoor:
CSV file - i added it because i thought that it was better than either:

1. Making a commit which does nothing besides creating a directory.
2. Not making a commit, and then creating this directory in each branch which adds an extension.

Error - makes sense, will try to keep that in mind.

Regular expression - this change is solving #46. When this pull request will be merged, issue will be closed and commit/request will be linked in it. I can link issue here or in pull request initial message, but not in commit message (i think).

mgolosova:
@Evildoor,
I see the problem.
For Git itself, there\`s nothing wrong in same dir created in different branches -- as it is interested mostly in files, to create or delete a directory is kind of a "side-effect" of adding/removing data from it. Git would only worry if you add different files with the same name, and then try to merge them.

However if there are too many branches, and you\`d like to avoid `mkdir` for each of them, I would recomend to create a branch like "stage-30-extensions", create empty dir [1] (or put there all the source data you are gonna need for the extensions), and then create branches for individual extensions from that one.
Then even if you create PR from that branches directly to the `master` branch, the "create empty dir" commit will disappear from all the PRs as soon as the first one is merged.

[1] As Git is not interested in any directory unless it has at least one tracked file, to add an empty dir is a tricky thing. But you can add something like README file to it, where you are going to make notes about "which data are for which extension" (if it is a directory for source data only) or describing extensions themselves (if their code to be stored in it as well), or something.

Evildoor:
I doubt that there will be too many branches with different extensions, but idea of creating stage-30-extensions branch looks appealing to me. Will do.

Evildoor:
Removed commit related to extensions, updated descriptions for other ones.

mgolosova:
@Evildoor,
Thank you, I think the PR can be merged now.
Just ine question: am I right and the 862e7cb is related to the issue #14? I`d like to mention it in the merge commit at least, as in console no one will see cross-links that will possibly appear later in the web-interface ;-)

Evildoor:
@mgolosova 
Good.
Uh, no, 862e7cb is related to #46, not #14.

mgolosova:
@Evildoor,
right, thank you. Was not that obvious ;)

mgolosova:
@Evildoor ,
request merged, thank you. Please remove the branch if it is no longer needed.

PR #84, opened 2017-11-24
Adding data from AMI

mgolosova:
Fields to add:
* k_factor
* cross_section
* cross_section_ref
* gen_filt_eff

ToDo:
- [x] change ES mapping;
- [x] add stage (say, 095_datasetInfoAMI):
   input from 093 (see PR #102), output -- to 019 (see PR #86);
   pyAMI with `GetPhysicsParamsForDataset` command will help;
- [x] introduce it to the `data4es` pipeline;
- [x] decide what to do with extra parameters taken from AMI (e.g. `processGroup`) -- remove, add to mapping, leave "as is"? *(`processGroup` and `mePDF` are added to mapping)* [1]
- [x] do it;
- [x] add input/output data samples;
- [x] test properly on aiatlas171;
- [x] update output example;
- [x] append 8a836c4 to where it belongs (69797cf)
- [x] append f9d740e to where it belongs (19a1d5d)
- [ ] ...
- [ ] ~batch mode for performance;~ [2]
  *However good this idea might look, it was decided to be addressed outside this PR.*
  *See #108 for details.*

---
This PR to be finished after ~#88~, ~#85~ and ~#102~ (_--@mgolosova_)

[1] However we are not sure if all the fields possible were found.
[2] AMI allows to get data for a comma-separated list of datasets and it will be faster than doing it one-by-one (see getMetadata.py by AMI):
```
datasetsToQuery = ",".join(dataset_values.keys())
res = client.execute(['GetPhysicsParamsForDataset',"--logicalDatasetName=%s"% datasetsToQuery,"--timestamp='%s'"%args.timestamp], format='dom_object')
```

PR #79, opened 2017-11-22
[WIP] Stage 091 (rucioDatasets)

mgolosova:
Add information about datasets from Rucio to the ora2es chain (009-019-069).

PR #86, opened 2017-11-27
Stage 019 (esFormat)

mgolosova:
Transform Stage 019 to accept data not only from Stage 009, but also 091 (see #79) and 092 (see #84).

ToDo:
- [x] rename to fit new task;
- [x] make index configurable;
- [x] input data special fields ('_id', '_type', '_parent') (update Stage 016, Stage 091);
- [x] rewrite `check_input()` to check the special fields.


mgolosova:
Closed accidentaly with inaccurate `git push` to the target branch.

PR #88, opened 2017-11-28
Stage 016 (task2es)

mgolosova:
Move data transformation functionality from Oracle Connector (Stage 009) to a new processing stage (016).

Main changes:
* added skeleton stage for processing stages;
* README updated;
* new stage 016 (task2es) took data transformation functionality from Stage 09;
* `hashtag_list` field is no longer a comma-separated list -- it is an actual list.

ToDo:
- [x] rebase the branch to current master (fixing data samples)
- [x] split dot-sepatated `output_formats` into a list (don`t forget to check mapping)


mgolosova:
In the commits, that update data samples, there sometimes changes are displayed though it is not obvious what has changed.
I believe it happened due to the field ordering within JSON (or to the lack of it).
When sample data really changed, diff looks more like "whole file rewritten" ;)

Can`t avoid this changes as sample data every time were reconstructed from scratch.

Utils/Dataflow/README
```diff
@@ -34,6 +38,8 @@ Dataflow 							Description
 (054|055)/output -> 060						upload links, papers, support docs and datasets TTL
 								  files to Virtuoso
 
+009/output -> 016/output -> 019/output -> 069                   Task metadata from Oracle to Elasticsearch
+
```
maria-grigorieva: I think here should be the description of stage 69 ?

mgolosova: Right you are, thank you!
Here it is a description of the whole chain, so I will add it to the previous section (where all the other stages are described).

mgolosova:
Rewrote commit `DF: update README (ora2es chain).`  to add missed info about Stage 069 (upload2es).

konvpalto:
Will review this one Really Soon (tm), yeah!

mgolosova:
Rebased on current master (780a679) to avoid additional `"Fix codestyle"` commits in the branch.

Utils/Dataflow/skel/stage.py
```diff
@@ -0,0 +1,66 @@
+#!/bin/env python
+"""
+DKB Dataflow stage XXX (StageName).
+
+Stage short description
+
+Authors:
+  Author Name (author@cern.ch)
+"""
```
maria-grigorieva: unknown author? ))

mgolosova: Yeah, I didn\`t want my name to appear in all the lated scripts as the one and only author\`s name. What about yours? ;-)

maria-grigorieva: But it's your script ) so, I think there is no problem.

mgolosova:
Added splitting of 'output_formats', but not sure about mapping. Do I need to change something there?
```
        "output_formats": {
          "type": "keyword",
          "ignore_above" : 256
        },
```

According to ES documentation,
```
In Elasticsearch, there is no dedicated array type. Any field can
contain zero or more values by default, however, all values
in the array must be of the same datatype. 
```
Also, for "hashtag_list" there is right the same field description, though it is an array of values.

So it seems to me that there`s nothing should be done with mapping... but I would like to see some confirmation for this :)

maria-grigorieva:
I suppose we can leave mapping "as is", as all values in "output_formats" so as in "hashtag_list" are the same type. 

Confirmation from ES documentation: 

> Multi-value fields and the inverted index

The fact that all field types support multi-value fields out of the box is a consequence of the origins of Lucene. Lucene was designed to be a full text search engine. In order to be able to search for individual words within a big block of text, Lucene tokenizes the text into individual terms, and adds each term to the inverted index separately.

This means that even a simple text field must be able to support multiple values by default. When other datatypes were added, such as numbers and dates, they used the same data structure as strings, and so got multi-values for free.

PR #85, opened 2017-11-24
Stage 091 (datasetsRucio)

mgolosova:
New stage to extract additional information about datasetd from Rucio

ToDo:
- [x] update ES mapping, adding new data;
- [x] implement basic functional;
- [x] make things work in the frame of pyDKB;
- [x] adjust output (one dataset -- one message);
- [x] tell Stage 019 about different index names;
- [x] tell Stage 019 about parent/child relationship;
- [x] adjust ora2es run-script to branch dataflow...
- [x] ~~OR make things Kafka-style.~~ _(removed in favor of previous item)_
- [x] merge PR #88 
- [x] rebase on new `master` after #88 
- [x] ~~update Dataflow/README~~ _(was erroneously updated in #87)_

PR #89, opened 2017-11-28
Dataflow ami2es maria

maria-grigorieva:
- Implemented the addition of AMI parameters in stage 092_datasetInfoAMI
- updated ES mapping with parent-child relations between tasks and datasets

### DIFF
--- a/Utils/Dataflow/092_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/092_datasetInfoAMI/amiDatasets.py
@@ -1,0 +1,128 @@
+#!/bin/env python
+import json
+import re
+import sys
+import os
+try:
+    import pyAMI.client
+    import pyAMI.atlas.api as AtlasAPI
+    import pyAMI.config
+except ImportError:
+    sys.stderr.write("Unable to find pyAMI client.")
+    sys.exit(1)
+
+base_dir = os.path.abspath(os.path.dirname(__file__))
+
+try:
+    dkb_dir = os.path.join(base_dir, os.pardir)
+    sys.path.append(dkb_dir)
+    import pyDKB
+except Exception, err:
+    sys.stderr.write("(ERROR) Failed to import pyDKB library: %s\n" % err)
+    sys.exit(1)
+
+ami_client = None
+PHYS_VALUES = [{"ami": "genFiltEff", "es": "gen_filt_eff"},
+               {"ami": "crossSection", "es": "cross_section"},
+               {"ami": "crossSectionRef", "es": "cross_section_ref"},
+               {"ami": "kFactor", "es": "k_factor"}]
+
+def main(argv):
+
+    stage = pyDKB.dataflow.stage.JSONProcessorStage()
+
+    stage.add_argument('--userkey', help='PEM key file', required=True)
+    stage.add_argument('--usercert', help='PEM certificate file', required=True)
+
+    exit_code = 0
+    try:
+        stage.parse_args(argv)
+        stage.process = process
+        init_ami_client(stage.ARGS.userkey, stage.ARGS.usercert)
+        stage.run()
+    except (pyDKB.dataflow.exceptions.DataflowException, RuntimeError), err:
+        if str(err):
+            str_err = str(err).replace("\n", "\n(==) ")
+            sys.stderr.write("(ERROR) %s\n" % str_err)
+        exit_code = 2
+    finally:
+        stage.stop()
+
+    sys.exit(exit_code)
+
+def init_ami_client(userkey, usercert):
+    """
+    Initialisation of AMI client into the global variable
+    :param userkey: user key pem file
+    :param usercert: user certificate pem file
+    :return:
+    """
+    global ami_client
+    try:
+        ami_client = pyAMI.client.Client('atlas', key_file=userkey, cert_file=usercert)
+        AtlasAPI.init()
+    except Exception:
+        sys.stderr.write(
+            "Could not establish pyAMI session. Are you sure you have a valid certificate?")
+        sys.exit(1)
+
+def process(stage, message):
+
+    stage.output(pyDKB.dataflow.messages.JSONMessage(amiPhysValues(message.content())))
+
+    return True
+
+def amiPhysValues(data):
+    """
+    Add elements in JSON string, according to theirs names in ES mapping
+    - gen_filt_eff
+    - cross_section
+    - k_factor
+    - cross_section_ref
+    """
+    dataset = data['datasetname']
+    container = remove_tid(dataset)
+    try:
+        res = ami_client.execute(['GetPhysicsParamsForDataset',
+                                  "--logicalDatasetName=%s" % container],
+                                 format='json')
+        json_str = json.loads(res)
+        for row in json_str['AMIMessage'][0]['Result'][0]['rowset'][0]['row']:
+            p_name, p_val = None, None
+            for field in row['field']:
+                if field['@name'] == 'paramName':
+                    p_name = field['$']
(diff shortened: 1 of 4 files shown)


Utils/Dataflow/092_datasetInfoAMI/amiDatasets.py
```diff
@@ -0,0 +1,123 @@
...
+    except:
+        sys.stderr.write(
+            "Could not establish pyAMI session. Are you sure you have a valid certificate?")
+        sys.exit(1)
+
+def process(stage, message):
+
+    str = amiPhysValues(message.content())
+    stage.output(pyDKB.dataflow.messages.JSONMessage(amiPhysValues(message.content())))
+
```
mgolosova: Two comments in a row:
* `str` is a Python keyword (namely the type), so reusing it as a variable name you can invite unpredictable behaviour in the code (say, in conditions like `if type(x) is str`);
* `amiPhysValues()` is called twice

Utils/Dataflow/092_datasetInfoAMI/amiDatasets.py
```diff
@@ -0,0 +1,123 @@
...
+        res = ami_client.execute(['GetPhysicsParamsForDataset',
+                                  "--logicalDatasetName=%s" % container],
+                                 format='json')
+        json_str = json.loads(res)
+        for row in json_str['AMIMessage'][0]['Result'][0]['rowset'][0]['row']:
+            for field in row['field']:
+                if field['@name'] == 'paramName':
+                    for f in row['field']:
+                        if f['@name'] == 'paramValue':
+                            data[field['$']] = f['$']
```
mgolosova: Took a while to understand what\`s going on :) Two loops over same `row['field']` made me afraid of the data structure...
I can\`t check it right now (failed to connect to AMI), but wouldn`t it be the same as
```
p_name, p_val = None, None
for field in row['field']:
    if field['@name'] == 'paramName':
        p_name = field['$']
    elif field['@name'] == 'paramValue':
        p_val = field['$']
    if p_name and p_val:
        data[p_name] = p_val
        p_name, p_val = None, None
        coninue
```
?
 I don\`t think there are so many `fields` in `row` to make two loops much worse than one, so just suggesting :)

Utils/Dataflow/092_datasetInfoAMI/amiDatasets.py
```diff
@@ -0,0 +1,123 @@
...
+
+def change_key_names(data):
+    """
+    Changing parameter names according to PHYS_VALUES dictionary.
+    :param data: JSON string
+    :return: JSON string
+    """
+    str = json.dumps(data)
+    for item in PHYS_VALUES:
+        str = str.replace("\""+item["ami"]+"\":", "\""+item["es"]+"\":")
```
mgolosova: Same as above about variable names. Possible misundestanding can also look like this:
```
>>> str.replace
<method 'replace' of 'str' objects>
>>> str.replace('aac', 'a', 'b')
'bbc'
>>> str = 'aac'
>>> str.replace
<built-in method replace of str object at 0x7ff56c8d7530>
>>> str.replace('aac', 'a', 'b')
TypeError: an integer is required
```
The one who see `str.replace` automatically thinks about first variant, while actually it is the second.

Utils/Dataflow/092_datasetInfoAMI/amiDatasets.py
```diff
@@ -0,0 +1,123 @@
...
+    return json.loads(str)
```
mgolosova: It surely must work if there only AMI fields in the original JSON, as field values are mostly numeric and will hardly interfire with names. But I\`m not sure about the rest of the document. Can\`t something similar appear in the dataset name?
I would suggest something like
```
for item in PHYS_VALUES:
    if item["ami"] in data:
        data[item["es"]] = data.pop(item["ami"])
```

And the same as above about variable names. Possible misundestanding can also look like this:
```
>>> str.replace
<method 'replace' of 'str' objects>
>>> str.replace('aac', 'a', 'b')
'bbc'
>>> str = 'aac'
>>> str.replace
<built-in method replace of str object at 0x7ff56c8d7530>
>>> str.replace('aac', 'a', 'b')
TypeError: an integer is required
```
The one who see `str.replace` automatically thinks about first variant, while actually it is the second.

Utils/Dataflow/092_datasetInfoAMI/amiDatasets.py
```diff
@@ -0,0 +1,122 @@
...
+def change_key_names(data):
+    """
+    Changing parameter names according to PHYS_VALUES dictionary.
+    :param data: JSON string
+    :return: JSON string
+    """
+    for item in PHYS_VALUES:
+        if item["ami"] in data:
+            data[item["es"]] = data.pop(item["ami"])
+    return json.loads(str)
```
mgolosova: We need to return data here, as there\`s no `str` left to return after last changes.

Utils/Dataflow/092_datasetInfoAMI/amiDatasets.py
```diff
@@ -0,0 +1,127 @@
...
+def change_key_names(data):
+    """
+    Changing parameter names according to PHYS_VALUES dictionary.
+    :param data: JSON string
+    :return: JSON string
+    """
+    for item in PHYS_VALUES:
+        if item["ami"] in data:
+            data[item["es"]] = data.pop(item["ami"])
+    return json.loads(data)
```
mgolosova: Why `json.loads()`?
`data` is supposed to be of a `dict` type initially (though in the docstring it is said to be `string`, as well as the return value).

Utils/Elasticsearch/mapping/prodsys.mapping
```diff
@@ -175,24 +175,47 @@
...
+                  "keyword" : {
+                      "type" : "keyword",
+                      "ignore_above" : 256
+                  }
+               }
+             },
+             "bytes" : {
+                 "type" : "long"
+             },
+             "deleted": {
```
mgolosova: Why to add Rucio parameters here? They are added in PR #85. And has nothing to do with the current AMI-branch.
I see two variants: 
1) to change mapping as if there were no Rucio parameters to join mappings later during rebase/merge (but please check the mapping from #85 carefully to avoid conflicts: e.g. `output_datasets` are called `output_dataset` there to correspond `task`, which are singular; after all, it is "type" of a document, so why would the "type" name be plural?)
OR
2) to cherry-pick commit 68a8252 from PR #85, and add AMI parameters in the next commit.

I believe that the second variant would be the easiest one -- unless we change something in the mapping before everything is merged. When we rebase this PR on top of the `master` after #85 is merged, this cherry-picked commit will become empty and thus will be removed automatically, while the AMI-related one will stay where it must be. Thus we will escape possible conflicts due to same parts of one file changed twice.

And I think it will be good to rename the file according to the `template` field. When we load the mapping to ES, it is named after the file name, while `template` is used for applying this mapping to the matching indices. Nothing wrong in having this two out of sync, it just makes easier to make a mistake and drop wrong template if needed ;) Or have two mappings for same index -- I don`t know how it`s gonna work, by the way.

maria-grigorieva: I did the 2nd variant and think it's must be OK now. Thank you for noticed this mistake :)

mgolosova: You`re most welcome. Now check the diff ;)

mgolosova:
Closed, as the mapping commit couldn`t be merged to the target branch without changes.
So the changes were cherry-picked to PR #84.

Issue #92, opened 2017-12-09
Stage 009 (Oracle Connector): refresh Oracle connection


mgolosova:
When the connectinon stays idle for quite a while, on the next query there might appear an error `cx_Oracle.DatabaseError: ORA-02396: exceeded maximum idle time, please connect again` and then the Stage fails to continue working.
There\`s nothing fatal in this error so all the `Connector` must do is to catch the error and re-initialize the connection.

mgolosova:
Fixed.

PR #93, opened 2017-12-09
Stage 009 (Oracle Connector): update

mgolosova:
Improvements and fixes.

Major changes:
* current offset is to be stored in a separate file (instead of main config) [1];
* `squash_records()` function was re-implemented so that it work as a generator [2];
* query files configuration is considered to be related to config dir;
* output in an unbuffered mode;
* issue #92 fixed (restoring of broken session/connection to Oracle);

Minor changes:
* docstrings added/updated;
* internal format for datetime became `datetime.datetime` (instead of `str`);
* parameter name changed for some functions;
* more universal algorithm for `squash_records()`;
* separate class for `dbConnection`;
* help message fixed.

ToDo:
- [ ] issue #92 *(need to check it again, it seems to me that it didn`t work in the "production")*
- [x] reset current offset if `initial` one is later

---
[1] It is required for automatic deployment via Puppet:
* the config file is to be generated automatically by Puppet agent running on the node, so on every run it is to be rewritten;
* we would like to restart the `data4es` chain if the config file has changed (say, Oracle access parameters or SQL queries changed). If we store the `current_offset` in the same file, it will be considered "changed" on every Puppet agent run.

[2] We try to avoid storing all the records received from Oracle in the memory. Instead we use generators to process one record at a time. The `squash_records` appeared the only function left as non-generator, so despite other functions supposed to operate with one record at a time, all the records from `datasets` query inevitably massed in the memory as soon as `squash_records` was called.


anastasiakaida:
As I understand, now this PR can be reviewed. If you don't mind, I'll make a review tomorrow. :)

mgolosova:
@anastasiakaida, it would be great. Thank you!

anastasiakaida:
My review will appear until today's evening.

anastasiakaida:
I've found [here](https://github.com/PanDAWMS/dkb/blob/59f7325ad525a1b4b007f824fce30aa4b124bf9f/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py#L48) the usage of ConfigParser(). Can we use SafeConfigParser()? In the documentation, there is a recommendation for SafeConfigParser() usage. You can see it [here](https://docs.python.org/2/library/configparser.html#ConfigParser.SafeConfigParser).
________
What does [this](https://github.com/PanDAWMS/dkb/blob/59f7325ad525a1b4b007f824fce30aa4b124bf9f/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py#L224) argument (I mean `1000`) mean? I tried to find it out and it didn't bring any result so I decided to ask. And, in advance, why exactly `1000`?

To be continued...

mgolosova:
@anastasiakaida,

> I've found here the usage of ConfigParser(). Can we use SafeConfigParser()? In the documentation, there is a recommendation for SafeConfigParser() usage. You can see it here.

I saw this recomendation too but as the documentation never explain even what exactly the "magical interpolation" (which has safer implementation in Safe one) is, I couldn`t be sure if it will affect anything.
Now I did a quick research, and exactly, they say that the improved safety is related to the string substitution in config files, so that you can put reusable values into some kind of variables:
```
[My Section]
foodir: %(dir)s/whatever
dir=frob
```
I think in our case Safe one can actually be used, just to ensure that if someone decide to use this feature everything will be alright :) 
Will replace it as soon as I can (as well as in pyDKB implementation for the `--config` parameter, by the way). Thank you!

> What does this argument (I mean 1000) mean? I tried to find it out and it didn't bring any result so I decided to ask. And, in advance, why exactly 1000?

1000 goes from `OracleConnection.results(..., arraysize, ...)` to `cx_Oracle.Cursor.fetchmany(arraysize)` (see details [here](http://cx-oracle.readthedocs.io/en/latest/cursor.html#Cursor.fetchmany)).
This parameter determines number of result rows taken from Oracle DB at once to be cached on our (client) side for further processing. This caching is used, on the one hand, to avoid sending too many requests to the remote database (one request per row would make 1000 requests "give me next row" ([fetchone](http://cx-oracle.readthedocs.io/en/latest/cursor.html#Cursor.fetchone)) instead of one "give me next 1000 rows"). On the other, it allows us not to store all the results in the local memory (as we would have to do if we first took all the rows form DB, and only then started to process them) ([fetchall](http://cx-oracle.readthedocs.io/en/latest/cursor.html#Cursor.fetchall)).

In theory we can query quite a lot of data (metadata for all the `tasks` even for a few days will give us millions of records), so we need this caching to avoid slowing down too much -- or running out of memory.

> To be continued...

Looking forward ;)

Utils/Dataflow/009_oracleConnector/OffsetStorage.py
```diff
@@ -0,0 +1,59 @@
+"""
+Classes representing Offset Storage.
+Provides methods:
+- init (create and open the storage)
+- get (get current offset)
+- commit (write new offset)
+"""
+
+import os.path
+
```
maria-grigorieva: import sys ?

maria-grigorieva:
In Oracle2JSON.py
------------------
File "Oracle2JSON.py", line 147, in init_offset_storage
    elif current_offset < initial_date:
TypeError: can't compare datetime.datetime to str

maria-grigorieva:
File "Oracle2JSON.py", line 166, in commit_offset
    timestamp = time.mktime(new_offset.timetuple())
AttributeError: 'str' object has no attribute 'timetuple'

mgolosova:
Tested Issue #92 in case of long timeout between two SQL requests to Oracle, and it worked fine.

Still I am not sure that "that`s it" and think I might have missed something...
Going to test PR #84 (AMI part of the dataflow) together with this PR (as these changes are vital for the proper work of the whole process), and if I see any sights of `ORA-02396`, I will update information here.

mgolosova:
History refactoring is coming. What will be done:
* fixup 5e9a4ec -- `Move offset ftom main config file`
 with:
  * 4707cc8 -- `Fix issue with reading from empty OffsetStorage`;
  * 22118cd -- `Fix missed import`;
  * 8061d04 -- `Fix codestyle`;
* fixup 59f7325 -- `Use initial date as offset if offset storage is empty.`
  with 7c30f51 -- `Fix issue with OffsetStorage initialization.`

mgolosova:
History update done.
Previous version of this PR is stored for reference in branch `pr-93-stage-009-01`.

This change:
```
diff --git a/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py b/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
index 27fc3bc..a11ecf4 100755
--- a/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
+++ b/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
@@ -512,5 +512,6 @@ def parsingArguments():
         sys.exit(1)
     return parser.parse_args()
 
+
 if __name__ == '__main__':
     main()
```
is related to the parallel PR #96 -- `pycodestyle` requires two empty lines *before* and *after* top-level function definition; `pep8` thought it is OK to have two empty lines only *between* them.

To avoid fixing it later, I decided to fix it right here -- the more so, as we have a separate `Fix codestyle` commit.

Utils/Dataflow/009_oracleConnector/OffsetStorage.py
```diff
@@ -0,0 +1,60 @@
...
+    def __init__(self, filename):
+        """ Create and open storage. """
+        try:
+            if os.path.exists(filename):
+                self.storage = open(filename, 'r+', 0)
+            else:
+                self.storage = open(filename, 'w+', 0)
+        except IOError, err:
+            sys.stderr("(ERROR) Failed to initialize offset storage (%s)\n"
+                       % self.__class__.__name__)
```
anastasiakaida: 'err' local variable is never used. Can it be removed?

anastasiakaida: [This](https://github.com/PanDAWMS/dkb/pull/93/commits/52a35decb5bad754eea69160f10cbddf8b45b0ab?diff=unified#diff-4c221825c8bb3980b06a244fe9499716R457) *if* statement looks repetitive. Maybe it would be better to create *def* like `phys_short_search(taskname, category)` and bind it with a dict of elements for checking.

I suppose, with new *def* we can check every 'taskname' in a loop using two or three strings (what if the number of task names for checking will expand in the future?).

Maybe it is not really urgent now to do this improvement. Just an assumption...

mgolosova: `err` was added just automaticaly, without second thought. But I think it would be better to output error, then remove varisble ;)

Replacing `if`s is also a good idea -- will do it right now, as I`m afraid I can simply lose it later...(

Thank you!

mgolosova: *Update*
Removed `err` -- as the exception is re-raised and the error message is to be output in the external code frame.

Just realized, that I just can`t replace `if` part here -- as it was moved to Stage 16 already.
Will do my best not to forget about it and address in the nearest future.

mgolosova:
History is rewritten again, to join "remove unused variable" with the commit where the variable appears.
Previous (approved) version of this PR is stored for reference in `pr-93-stage-009-02` branch.

@anastasiakaida,
thank you for the approve in advance. Unfortunately after pushing changes to the PR its status changes back to "Review required" (which is fair enough: who knows, what changes I have just pushed? ;) )
I promise I haven`t change anything but removing `err` variable, but please check :)

PR #94, opened 2017-12-11
Travis: update codestyle checks

mgolosova:
Change PEP8 codestyle checks.
Since now the only ignored by `pep8` (https://pypi.python.org/pypi/pep8) codes are:
* W503 (line break before binary operator):
  https://www.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator
* W292 (no newline in the end of file):
  as we work in different OSs, it might be a problem

If there are any other codes that you think must be ignored, please write it in the comments..

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_install:
     - pip install pep8
 
 script:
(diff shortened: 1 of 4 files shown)


mgolosova:
@IATkachenko ,
Unfortunately I couldn`t find how to handle "no newline at end of file" with Git itself.
I think the PR will be merged as is, but if later the way will be found -- we`ll adjust the Travis settings (as well as if/when we decide to add more PEP8 codes to the ignore list for the Travis checks).

PR #95, opened 2017-12-11
Travis: add Python syntax check.

mgolosova:
To run Python syntax check in parallel with PEP8 check, we added both
checks to the sama stage.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,21 +1,35 @@
-language: python
-python:
-    - 2.7
-
 notifications:
     email: false
 
-before_install:
-    - pip install pep8
+jobs:
+  include:
+    - stage: python_check
+      language: python
+      python: 2.7
+      before_install:
+        - pip install pep8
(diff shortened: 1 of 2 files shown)


mgolosova:
@IATkachenko , 
thank you for the guidlines.

I made changes basing on bb0f0a4 and the original check for PEP8 codestyle, but I have a question: wouldn`t it be OK to use something like
```
      script:
        - python -m py_compile `find . -name "*.py"`
```
instead of https://github.com/PanDAWMS/dkb/pull/95/files#diff-354f30a63fb0907d4ad57269548329e3R27-R35?

The command returns non-zero code if errors occured, and writes information about errors to STDERR.
Will it break something?

mgolosova:
@IATkachenko,
I see. Hardly we will ever have more than 2 million Python files in the repo, but you\`re right -- we`ll better keep an eye on such things.

PR #96, opened 2017-12-13
Travis: pep8 checks update

mgolosova:
As `pep8` is obsolete, we may switch to the recomended `pycodestyle`.
However `pycodestyle` is a bit more strict about some things (see the Travis checks for this PR).
Particularly, it suggests to:
* E741: avoid ambiguous variable names (like 'l', 'O', 'I');
* E722: avoid "bare except".

The first, I think, need no additional explaination. It is just a question of "shell we pretend there`s no ambiguity" or fix it now.

For the second I would refer to the PEP-8 itself:

    A bare "except:" clause will catch SystemExit and KeyboardInterrupt exceptions,
    making it harder to interrupt a program with Control-C, and can disguise other problems.
    If you want to catch all exceptions that signal program errors, use "except Exception:"
    (bare except is equivalent to "except BaseException:").

    A good rule of thumb is to limit use of bare 'except' clauses to two cases:

    1. If the exception handler will be printing out or logging the traceback;
       at least the user will be aware that an error has occurred.
    2. If the code needs to do some cleanup work, but then lets the exception
       propagate upwards with raise. try...finally can be a better way to handle this case.

In those cases `pycodestyle` catches none of these two variants is not the case.

I can fix both codestyle issues here (or authors could do it themselves, as I can ruin something by accident, especially in case of variable naming).
Or I can add these two codes to ignore list of Travis checks, if there are any reasons we do intend to ignore it.

---

@Evildoor, mostly it affects the Stage 030, so I would like to hear from you in the comments.

### DIFF
--- a/Utils/Dataflow/030_PDFAnalyzer/manager.py
+++ b/Utils/Dataflow/030_PDFAnalyzer/manager.py
@@ -546,71 +546,10 @@ class Paper:
     def delete(self):
         """ Delete all files associated with paper. """
         shutil.rmtree(self.dir)
-# def find_title(self):
-# """ New title determining method, does not works ideally yet.
-# Titles consisting of several lines are problematic to
-# determine.
-# """
-#        lines = self.get_xml_page(1)
-#
-#        d = {}
-# for l in lines:
-#            m = re_xml_symbol.match(l)
-# if m:
-#                size = float(m.group(1))
-#                text = m.group(2)
-# if size in d.keys():
-#                    d[size] += text
-# else:
-#                    d[size] = text
-# elif re_xml_empty_symbol.match(l):
-#                d[size] += " "
-#        xml_title = False
-# print d
-# while True:
-#            size = max(d.keys())
-#            valid = True
-# try:
-# d[size].decode()
-# except:
-#                valid = False
-# if not "atlas note" in d[size].lower() and valid:
-#                xml_title = d[size]
-# break
-# else:
-#                del d[size]
-#
-# print xml_title
-# if not xml_title:
-# return False
-#
-#        lines = self.get_txt_page(1)
-#        title = ""
-# for l in lines:
-# if len(l) <= 4 or l.startswith("Supporting Note")\
-# or l.startswith("ATLAS NOTE"):
-# continue
-#            words = l.split()
-#            i = 0
-# for w in words:
-# try:
-# This throws exception sometimes, something about
-# ascii codec unable to decode.
-#                    w_in = w in xml_title
-# except:
-#                    w_in = False
-# if len(w) > 1 and w_in:
-#                    i += 1
-# if i > 1 or (len(words) == 1 and i == 1):
-#                title += l.replace("\n", " ")
-# elif title:
-# break
-# return title
 
     def find_attributes_general(self):
         """ Find general attributes in a document. """
         attrs = {}
-#        attrs["title"] = self.find_title()
         text = self.get_text()
 
         attrs["campaigns"] = []
--- a/Utils/Dataflow/030_PDFAnalyzer/manager.py
+++ b/Utils/Dataflow/030_PDFAnalyzer/manager.py
@@ -67,8 +67,7 @@ def path_join(a, b):
 if __name__ == "__main__":
     try:
         import Tkinter
-        from tkFileDialog import askdirectory, askopenfilename,\
-            askopenfilenames, asksaveasfile
+        from tkFileDialog import askdirectory, askopenfilenames, asksaveasfile
         import tkMessageBox
     except Exception as e:
         sys.stderr.write("Exception while loading Tkinter: %s\n" % e)
--- a/Utils/Dataflow/030_PDFAnalyzer/manager.py
+++ b/Utils/Dataflow/030_PDFAnalyzer/manager.py
@@ -391,24 +391,24 @@ def process_diapason(d):
     return values
 
 
-def check_all_button(v, l):
+def check_all_button(v, lst):
     """ Command for handling Tkinter global checkbuttons.
 
     Command (un)checks all the checkbuttons in the list
     v - a VarInt variable associated with the global checkbutton.
-    l - a list of VarInt variables associated with checkbuttons in the
+    lst - a list of VarInt variables associated with checkbuttons in the
     list.
     """
     s = 0
-    for i in l:
+    for i in lst:
         s += i.get()
-    if s == len(l):
+    if s == len(lst):
         v.set(0)
-        for i in l:
+        for i in lst:
             i.set(0)
     else:
         v.set(1)
-        for i in l:
+        for i in lst:
             i.set(1)
 
 
@@ -933,8 +933,8 @@ class Manager:
                      self.show_paper_info(False, paper))
             b.grid(row=r, column=0)
             if p.title is not None:
-                l = Tkinter.Label(self.frame, text=p.title)
-                l.grid(row=r, column=1)
+                lbl = Tkinter.Label(self.frame, text=p.title)
+                lbl.grid(row=r, column=1)
             r += 1
 
         self.frame.update_idletasks()
@@ -960,29 +960,29 @@ class Manager:
 
         frame = Tkinter.Frame(w)
 
-        l = Tkinter.Label(frame, text="Working directory")
-        l.grid(row=0, column=0)
+        lbl = Tkinter.Label(frame, text="Working directory")
+        lbl.grid(row=0, column=0)
         e = Tkinter.Entry(frame, width=100, textvariable=work_dir)
         e.grid(row=0, column=1)
 
-        l = Tkinter.Label(frame, text="Determine papers' titles")
-        l.grid(row=1, column=0)
+        lbl = Tkinter.Label(frame, text="Determine papers' titles")
+        lbl.grid(row=1, column=0)
         b = Tkinter.Checkbutton(frame, variable=determine_title)
         b.grid(row=1, column=1)
 
-        l = Tkinter.Label(frame, text="Open intervals in text")
-        l.grid(row=2, column=0)
+        lbl = Tkinter.Label(frame, text="Open intervals in text")
+        lbl.grid(row=2, column=0)
         b = Tkinter.Checkbutton(frame, variable=open_intervals_text)
         b.grid(row=2, column=1)
 
-        l = Tkinter.Label(frame, text="Open intervals in tables")
-        l.grid(row=3, column=0)
+        lbl = Tkinter.Label(frame, text="Open intervals in tables")
+        lbl.grid(row=3, column=0)
         b = Tkinter.Checkbutton(frame, variable=open_intervals_tables)
         b.grid(row=3, column=1)
 
         txt = "Extract dataset IDs instead of full tables"
-        l = Tkinter.Label(frame, text=txt)
-        l.grid(row=4, column=0)
+        lbl = Tkinter.Label(frame, text=txt)
+        lbl.grid(row=4, column=0)
         b = Tkinter.Checkbutton(frame, variable=tables_ids_only)
         b.grid(row=4, column=1)
 
@@ -1120,8 +1120,8 @@ class Manager:
             for c in window.winfo_children():
                 c.destroy()
         window.title("Info: %s" % paper.fname)
-        l = Tkinter.Label(window, text="File name: %s" % paper.fname)
-        l.grid(row=0, column=1)
+        lbl = Tkinter.Label(window, text="File name: %s" % paper.fname)
+        lbl.grid(row=0, column=1)
         b = Tkinter.Button(window, text="Save",
                            command=lambda paper=paper: self.save_paper(paper))
         b.grid(row=0, column=2)
@@ -1135,13 +1135,13 @@ class Manager:
         b = Tkinter.Button(window, text="Delete", command=lambda window=window,
                            paper=paper: self.delete_paper(window, paper))
         b.grid(row=0, column=5)
-        l = Tkinter.Label(window, text="Title: %s" % paper.title)
-        l.grid(row=1, columnspan=5)
-        l = Tkinter.Label(window, text="Pages: %d" % paper.num_pages)
-        l.grid(row=2, columnspan=5)
-        l = Tkinter.Label(window,
-                          text="Rotated pages: %s" % str(paper.rotated_pages))
-        l.grid(row=3, columnspan=5)
+        lbl = Tkinter.Label(window, text="Title: %s" % paper.title)
+        lbl.grid(row=1, columnspan=5)
+        lbl = Tkinter.Label(window, text="Pages: %d" % paper.num_pages)
+        lbl.grid(row=2, columnspan=5)
+        lbl = Tkinter.Label(window, text="Rotated pages: %s" %
+                            str(paper.rotated_pages))
+        lbl.grid(row=3, columnspan=5)
         b = Tkinter.Button(window, text="Attributes",
                            command=lambda window=window,
                            paper=paper: self.show_paper_attributes(window,
@@ -1187,8 +1187,8 @@ class Manager:
         d = {}
         r = re.compile("^<text[^>]+ size=\"([0-9.]+)\">(.+)</text>$")
         empty = re.compile("^<text> </text>$")
-        for l in lines:
-            m = r.match(l)
+        for line in lines:
+            m = r.match(line)
             if m:
                 size = m.group(1)
                 text = m.group(2)
@@ -1196,12 +1196,12 @@ class Manager:
                     d[size] += text
                 else:
                     d[size] = text
-            elif empty.match(l):
+            elif empty.match(line):
                 d[size] += " "
 
         msg = "Please, select a string which looks like a title of the paper"
-        l = Tkinter.Label(window, text=msg, font=HEADING_FONT)
-        l.grid(row=0, columnspan=2)
+        lbl = Tkinter.Label(window, text=msg, font=HEADING_FONT)
+        lbl.grid(row=0, columnspan=2)
         r = 1
         selection = Tkinter.StringVar()
         for key in d:
@@ -1237,11 +1237,11 @@ class Manager:
         possible_title = possible_title_variable.get()
 
         title = ""
-        for l in lines:
-            if len(l) <= 4 or l.startswith("Supporting Note")\
-               or l.startswith("ATLAS NOTE"):
+        for line in lines:
+            if len(line) <= 4 or line.startswith("Supporting Note")\
+               or line.startswith("ATLAS NOTE"):
                 continue
-            words = l.split()
+            words = line.split()
             i = 0
             for w in words:
                 try:
@@ -1253,12 +1253,13 @@ class Manager:
                 if len(w) > 1 and w_in:
                     i += 1
             if i > 1 or (len(words) == 1 and i == 1):
-                title += l.replace("\n", " ")
+                title += line.replace("\n", " ")
             elif title:
                 break
-        l = Tkinter.Label(window, text="Please, correct the title, if needed.",
-                          font=HEADING_FONT)
-        l.grid(row=0, columnspan=2)
+        lbl = Tkinter.Label(window,
+                            text="Please, correct the title, if needed.",
+                            font=HEADING_FONT)
+        lbl.grid(row=0, columnspan=2)
         e = Tkinter.Entry(window, width=150)
         e.insert(0, title)
         e.grid(row=1, columnspan=2)
@@ -1328,22 +1329,22 @@ class Manager:
             for a in paper.attributes_general:
                 if a == "links":
                     if paper.__dict__[a]:
-                        l = Tkinter.Label(window, text="Links:")
-                        l.grid(row=r, column=0)
+                        lbl = Tkinter.Label(window, text="Links:")
+                        lbl.grid(row=r, column=0)
                         r += 1
                         for key in paper.__dict__[a]:
                             txt = "%s %s" % (key, paper.__dict__[a][key])
-                            l = Tkinter.Label(window, text=txt)
-                            l.grid(row=r, column=0)
+                            lbl = Tkinter.Label(window, text=txt)
+                            lbl.grid(row=r, column=0)
                             r += 1
                     else:
-                        l = Tkinter.Label(window, text="No links")
-                        l.grid(row=r, column=0)
+                        lbl = Tkinter.Label(window, text="No links")
+                        lbl.grid(row=r, column=0)
                         r += 1
                 else:
                     txt = "%s:%s" % (a, str(paper.__dict__[a]))
-                    l = Tkinter.Label(window, text=txt)
-                    l.grid(row=r, column=0)
+                    lbl = Tkinter.Label(window, text=txt)
+                    lbl.grid(row=r, column=0)
                     r += 1
             b = Tkinter.Button(window, text="Back",
                                command=lambda window=window,
@@ -1369,8 +1370,8 @@ class Manager:
                 r = 0
                 dataset_entries = {}
                 for c in datasets:
-                    l = Tkinter.Label(frame, text=c, font=HEADING_FONT)
-                    l.grid(row=r, column=0, columnspan=2)
+                    lbl = Tkinter.Label(frame, text=c, font=HEADING_FONT)
+                    lbl.grid(row=r, column=0, columnspan=2)
                     c_s = Tkinter.IntVar()
                     c_s.set(1)
                     check_category_b = Tkinter.Checkbutton(frame, var=c_s)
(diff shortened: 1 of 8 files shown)


Evildoor:
I skimmed through Travis checks and got some ideas, but i would like to educate myself more thoroughly on this matter before i say anything. How urgent is this?

mgolosova:
@Evildoor,
Thank you for the fast response!
There`s no urgency, actually -- for now it`s just a matter of noise in the Travis check logs, and does not affect anything else.
I think, if you could do it in a week or two, it would be absolutely perfect -- but if for some reason it will take longer, no problem as well.

Evildoor:
@mgolosova 
So, if i understand correctly, we want to move from tool pep8 to tool pycodestyle because the latter is pretty much a renamed and updated former. If yes, then i have no objections.

Regarding two issues:
E741
If this issue means that in some cases symbols "letter l", "letter i" and "digit one" can be written in similar way and hard to distinguish (same for "letter o" and "digit zero") - well, i've never run into such problem yet, but i admit that it is possible to in the future, so i can fix this just in case.
E722
This issue makes sense, and i will fix it.

Evildoor:
#97. Note - it seems that pycodestyle lacks cycle checks when searching for E741. It finds issue in code like:
l = 1
print l

But not in:
for l in range(1, 10):
(INDENT)print l

mgolosova:
All the changes related to the issue are done.
Please review :)

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1486,9 +1487,9 @@ def show_paper_datasets(self, window, paper):
                                             datasets)
         else:
             if not paper.datasets:
-                l = Tkinter.Label(window, text="No datasets found",
-                                  font=HEADING_FONT)
-                l.grid(row=0)
+                lbl = Tkinter.Label(window, text="No datasets found",
+                                    font=HEADING_FONT)
```
konvpalto: This whitespace-only change should go in its own.

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1601,9 +1603,9 @@ def show_paper_datatables(self, window, paper):
                                             datatables)
         else:
             if not paper.datatables:
-                l = Tkinter.Label(window, text="No datatables found",
-                                  font=HEADING_FONT)
-                l.grid(row=0)
+                lbl = Tkinter.Label(window, text="No datatables found",
+                                    font=HEADING_FONT)
```
konvpalto: This whitespace-only change doesn't belong to this commit.

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1663,9 +1666,9 @@ def show_paper_page_tables(self, window, paper, e=False):
             for c in window.winfo_children():
                 c.destroy()
             window.title("Select page")
-            l = Tkinter.Label(window, text="Page number(1 - %d):"
-                              % paper.num_pages)
-            l.grid(row=0, column=0)
+            lbl = Tkinter.Label(window, text="Page number(1 - %d):"
+                                % paper.num_pages)
```
konvpalto: Whitespace-only: move to another commit.

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1692,23 +1695,23 @@ def show_paper_page_tables(self, window, paper, e=False):
                 tables = xmltable.get_tables_from_text(text)
                 for table_num in range(0, len(tables)):
                     frame = Tkinter.Frame(window)
-                    l = Tkinter.Label(frame, text="Table %d" % table_num)
-                    l.grid(row=0, column=0,
-                           columnspan=len(tables[table_num].rows[0]))
+                    lbl = Tkinter.Label(frame, text="Table %d" % table_num)
+                    lbl.grid(row=0, column=0,
+                             columnspan=len(tables[table_num].rows[0]))
```
konvpalto: Whitespace-only: belongs to elsewhere.

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1724,9 +1727,9 @@ def show_paper_visual(self, window, paper, e=False):
             for c in window.winfo_children():
                 c.destroy()
             window.title("Select page")
-            l = Tkinter.Label(window, text="Page number(1 - %d):"
-                              % paper.num_pages)
-            l.grid(row=0, column=0)
+            lbl = Tkinter.Label(window, text="Page number(1 - %d):"
+                                % paper.num_pages)
```
konvpalto: Whitespace-only: you should get the idea already.

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1769,9 +1772,10 @@ def show_paper_visual(self, window, paper, e=False):
                     else:
                         header_row = False
                         color = "black"
-                    for l in row:
-                        cnvs.create_rectangle((l.left, l.top + 10, l.right,
-                                               l.bottom + 10), outline=color)
+                    for line in row:
+                        cnvs.create_rectangle((line.left, line.top + 10,
+                                               line.right, line.bottom + 10),
+                                              outline=color)
```
konvpalto: Align this one properly.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
```diff
@@ -138,8 +138,8 @@ def merge(self, line):
 def row_centery(row):
     """ Calculate average y-center in a row. """
     c = 0
-    for l in row:
-        c += l.center[1]
+    for line in row:
```
konvpalto: May be "cell" will be better than "line" here?

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
```diff
@@ -247,9 +244,9 @@ def construct_row(self, line, used_lines):
         Lines which were already used are not used again.
         """
         row = [line]
-        for l in self.lines:
-            if l != line and l not in used_lines and line.same_row(l):
-                row.append(l)
+        for ln in self.lines:
```
konvpalto: I'd rename "line" from the function parameters to "first_line" and use "line" as the replacement for "l" (or the new "ln"): makes it more obvious on what's the meaning of the parameter and avoids confusion with iteration variable.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
```diff
@@ -260,8 +257,8 @@ def row_text(self, num=None, row=False):
         if num is not None or num == 0:
             row = self.rows[num]
         text = ""
-        for l in row:
-            text += l.text + "!"
+        for line in row:
```
konvpalto: For this we have map + join: return ("!".join(map(lambda r: r.text), row) + "!")

Not sure, by the way, that the trailing '!' is needed: it contradicts with the function description, since it says "between lines".

konvpalto:
Don't think that blindly changing "except e" to "except Exception:" is a good thing: it will still catch SyntaxError, see exception hierarchy: https://docs.python.org/3/library/exceptions.html#exception-hierarchy.

And that's not what people typically really want: it will propagate syntax errors as exceptions (or just silently mask them) instead of bailing out during the compilation time.

konvpalto:
My current opinion is:
 * 6624a8273aae535c569284c4d74286b55c0cd912: fine
 * 0730808790743751d0407c3251167637957948ca: fine
 * e4d60fad1ccf01f85558c9b8797e771adab69bcd and
 * 5e58375c937548aa5992e141997aeaadb558b3a5: merge together
 * 5ba7a8f73bff4cb2c0425774e30840259adf66d0: has comments
 * f10530f5b9dacbff34f3e91a8bf87eaf25e16e04: expand commit log
 * 49e680356d4fb361c67079a2c1715e7fe2fad904: doesn't belong to this topic
 * 5886bf3e50d7322959be80b4d409c77ea493d95e: may need further polishing of catched exception types
 * df41dc839d2843b40bedfad19eb56971583a31f1: fine

Utils/Dataflow/030_PDFAnalyzer/manager.py
mgolosova: @konvpalto,
not sure. Why not to fix indentation, broken due to the change of the variable name, in the same commit where the variable name was changed?
I mean, if it was a standalone codestyle fix -- you`re right. But if we try to follow the codestyle, it looks OK for me to make changes aimed **not to break** the codestyle along with the changes that provoke the anomaly... no?

The logic of "whitespace-only" commits is understood, but is it *really* applicable for this case?

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1618,27 +1620,28 @@ def show_paper_datatables(self, window, paper):
                 for k in keys:
                     (header, data) = paper.datatables[k]
                     t_frame = Tkinter.Frame(frame)
-                    l = Tkinter.Label(t_frame, text=header, font=HEADING_FONT)
+                    lbl = Tkinter.Label(t_frame, text=header,
+                                        font=HEADING_FONT)
```
mgolosova: @konvpalto,
and talking about "whitespace-only change does not belong here" -- then does the line breaking one belong?

Utils/Dataflow/030_PDFAnalyzer/manager.py
mgolosova: @konvpalto,
however strange it might look, it *does* be aligned properly.
The `outline=color` is aligned to the first symbol after `create_rectangle()` opening bracket; previous line is aligned to the one of the first argument tuple. 

Moving the last line to the right would lead to the `pycodestyle` alarm:
```
E127 continuation line over-indented for visual indent
    Continuation lines indentation.

    Continuation lines should align wrapped elements either vertically
    using Python's implicit line joining inside parentheses, brackets
    and braces, or using a hanging indent.

    When using a hanging indent these considerations should be applied:
    - there should be no arguments on the first line, and
    - further indentation should be used to clearly distinguish itself as a
      continuation line.
```

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
mgolosova: @Evildoor,
please consider.

Utils/Dataflow/030_PDFAnalyzer/manager.py
Evildoor: I agree with @mgolosova on this one. Whitespace changes could have been moved into a separate commit if, for example, the code was literally peppered with whitespace-related problems, so it would have been reasonable to fix other things while ignoring them, and then fix the whitespace in a separate commit. This is not the case here, here we pick a chunk of code with 1 problem, and transform it into a chunk of code with 0 problems. I see no reason to explicitly state in our commit history that it was achieved by performing not 1 but 2 changes, both of which are cosmetic and bear no real impact on program logic.

Same for other whitespace and linebreaking comments.

Utils/Dataflow/030_PDFAnalyzer/manager.py
Evildoor: @mgolosova's explanation leaves me no room for comments. I can only add a side thought that sometimes pycodestyle's indentation logic looks weird, but not here.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
Evildoor: This sounds like a good idea to me, thanks.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
Evildoor: I don't think so. We had our share of trouble with naming elements of tables in this code, and for now settled on table consisting of rows, and row consisting of lines. This function correlates with the code by using these names. 

If "line" is a bad name, then it should be changed in the whole code, not just here. "Cell" does not sounds right to me because our lines do not correspond to cells in original tables in PDF documents.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
Evildoor: It seems that row should be the second argument of map in your example. I think i'll use [A for B in C] instead of map, since that construct is more familiar to me.

Comment about trailing "!" makes sense, thanks.

mgolosova:
@konvpalto,
thank you for your comments.

> * e4d60fa and
> * 5e58375: merge together

Did you mean to join commit e4d60fa (Removed unused function related to E741) with 5ba7a8f (Fixed E741 - ambiguous variable names.) -- not 5e58375 (Fixed E722)?

 About this:
> f10530f: expand commit log

Not sure I got the point: add to the log the name of the removed module? or explain why and how it became unused (which is out of my scope, but I can try to restore it from history)?


>Don't think that blindly changing "except e" to "except Exception:" is a good thing: it will still catch SyntaxError from eval and NameError/ValueError, see exception hierarchy: https://docs.python.org/3/library/exceptions.html#exception-hierarchy.
>
>And that's not what people typically really want: it will propagate syntax errors as exceptions (or just silently mask them) instead of bailing out during the compilation time.

Whatever is caught by try/except, happens on the runtime; compilation time errors won\`t be affected (as you said, even `SyntaxError` will be affected only in case of `eval()`). Compilation during `eval()`, for me, sounds pretty much like "runtime error" -- if one needs to use `eval()`, most probably he wants to evaluate some dynamic statement. And this syntax error in wrongly constructed statement is neither better nor worse than any other error, so why not to catch it same way ;)

But `NameError` is worse in this sense, you are right.

I\`ll take a look if I can do something about it, but not sure if I will be able to list all possible exceptions author meant to catch in every particular case... so it was not "blind", it was maximum effect with a minimum of efforts ;)

Evildoor:
### Merging of e4d60fa
Not sure if this is necessary. If i understand correctly, then the picture is as following:

- 5e58375 fixes E722.
- 5ba7a8f fixes E741.
- e4d60fa removes obsolete commented out function which contains both E741 (as i've written in commit message) and E722 (as @konvpalto correctly pointed out).

So i don't see how merging e4d60fa with any of these two commits will improve the situation, since it is related to both of them, but 5e58375 and 5ba7a8f should not be merged together because they deal with separate problems. It is possible to perform a significant amount of sticking to the rules by: 
1) Uncommenting the function.
2) Making it work (which is unnecessary).
3) Fixing E722 and E741 and adding these fixes to two first commits.
4) Deleting the function in a separate commit which would not even fit into this pull request, same goes for commit which will perform 1 and 2.

But i doubt if it is wise. Deleting a commented function in a separate commit e4d60fa looks like a best option here to me.

### Changing "except" to "except Exception"

If i understand correctly, then "except Exception" is a step forward, since it does not interferes with SystemExit and KeyboardInterrupt. True, it may still have problems like @konvpalto has explained, but at the very least it brings no harm compared to bare "except", so why not use it? If there are suggestions with better ways of handling these exceptions, then i'd like to hear them.

### 49e6803: doesn't belong to this topic

This change was performed after @anastasiakaida has discovered that a variable is not used anywhere with flake8 (see #97). I've searched a bit, and according to https://pypi.python.org/pypi/flake8:

>Flake8 is a wrapper around these tools:
>
>PyFlakes
>pycodestyle
>Ned Batchelder’s McCabe script

Since i've not found this problem with pycodestyle, it seems that one of the two remaining components raised it. So, we can either leave things as they are since this change is also related to coding style, or decide that this pull request deals strictly with pycodestyle, so removing obsolete variable must be performed elsewhere. Which is it?

### f10530f: expand commit log

I agree with @mgolosova - what's the problem? If this
> Care to specify what pycodestyle error/warning it belongs to?

is about f10530f, then it's also flake8's doing, so same thoughts as with 49e6803 above.
> (which is out of my scope, but I can try to restore it from history)

No need. I'll write the explanation if we'll decide that it's necessary.

### Fixing things

@mgolosova, how should we proceed here? There are some things which require rewriting history, and i have no problem with doing so in my branches, but this one is not, and i don't want to mess up others' work. Should i clone the branch, rewrite my code in new one, and then we'll merge it here?

mgolosova:
@Evildoor,

> how should we proceed here? There are some things which require rewriting history, and i have no problem with doing so in my branches, but this one is not, and i don't want to mess up others' work. Should i clone the branch, rewrite my code in new one, and then we'll merge it here?

I think you`re right, cloning the branch and rewriting your commits would be OK.
As soon as you`re done, I`ll cherry-pick them to this branch and replace current commits with new ones.

As a result we will lose information about PR #97 in the history, but I think it`s a fair price for more detailed discussion we have here.

Utils/Dataflow/030_PDFAnalyzer/manager.py
konvpalto: My only wish there was to split the whitespace-only part into the second commit to allow for easier review: "look, ma, just whitespace stuff here".  It makes one to be sure what to check in the commit.

Besides: what's the commit message here? Bare "Fixed E741 - ambiguous variable names"?  Let me pretend to review this loudly:
 - Hmm, no mentions about whitespace changes?  Did they really wanted that?  Typo?
 - OK, let me try to study, may be that was an alignment problem (tab/space mix)?  No?
 - No, just whitespaces.  OK, probably, that what they also meant.

In reality, there are two questions.  The first one: "can this commit be improved?".  The answer is "yes" and I can bet for my life on that.  Because (taking "consistency" as the correspondence between various parts of an object and the fact that commit log and commit contents are *equally important* parts of the object /that is commit/) your commit is inconsistent, I am awfully sorry to say it.

And I see at least two ways to improve:
 - extend commit log;
 - split into two commits.
May be there are other ones.

The second question is "will those changes worth that/justify the time spent/etc?".  I dunno.  And, honestly, I don't care much: improvement for me is the justification on its own. For you -- you tell me.


And there is the third question, specifically to you, Vasily: why are you asking me your questions with "I agree with ..." stanza?  "Peppered" is a good and strong word, it, may I assume, provides the said justification: "oh, it's peppered, let's now unpepper that, it adds The Real Value (tm)".  What's the difference between one grain of pepper and (42-1) grans?  Just a loud thought, can be skipped.

But splitting commits is easy: so, what are you really trying to say or ask about?  Or, more explicitely: are you really trying to understand?  I'd just split it: no additional load on the repository, definitely, easier to review, easier to skim through in the future.  I used to answer such questions "Really Smart Boys can cope with such commits easily, whazzup?".  Yep, they do.  What's your idea here?  Can be moved to e-mail (or skipped: won't be rude by any means).


It is not a showstopper by itself, but, please, remember: the devil lies in the details. And, it seems to me, most of the time -- only in them.

---
And the last thing: it is just my opinion, but I don't currently see any arguments not to do that.  Everyone is free to skip this, do that, adopt or ask again.  The only thing that I'd forbid (if I could) is to make this a showstopper to the evolution (of the code).

Same for other comments from /me ;)

Utils/Dataflow/030_PDFAnalyzer/manager.py
konvpalto: Yep, I was wrong here, thanks for revealing my mistake!

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
konvpalto: If the rest of the code is to be fixed with respect to the naming -- it should be fixed.  If it shouldn't -- OK, I was wrong and I apologize.  If you'd show me what you mean by "our lines do not correspond to cells in original table in PDF documents" (simple link to the PDF document and few words explaining where to look are enough) -- I'd be rather grateful.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
konvpalto: return ("!".join(map(lambda r: r.text, row)) + "!"), exactly.

Utils/Dataflow/030_PDFAnalyzer/manager.py
konvpalto: It is about "Whitespace-only: belongs to elsewhere", yes?  My wording is ambigious (and, thus, bad): what I meant is "belongs to another commit", not the "another pull".

Whitespace-only fixes with not functional changes are OK to go along with anything, especially if they correspond to the same lines of code (or touch some of them) that were fixing the functionality/code-style related ones, in my view.  They just should be marked as such, nothing more.

konvpalto:
Yep,  I meant merging e4d60fa and 5ba7a8f, sorry.

For f10530f5 that was "Care to specify what pycodestyle error/warning it belongs to?" I was referring to.  Seems like I should stop doing fancy stuff in the boxes near the comments and just revert to the plain text + diff until I'll learn how to properly use boxes.  Will do.

konvpalto:
> But i doubt if it is wise. Deleting a commented function in a separate commit e4d60fa looks like a best option here to me.

No problems: just order it differently: deletion first, fix for E741 -- after it.

Evildoor:
>>But i doubt if it is wise. Deleting a commented function in a separate commit e4d60fa looks like a >>best option here to me.
>
>No problems: just order it differently: deletion first, fix for E741 -- after it.

Good idea, will do.

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
Evildoor: Will answer via email.

Utils/Dataflow/030_PDFAnalyzer/manager.py
Evildoor: @konvpalto 
I was stating my current position regarding this point and trying to explain my logic behind it, not asking questions.

>"Peppered" is a good and strong word, it, may I assume, provides the said justification: "oh, it's peppered, let's now unpepper that, it adds The Real Value (tm)". What's the difference between one grain of pepper and (42-1) grans? 

The difference in this case is amount of additional work created by task of removing pepper from somewhere. Removing one grain is pretty easy, while large amount of pepper may require the worker to get an empty sack to store removed pepper in it, and then dispose of the sack filled with pepper. You want the worker to get a sack for one grain as well. I don't insist on this comparison being 100% accurate/complete/whatever and using it to decide things in our problem, merely illustrating my point.

>Besides: what's the commit message here? Bare "Fixed E741 - ambiguous variable names"? Let me pretend to review this loudly:
>- Hmm, no mentions about whitespace changes? Did they really wanted that? Typo?
>- OK, let me try to study, may be that was an alignment problem (tab/space mix)? No?
>- No, just whitespaces. OK, probably, that what they also meant.

Ok, let me explain my logic as well. "Fixed E741, then fixed whitespace problem caused by doing so". Hmm, why haven't they told me how they powered up their PC before doing so? True, doing so does not affects the code (unlike your example), but it's pretty much the same to me, because i think that in most cases solutions for problems caused by fixes should be included in these fixes. Especially here, since the secondary problem does exactly the same thing as the main one - breaks the code style.

konvpalto: > I was stating my current position regarding this point and trying to explain my logic behind it, not asking questions.

Then I am not interested in continuing this topic in its current state: I had already said everything I wanted to.  The additional work from splitting the commit into two is negligible, but if we're making two long comments trying to pursue (not get any new knowledge, just pursue) each other -- it doesn't worth it.  My point is that review-wise splitting is good (not for me, by the way: I am technically fine with commits that have more problems that this one); your point is that adding (helpful, but "obvious") information into the commit log is pointless.

My experience in digging up the code in various open-source repositories many times when it was needed or even been urgent tells me that you're wrong.  Perhaps you're doing more such digging than me and, at the same time, still find your point to be right; then, most probably, my technicall skills are miserable here.  If not -- you'll start to work with other's code, you'll learn.  Or not.

Dixi.

konvpalto: And, regarding this splitting issue: everyone may pretend that it can be ignored and do that.

mgolosova:
Rewrote the PR history:
* rebased on the current version of `master`;
* "bare `except`" from `except Exception` became more situation-specific [1];
* Stage30: separate commit with removal of commented pieces of code before codestyle-fix commits;
* Stage30: improved log message about obsolete import;
* Stage30: improved log message about renaming variables [3].

@konvpalto, @evildoor, could you check if I missed something from the discussion above and something (not mentioned in this comment) can be improved yet?

---
[1] With the exception of `030_PDFAnalyzer/manager.py`, as there is a lot of `except Exception` statements, and if improve it, we need to improve it stage-wide [2], not in a single line -- and it is beyond scope of the given PR.
[2] ...or better repository-wide, as the same can be seen in the code of many stages.
[3] We have not splitted commit that fixes ambiguous variable names into "renaming variables" and "fixing indentation broken during variable renaming", as I still not convinced that it would be a better solution (I am not pretending I have heard/read nothing; I am just not convinced with it). Perhaps it is just a lack of experience, but I`m working towards this goal and, keeping these arguments in memory, will try to understand the question better.

Evildoor:
I took a quick look, if there are problems here then I failed to notice them.

mgolosova:
@konvpalto, @evildoor,
is there any reason not to approve this requiest?

mgolosova:
@Evildoor,
the thing is that we can not proceed unless someone adds approving review to this PR. I can not be that someone, as the PR is created by me ;) I thought anyone having write access to the repository has permission to make a review, but I am not sure... if you don`t have one, tell me please and I'll see if I can do something. If you have, please use it ;)

mgolosova:
@Evildoor, thank you!
Merged.

PR #97, opened 2017-12-15
Stage 030 pycodestyle

Evildoor:
Fixed two issues discovered by pycodestyle.

### DIFF
--- a/Utils/Dataflow/030_PDFAnalyzer/manager.py
+++ b/Utils/Dataflow/030_PDFAnalyzer/manager.py
@@ -546,71 +546,10 @@ class Paper:
     def delete(self):
         """ Delete all files associated with paper. """
         shutil.rmtree(self.dir)
-# def find_title(self):
-# """ New title determining method, does not works ideally yet.
-# Titles consisting of several lines are problematic to
-# determine.
-# """
-#        lines = self.get_xml_page(1)
-#
-#        d = {}
-# for l in lines:
-#            m = re_xml_symbol.match(l)
-# if m:
-#                size = float(m.group(1))
-#                text = m.group(2)
-# if size in d.keys():
-#                    d[size] += text
-# else:
-#                    d[size] = text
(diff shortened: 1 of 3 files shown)


mgolosova:
Thank you!
The request will be reviewed till the end of tomorrow.

Evildoor:
Thanks for finding those cases, will check them later this week.

anastasiakaida:
@Evildoor,

Unfortunately, I've found some errors with flake8. I think it would be great to fix it:
`manager.py:70:9: F401 'tkFileDialog.askopenfilename' imported but unused
`
In `manager.py` you use _askopenfilenames_ so it would be nice to exclude _askopenfilename_ if you are not going to use it next time.

`manager.py:1273:21: F841 local variable 'bt' is assigned to but never used`
For what purpose `bt` was created if it is unused?
___________
Also, I have some questions for better understanding how this module works. I'll write it tomorrow.

Thank you!

Evildoor:
@anastasiakaida 

askopenfilename - at some point it was used, then all instances of it were removed or replaced with askopenfilenames.

bt - i don't remember exactly since this function (half-interactive title determination) was created long time ago and it's not used because we have better methods of obtaining paper title (outside of PDF Analyzer). By the looks of the code, it seems that it was used for shortening long possible titles, which is no longer necessary since Tkinter displays them just fine.

Thanks for bringing these problems to my attention, i will fix them.

Sure, i'm open to questions regarding this module. However, unless they are related to coding style discussed here, they should be asked elsewhere (via email, for example).

mgolosova:
I am really sorry for the delay. Seems like it is totally OK and can be merged.

My question is: may I merge this PR into #96, so that all the related changes will get to master with one merge (maybe a bit later, but all at once)?

Evildoor:
No worries, this is not urgent.

Sure, go ahead.

mgolosova:
@Evildoor,
thank you. The request is merged.

Please remove the branch if it is no longer needed.

PR #98, opened 2017-12-18
pyDKB: EOM/EOP markers

anastasiakaida:
- [x] Implement EOM/EOP shortcuts.
- [x] Implement custom markers from the console.
- [x] Mention the shortcut table in the `--mode` parameter description or `epilog`.
- [x] Solve buf-size issue.

### DIFF
--- a/Utils/Dataflow/pyDKB/common/__init__.py
+++ b/Utils/Dataflow/pyDKB/common/__init__.py
@@ -5,4 +5,5 @@ Common modules.
 from exceptions import *
 import hdfs
 import json_utils as json
+import custom_readline
 from Type import Type
--- a/Utils/Dataflow/pyDKB/common/custom_readline.py
+++ b/Utils/Dataflow/pyDKB/common/custom_readline.py
@@ -4,10 +4,16 @@ import fcntl
 
 
 def custom_readline(f, newline):
-    """Custom readline() function. It separates content from a text file 'f'
+    """Custom readline() function.
+
+    Custom_readline() separates content from a text file 'f'
     by delimiter 'newline' to distinct messages.
     The last line can be incomplete, if the input data flow is interrupted
     in the middle of data writing.
+
+    Keyword arguments:
+    f -- file/stream to read
+    newline -- custom delimiter
     """
     poller = select.poll()
     poller.register(f, select.POLLIN)
--- a/Utils/Dataflow/pyDKB/common/custom_readline.py
+++ b/Utils/Dataflow/pyDKB/common/custom_readline.py
@@ -1,0 +1,13 @@
+def custom_readline(f, newline):
+    buf = ""
+    max_buf_size = 1
+    while True:
+        while newline in buf:
+            pos = buf.index(newline)
+            yield buf[:pos]
+            buf = buf[pos + len(newline):]
+        chunk = f.read(max_buf_size)
+        if not chunk:
+            yield buf
+            break
+        buf += chunk
--- a/Utils/Dataflow/pyDKB/common/custom_readline.py
+++ b/Utils/Dataflow/pyDKB/common/custom_readline.py
@@ -1,4 +1,9 @@
 def custom_readline(f, newline):
+    """Custom readline() function. It separates content from a text file 'f'
+    by delimiter 'newline' to distinct messages.
+    The last line can be incomplete, if the input data flow is interrupted
+    in the middle of data writing.
+    """
     buf = ""
     max_buf_size = 1
     while True:
--- a/Utils/Dataflow/pyDKB/common/custom_readline.py
+++ b/Utils/Dataflow/pyDKB/common/custom_readline.py
@@ -1,12 +1,6 @@
 import select
-import sys
 import os
 import fcntl
-
-poller = select.poll()
-poller.register(sys.stdin, select.POLLIN)
-flags = fcntl.fcntl(sys.stdin.fileno(), fcntl.F_GETFL)
-fcntl.fcntl(sys.stdin.fileno(), fcntl.F_SETFL, flags | os.O_NONBLOCK)
 
 
 def custom_readline(f, newline):
@@ -15,6 +9,10 @@ def custom_readline(f, newline):
     The last line can be incomplete, if the input data flow is interrupted
     in the middle of data writing.
     """
+    poller = select.poll()
+    poller.register(f, select.POLLIN)
+    flags = fcntl.fcntl(f.fileno(), fcntl.F_GETFL)
+    fcntl.fcntl(f.fileno(), fcntl.F_SETFL, flags | os.O_NONBLOCK)
     buf = ""
     while True:
         if poller.poll(500):
--- a/Utils/Dataflow/pyDKB/common/custom_readline.py
+++ b/Utils/Dataflow/pyDKB/common/custom_readline.py
@@ -1,3 +1,14 @@
+import select
+import sys
+import os
+import fcntl
+
+poller = select.poll()
+poller.register(sys.stdin, select.POLLIN)
+flags = fcntl.fcntl(sys.stdin.fileno(), fcntl.F_GETFL)
+fcntl.fcntl(sys.stdin.fileno(), fcntl.F_SETFL, flags | os.O_NONBLOCK)
+
+
 def custom_readline(f, newline):
     """Custom readline() function. It separates content from a text file 'f'
     by delimiter 'newline' to distinct messages.
@@ -5,14 +16,14 @@ def custom_readline(f, newline):
     in the middle of data writing.
     """
     buf = ""
-    max_buf_size = 1
     while True:
+        if poller.poll(500):
+            chunk = f.read()
+            if not chunk:
+                yield buf
+                break
+            buf += chunk
         while newline in buf:
             pos = buf.index(newline)
             yield buf[:pos]
             buf = buf[pos + len(newline):]
-        chunk = f.read(max_buf_size)
-        if not chunk:
-            yield buf
-            break
-        buf += chunk
--- a/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
+++ b/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
@@ -313,8 +313,7 @@ class AbstractProcessorStage(AbstractStage):
             for i, line in enumerate(iterator):
                 yield self.parseMessage(line)
         except KeyboardInterrupt:
-            sys.stderr.write("(WARN) Input stream processing interrupt"
-                             " on line %s\n" % i)
+            sys.exit()
 
     def file_input(self, fd):
         """ Generator for input messages.
--- a/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
+++ b/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
@@ -313,6 +313,7 @@ class AbstractProcessorStage(AbstractStage):
             for i, line in enumerate(iterator):
                 yield self.parseMessage(line)
         except KeyboardInterrupt:
+            sys.stderr.write("(INFO) Interrupted by user.\n")
             sys.exit()
 
     def file_input(self, fd):
--- a/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
+++ b/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
@@ -309,8 +309,12 @@ class AbstractProcessorStage(AbstractStage):
             iterator = iter(fd.readline, "")
         else:
             iterator = custom_readline(fd, self.ARGS.eom)
-        for line in iterator:
+        try:
+            for i, line in enumerate(iterator):
                 yield self.parseMessage(line)
+        except KeyboardInterrupt:
+            sys.stderr.write("(WARN) Input stream processing interrupt"
+                             " on line %s\n" % i)
 
     def file_input(self, fd):
         """ Generator for input messages.
--- a/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
+++ b/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
@@ -49,6 +49,7 @@ from . import messageType
 from . import Message
 from pyDKB.dataflow import DataflowException
 from pyDKB.common import hdfs
+from pyDKB.common.custom_readline import custom_readline
 
 
 class AbstractProcessorStage(AbstractStage):
@@ -309,8 +310,8 @@ class AbstractProcessorStage(AbstractStage):
             for line in iterator:
                 yield self.parseMessage(line)
         else:
-            raise NotImplementedError("Stream input is not implemented for"
-                                      " custom EOMessage marker.")
+            for line in custom_readline(fd, self.ARGS.eom):
+                yield self.parseMessage(line)
 
     def file_input(self, fd):
         """ Generator for input messages.
--- a/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
+++ b/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
@@ -78,9 +78,6 @@ class AbstractProcessorStage(AbstractStage):
     __input_message_class = None
     __output_message_class = None
 
-    __stream_EOMessage = '\n'
-    __stream_EOProcess = '\0'
-
     def __init__(self, description="DKB Dataflow data processing stage."):
         """ Initialize the stage
 
@@ -94,8 +91,6 @@ class AbstractProcessorStage(AbstractStage):
         self.__input = []
         self.__output_buffer = []
         self.__stoppable = []
-        self.__EOMessage = '\n'
-        self.__EOProcess = '\n'
         super(AbstractProcessorStage, self).__init__(description)
 
     def _set_input_message_class(self, Type=None):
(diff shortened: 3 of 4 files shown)


Utils/Dataflow/pyDKB/common/custom_readline.py
```diff
@@ -0,0 +1,13 @@
+def custom_readline(f, newline):
```
mgolosova: Please, add docstrings here.
Although the function is not over-complicated, it would be easier to read it when you know in advance what is `f` and `newline`, and what is expected as a function result :)

Utils/Dataflow/pyDKB/common/custom_readline.py
```diff
@@ -0,0 +1,13 @@
...
+    max_buf_size = 1
+    while True:
+        while newline in buf:
+            pos = buf.index(newline)
+            yield buf[:pos]
+            buf = buf[pos + len(newline):]
+        chunk = f.read(max_buf_size)
+        if not chunk:
+            yield buf
+            break
```
mgolosova: Here it does return the last data read from `f` even if they are not terminated with EOM.
On the one hand, if it is a full message, we will be able to parse it (JSON or whatever); if not -- an error will occur and the message will be skipped.
On the other hand, if the data format does not allow to tell full message from its part (for text messages, for example), we will end up with wrong data.

Currently we use JSON, so it can be left as is, yet at least in the docstrings it should be noted -- that the last `line` can be incomplete, if the input data flow was interrupted in the middle of data writing.

Utils/Dataflow/pyDKB/dataflow/stage/AbstractStage.py
```diff
@@ -29,7 +29,22 @@ def __init__(self, description="DKB Dataflow stage"):
...
+                mode | source | dest | eom | eop
+                -----+--------+------+-----+-----
+                  s  |    s   |   s  |  \\n |  \\0
+                -----+--------+------+-----+-----
+                  f  |   f/h  |  f/h |  \\n |
+                -----+--------+------+-----+-----
+                  m  |   s/h  |   s  |  \\n |
+                -----+--------+------+-----+-----
+                  h  |    h   |  h/f |  \\n |
+                -----+--------+------+-----+-----''')
```
mgolosova: As you said, it looks a bit misplaced in the full help message:
```
>>> parser.print_help()
usage: [-h] [-m [MODE]] [-e [EOM]] [-E [EOP]]

Stage description

optional arguments:
  -h, --help            show this help message and exit
  -m [MODE], --mode [MODE]
                        Processing mode: (f)ile, (s)tream or (m)ap-reduce
                        (default: f).
  -e [EOM], --end-of-message [EOM]
                        Custom end of message marker.
  -E [EOP], --end-of-process [EOP]
                        Custom end of process marker.

                Shortcut table:
                
                mode | source | dest | eom | eop
                -----+--------+------+-----+-----
                  s  |    s   |   s  |  \n |  \0
                -----+--------+------+-----+-----
                  f  |   f/h  |  f/h |  \n |
                -----+--------+------+-----+-----
                  m  |   s/h  |   s  |  \n |
                -----+--------+------+-----+-----
                  h  |    h   |  h/f |  \n |
                -----+--------+------+-----+-----
```

I think it would really be better to move it to the left edge and add a couple of words to explain what are we talking about, to get something like this:
```
>>> parser.print_help()
usage: [-h] [-m [MODE]] [-e [EOM]] [-E [EOP]]

Stage description

optional arguments:
  -h, --help            show this help message and exit
  -m [MODE], --mode [MODE]
                        Processing mode: (f)ile, (s)tream or (m)ap-reduce
                        (default: f).
  -e [EOM], --end-of-message [EOM]
                        Custom end of message marker.
  -E [EOP], --end-of-process [EOP]
                        Custom end of process marker.

NOTES

Processing mode
    is defined as a combination of data source and destination type (local/HDFS
    file(s) or standard stream) and pre-defined EOP and EOM markers:

    mode | source | dest | eom | eop
    -----+--------+------+-----+-----
      s  |    s   |   s  |  \n |  \0
    -----+--------+------+-----+-----
      f  |   f/h  |  f/h |  \n |
    -----+--------+------+-----+-----
      m  |   s/h  |   s  |  \n |
    -----+--------+------+-----+-----
      h  |    h   |  h/f |  \n |
    -----+--------+------+-----+-----
```

Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
```diff
@@ -309,8 +310,8 @@ def stream_input(self, fd):
             for line in iterator:
                 yield self.parseMessage(line)
         else:
-            raise NotImplementedError("Stream input is not implemented for"
-                                      " custom EOMessage marker.")
+            for line in custom_readline(fd, self.ARGS.eom):
+                yield self.parseMessage(line)
```
mgolosova: Not sure, but maybe it would be better to make `iterator` variable defined depending on `eom` value, and leave common loop `for line in iterator: yield` outside the `if/else` clause?

anastasiakaida: Got it. I think you are right. Now it looks a bit odd.

anastasiakaida:
Actually, I'm still not sure about custom_readline() behavior on real data. Tests using sample script were executed correctly.

mgolosova:
I think, commit log for 3faaec0 (`pyDKB/fix: epilog improvements`) might be extended to explain what exactly was changed.

However, I also suspect that this commit will be joined with 6e30649 as soon as we start rewriting history, so this is just a comment for the future -- nothing to be changed here ;)

Utils/Dataflow/pyDKB/common/custom_readline.py
mgolosova: Thank you for 34c8505.

One more comment: PEP8 [suggests](https://www.python.org/dev/peps/pep-0008/#documentation-strings) to follow the [docstring conventions](https://www.python.org/dev/peps/pep-0257/).
In short, the last states:
* one-string docstrings (for obvious cases) are to look like this:
```
def function(a, b):
    """Do X and return a list."""
```
* multiline docstrings are to look like this:
```
def complex(real=0.0, imag=0.0):
    """Form a complex number.

    Keyword arguments:
    real -- the real part (default 0.0)
    imag -- the imaginary part (default 0.0)
    """
```
, where the first line is a "summary", then goes a blank line, and then -- everything you would like to say about the function in addition to the summary (just like in commit logs, right?)

We (or, at least, I) do not use tools for auto-generating of project documentation (yet ;) ), but even for a human eye it looks more structured if follows the recomendations.

The PEP document also provides some recommendations about what to write in the function description, so it might be useful to read the full version, not only my interpretation ;)

Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
```diff
@@ -309,8 +309,12 @@ def stream_input(self, fd):
             iterator = iter(fd.readline, "")
         else:
             iterator = custom_readline(fd, self.ARGS.eom)
-        for line in iterator:
+        try:
+            for i, line in enumerate(iterator):
                 yield self.parseMessage(line)
+        except KeyboardInterrupt:
+            sys.stderr.write("(WARN) Input stream processing interrupt"
+                             " on line %s\n" % i)
```
mgolosova: `KeyboardInterrupt` (raised after pressing `Ctrl+C` in terminal, for example) is meant to stop the program (in a more or less gentle way), right?

Now let`s see what happens if the exception was caught here.

1. The `stream_input()` comes to its end, so `StopIteration` is raised.
2. This `StopIteration` stops loop in `run()` method, which then returns nothing (`None`, to be precise):
https://github.com/PanDAWMS/dkb/blob/48599f0b26af1a1de121c43d9ada5cc4a616497e/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py#L229-L240
3. Generally it means that the main program will also come to its end, like it does here:
https://github.com/PanDAWMS/dkb/blob/48599f0b26af1a1de121c43d9ada5cc4a616497e/Utils/Dataflow/054_docContent2TTL/docContent2TTL.py#L161-L171
  ...but nobody said that the one who uses pyDKB library can\`t do something else: e.g., restart processing in a `while True` loop. If one does, then pressed `Ctrl+C` will interrupt processing just to restart it again, and again -- till user gives up and kills the program in a less gently way. I met such a behaviour and it was a bit annoying ;)

Of course, the villain is the one who added `while True` loop aroun `run()` method call ;) Still I think we must let him know that the program was somehow interrupted, **not losing or masking** this event somewhere inside the library.

---

Actually, this is a question I wanted to discuss concerning [this](904674a284df43e629b055d20987630d144b25a6) commit of mine (see PR #99 ), which does much the same as 48599f0 here -- but in `run()` method. "Closer to the edge", so to speak :)

If we want to output warning/information message here (which is fair enough), the `try/except` block is on its proper place, in both cases (in PR #99 it ends up in `output_error()` method of `AbstractStage`, that outputs traceback OR says "Interrupted by user" in case of `KeyboardInterrupt`).
But what do we do after this, if we want to ensure the main program that uses `pyDKB` will also be interrupted?
We can re-raise `KeyboardInterrupt` (or any other `BaseException` (not `Exception`) caught), after it was handled some way in the library, or raise `SystemExit` (same to call `sys.exit()`), or...?

In the #99 I chose to use `sys.exit()`, so that the program ends, whatever is in the main code after `run()` method call (if no one catches `SystemExit`, of course):
https://github.com/PanDAWMS/dkb/blob/904674a284df43e629b055d20987630d144b25a6/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py#L246-L260

The idea was to screen final user from any traceback output in those cases when it is obvious that the program must die (namely, the `KeyboardInterrupt` case) -- but provide more or less uniform traceback output in other cases. And by doing this here, in the library -- save the main program developer that uses `pyDKB` from solving this problem again and again.

The only thing is that I am not sure it was done good enough, that nothing was forgotten of missed, and that the logic is not inverted somehow :)

---

Concerning this PR (sorry for the long lyrical digression):
1. `KeyboardInterrupt` is masked now: code outside `stream_input()` doesn`t know, if the input data are over or the program was interrupted by user. Both variants look the same now. Was it intented?
1. I see three possible ways to improve the situation:
    * re-raise exception
    *(then after #99 there will be two messages about "Interrupted by user.", but I will fix it after merging actual version to #99)*
    * call `sys.exit()` right here 
    *(then for now the program will simply exit here -- looks good enough for me -- and it will become a good reason to properly handle `SystemExit` in #99)*
    * leave `KeyboardInterrupt` unhandled and wait for #99
    *(which is just not fair, since you`ve noticed that it is unhandled, thought of how to fix it AND read to the end of this comment ;) )*

anastasiakaida: @mgolosova 
Thank you for all comments. I'll fix everything tomorrow.
____________
Asking to your questions:
1. No. It wasn't intented.
2. Also, I prefer to call `sys.exit()` as it was in your second offer. :) Should I call it in added `KeyboardInterrupt`?

Utils/Dataflow/pyDKB/common/custom_readline.py
anastasiakaida: Ok, I'll fix it according to your comment! :)

anastasiakaida:
Can I add the explanation of 3faaec0 and 6e30649 in a new branch (for rewriting history) or I should do it exactly here in advance?

Utils/Dataflow/pyDKB/common/custom_readline.py
mgolosova: Thank you!

mgolosova:
@anastasiakaida,
no. As I said -- the 3faaec0 will most probably be joined with 6e30649 (which log, by the way, contains information about technical desicions, but nothing about the purpose of adding the epilog itself) later.
So log message of the first one (3faaec0) will disappear, while the second can be updated / rewritten / whatever in the process of branch rebasing.

Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
mgolosova: Sorry, missed your reply.

@anastasiakaida wrote:
> Should I call it in added KeyboardInterrupt?

You mean, in the `except` part, where the `KeyboardInterrupt` was caught? Yes, output something to explain why the program has stopped and call `exit()`.
It seems to me that it will be Ok to exit with the default return value, as it is a regular situation -- so no need to return some error code.

anastasiakaida:
I fixed all mistakes that were mentioned, now it seems to be ready for checking. If everything is ok, I'll rewrite commit history for merging.

anastasiakaida:
@mgolosova 
Done. Thank you!

Utils/Dataflow/pyDKB/common/__init__.py
```diff
@@ -5,4 +5,5 @@
 from exceptions import *
 import hdfs
 import json_utils as json
+import custom_readline
```
mgolosova: (just a suggestion)
Maybe `from custom_readline import custom_readline`, so that in `AbstractProcessorStage.py` (and in any other place) it could be addressed simpty as `from pyDKB.common import custom_readline`?
I think it is more readable way to organize `common` submodule, especially when a sub-submodule contains single homonymous function.

anastasiakaida: I suppose it's a rather good idea. :)

Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
```diff
@@ -311,13 +305,16 @@ def stream_input(self, fd):
...
+        if self.ARGS.eom == '\n':
             iterator = iter(fd.readline, "")
-            for line in iterator:
-                yield self.parseMessage(line)
         else:
-            raise NotImplementedError("Stream input is not implemented for"
-                                      " custom EOMessage marker.")
+            iterator = custom_readline(fd, self.ARGS.eom)
+        try:
+            for i, line in enumerate(iterator):
```
mgolosova: Do we really need `i` here? Won\`t simple `for line in iterator` also work?

anastasiakaida:
@mgolosova 
Well, I'm going to rewrite history as follows:
c5a7460 -- pyDKB/enhance: add configurable EOM/EOP markers
d076a61 -- pyDKB/enhance: empty EOM marker exception
b465240 -- pyDKB/enhance: custom readline implementation + 1befa5b, 1c815a6, 38418ef, 5dbd54b
6e30649 -- pyDKB/enhance: ArgumentParser epilog + 0952cc9
55c6ef1 -- pyDKB/enhance: KeyboardInterrupt handling + cffadc1
772e525 -- pyDKB/enhance: stream_input() improvements

+ all codestyle fixes during adding other commits.

I hope, I didn't miss anything.

Utils/Dataflow/pyDKB/dataflow/stage/AbstractStage.py
```diff
@@ -54,6 +89,31 @@ def parse_args(self, args):
...
+        if self.ARGS.eom is None:
+            self.ARGS.eom = '\n'
+        elif self.ARGS.eom == '':
+            raise ValueError("(ERROR) Empty EOM marker specified.\n")
+        else:
+            try:
+                self.ARGS.eom = self.ARGS.eom.decode('string_escape')
+            except (ValueError), err:
+                sys.stderr.write("(ERROR) Failed to read arguments.\n"
+                                 "Case: %s\n" % (err))
```
mgolosova: For multiline log messages every line should be prepended with `(ERROR)` (or only the first line, and all the others with `(==)` [1].

It is required due to the implementation of the [ExternalProcessLogger](https://github.com/PanDAWMS/dkb/blob/master/Utils/Dataflow/000_kafka/ru/kiae/dkb/kafka/common/external/ExternalProcessLogger.java). Basically, it does the following: it looks for `(ERROR|WARN|WARNING|INFO|DEBUG|INFO)` marker at the beginning of the line read from the subprocess\` STDERR, and writes the line to the corresponding log level of the Java program. So when the Java program runs this or that DKB stage script, it has log messages from the stage in its own log file, and the one reading this log can check what happened during the data processing on every stage.

However when there\`s no log level marker, ExternalProcessLogger takes `(TRACE)` level as default -- and if logging level is set to something else (say, `(INFO)`), this line, even being a second line of the `(ERROR)` message, won`t appear in the log file and important information will be missed.

---
[1] The part of `(==)` is not implemented yet on the Java part, yet one day it will be and will work as "same as above".

mgolosova:
@anastasiakaida, 
Alright, I`m done with my review :) Thank you for your patience ;)

Go ahead!
And do not forget to store the original variant in a separate branch, just in case someone who does not have a local copy would like to join the conversation ;)

Just a couple of questions/comments.
You seem to leave `KeyboardInterrupt` related commits (55c6ef1, cffadc1, 13f21fd) as they are, right?
It is not good nor bad, just to ensure it was intended ;)

And look carefully on your commit logs -- I mean with the rebasing, some will require updates, though sometimes it is really tempting to leave them "as is" ;)
Remember, that first of all the commit log should describe ***what*** happens here  (say, what exactly was improved in 772e525?) and ***why*** it is done (say, why you decided to handle `KeyboardInterrupt` and what happens if you leave it "as is"?). And then, if it looks questionable -- ***why*** it is done ***exactly this way*** (as you did, adding epilog to parser in 6e30649).

Good luck!

anastasiakaida:
@mgolosova 
Just some words about KeyboardInterrupt(). I'm going to combine some commits when it's possible, because, in my opinion, sometimes there is no reason to left it 'as is' (e.g. fixes). And I'm going to do it with other logical parts of this branch the same way.

Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
anastasiakaida: @mgolosova 
It was created for printing the line number in a previous variant of KeyboardInterrupt exception handler. I forgot to turn this line back into the previous form. I'll fix it. Thank you!

anastasiakaida:
@mgolosova 
Rewritten commit history is almost ready! I'm going to upload it to a new branch and close this branch (with the reason to left everything here as it was).

PR #99, opened 2017-12-19
[WIP] pyDKB: improvements

mgolosova:
Main changes:
* unified runtime exceptions handling
* unified argument parse error handling
* ...

ToDo:
- [ ] rethink the logic of output file naming and store directory. It is necessary as when we run json2JSON processor, it is easy to get the collision of input and output files.
- [x] more accurate exception handling (so that app using pyDKB had no need in `try...except` around stage methods)
- [ ] add `--config FILE` as default parameter, with config parsing
- [ ] ...


maria-grigorieva:
Developing version of pyDKB was tested on stages 016 and 055 on branch 'pyDKB-dev-test'. 
Everything works fine in 'file' and 'stream' modes. 
So, I'm sure that it will function as expected in other cases and can be approved.

mgolosova:
@anastasiakaida,
as we are going to make some more improvements to pyDKB, could you please take a look at this PR?
I think it would be better to merge it now, then to add more and more unrelated changes.

anastasiakaida:
@mgolosova,
Ok, I'll do it until the day's end.

anastasiakaida:
@mgolosova, 
The code looks fine. I think it would be better if I'll test it with stage 054. I'll do it... I think, it would be better to say 'today'.
 :)

mgolosova:
@anastasiakaida,
thank you! :)

anastasiakaida:
@mgolosova,
I've (suddenly :) ) found a typo in a word 'Delimited' [here](https://github.com/PanDAWMS/dkb/blob/aec29c2ca96bc43afd723e3219ff267d95334eed/Utils/Dataflow/pyDKB/dataflow/stage/processors.py#L33)

I have some questions about stage 054, but it is not connected with this improvements, I suppose.

Everything else looks good.

mgolosova:
@anastasiakaida,
Right, fixed the typo. Thank you!

PR #100, opened 2017-12-22
Sample ElasticSearch requests

maria-grigorieva:
Added the following ES request examples:
* dataset_aggregations.json - average dataset size, categorized by task parameters
* dataset_by_taskid.json - search datasets by their's IDs
* event_summary.json - Event Summary Report, categorized by category and production step
* keyword_search.json - search tasks by a set of keywords
* regexp_search.json - search datasets by regexp
* tasks_filtered.json - example of task search with filtration by a set of terms


mgolosova:
Will do my best to review till today`s end.

mgolosova:
Questions, pt 1.
---
Am I right and the first example ([aggregated summary](https://github.com/PanDAWMS/dkb/pull/100/files#diff-d99e8bfca6942271fe4b81a3a2b04db9R1)) aggregates tasks (and their datasets) by Task field "output_formats", and if there were more than one value for that field (say, "AAA" and "BBB"), we would count same output datasets more than once? I mean, would they give input into average value for type "AAA" as well as for "BBB"?

And what I noticed: we need to fix "output_formats" detection, as in Oracle this field has a dot-separated value, while we, I believe, would like to see an array (added to ToDo for #88, Stage 016).

The [last query](https://github.com/PanDAWMS/dkb/pull/100/files#diff-d99e8bfca6942271fe4b81a3a2b04db9R96) in "dataset_aggregations" seems to return datasets records themselves, as well as the average dataset size. Is it how it supposed to be? If yes, I think we need to change the query description, as it says only "Average dataset size".

Do we need "match_all" part [here](https://github.com/PanDAWMS/dkb/pull/100/files#diff-f9605ab02d077574a678fd6140ee3ddd)? Seems like it returns same number of documents with or without it. I believe it does not affect the performance at all, but it can be a bit confusing for those (like me) who is not that experienced in ES query design ;)

Stopped at "event_summary" query. Will return to it later today, or tomorrow.

mgolosova:
Pt.2
---

1. Found couple of typos:
* [here](https://github.com/PanDAWMS/dkb/pull/100/files#diff-3fcd546a2fef35e74e6331a3d0bf3fb2R2): `...search IN index...`
* and [here](https://github.com/PanDAWMS/dkb/pull/100/files#diff-e8d8ee67836f7ec8ce6f48f7c5fa35edR24): comma is missed in the end of line.

2. Formatting: noticed, that in a couple of places indentation is a bit shifted. It does not break the query itself, but as there a lot of nested and over-nested parts, it breaks readability :(
I think that there must be some JSON normalizers, that might help to fix it painlessly :)

3. I have a suggestion concerning query templates (where placeholders like `<CAMPAIGN>` are used): perhaps, we should list all the parameters in the heading comment? I tried some queries few times before realized that not all the placeholders were found and replaced with real values :)

4. And the last question is more about mapping: namely, the `keyword` subfields usage.

It was raised by the difference in [this](https://github.com/PanDAWMS/dkb/pull/100/files#diff-f9605ab02d077574a678fd6140ee3dddR1) and the [last](https://github.com/PanDAWMS/dkb/pull/100/files#diff-e8d8ee67836f7ec8ce6f48f7c5fa35ed) examples. First uses `{"term": {"<TERM_NAME>": "<TERM_VALUE>" }}`, while the last one -- `{"term": {"<TERM_NAME>.keyword": "<TERM_VALUE>"}}`. Was it done on purpose, or in the first template one should also use "keyword"? Or in the last one -- pure `<TERM_NAME>`?

It took me a while, but finally I think I got the background.
As I understand, the difference is that when we specify `*.keyword`, we mean "this field has a sub-field, indexed as `keyword`, and for exact match we want to lookup by it"; while simpty using `"<TERM_NAME>"` we use the field\`s native indexing (mostly the `text` one). I am not sure if it affects performance somehow, when we use "term query" -- at least for single-word fields, like "campaign". "Term query" returns documents with exact matched terms in the given field, and for single-word field we have exactly one term -- the same word, only normalized.
However, when I tried to use `status.keyword` in the search, I got an empty response (while for pure `status` it worked as I expected), and only by checking the mapping I realized that this one is natively indexed as `keyword` and thus has no subfield for `keyword` indexing... so, I must say, it is a bit confusing matter :)

**My question is: do we really need subfield `keyword` for the most of the fields?**
Say, if we look for the tasks from MC16 campaign:
* `"term": {"campaign": "MC16"}` -- we will get nothing, as the "campaign" field is `text`-indexed so it contains only "normalized" terms: ["mc16"];
* `"term": {"campaign": "mc16"}` -- it will give us both documents with "campaign": "MC16" and with "campaign": "mc16", as being `text`-indexed, they contain same set of terms;
* `"term": {"campaign.keyword": "mc16"}` -- only documents with "campaign": "mc16"
* `"term": {"campaign.keyword": "MC16"}` -- only documents with "campaign": "MC16"

Do we *really* need any of this variants but the second one?

I understand it is more about mapping, not the queries and this PR itself, but only with queries we can understand how the mapping works, exactly ;) So it would fit the PR theme if we decide to change mapping here... after all, the mapping is to allow us to make any request we want in the most straightforward way.

With this new understanding, I think I need to re-read the queries again, to make sure that this `keyword-` and `term-`related thing does not lead to an unexpected results somewhere...

Utils/Elasticsearch/requests/datasets_by_task_params.json
```diff
@@ -0,0 +1,42 @@
+/*
+  Search  datasets by task parameters
+*/
+GET prodsys_rucio_ami/output_dataset/_search
+{
+  "query": {
+    "has_parent": {
```
maria-grigorieva: "match_all" has been removed from request, as it's really not needed here.

Utils/Elasticsearch/requests/dataset_aggregations.json
```diff
@@ -0,0 +1,155 @@
...
+   "aggs": {
+     "campaign": {
+       "terms": {
+         "field": "campaign.keyword"
+       },
+       "aggs": {
+         "format": {
+            "terms": {
+              "field": "output_formats"
+            },
```
maria-grigorieva: Currently this request returns average dataset size for all cases of 'output_formats', which is processing like string. When we implement new mapping, with 'pattern' analyzer - this request will be changed to group dataset formats, according to 'datasetname' value.

Utils/Elasticsearch/requests/dataset_aggregations.json
```diff
@@ -0,0 +1,155 @@
...
+
+/*
+  Real example:
+  Average dataset size
+  Search by datasets
+*/
+GET prodsys_rucio_ami/output_dataset/_search
+{
+  "size": 0,
+  "query": {
```
maria-grigorieva: To prevent returning a search hits, we should user parameter "size": 0. So, only aggregation results will be returned.

maria-grigorieva:
Obviously, for campaign we don't need a keyword, because when we want 'MC16' - it must matches 'mc16' and 'MC16'. 
I added *.keyword as in most cases we need this parameter, but not necessary. So, it's just an common example. 
I agree that we have to think carefully about possible queries and suitable mapping.

maria-grigorieva:
I think that in Web interface we should convert all input parameters to lower-case. For example, string like this 'MC16a, AOD, AODMerge_tf.py' will be converted to 'mc16a, aod, aodmerge_ft.py'.   
But we must return query results as they are stored in ES, case-sensitive. For example, the end-user doesn't want to get the result like 'atlasoffline-21.0.8', but he want to get 'AtlasOffline-21.0.8' as he used to see it. So, we have to maintain case-sensitivity in ES, but for requests - parameters should be lower-cased to provide more comprehensive query results.

Utils/Elasticsearch/requests/dataset_aggregations.json
mgolosova: So, we should not split this dot-splitted list into real list on Stage 016?

mgolosova:
@maria-grigorieva,
seems like in case of "query string" ES query the query string is being splitted by spaces (leaving phrases in quotes unsplitted) and then every word (or phrase) is being analyzed, including lower-casing the terms. So there`s no difference if we do convert the search string to lowercase or not, and we can forget about it in the Web part:
```
[dkb@aiatlas171 ~]$ cat << "EOF" | curl -XGET 'http://127.0.0.1:9201/prodsys_rucio_ami/task/_search?pretty' --data-binary @-
{
  "size": 1,
  "query": {
    "query_string": {
      "query": "description:\"a3 TUNE mc\""
    }
  }
}
EOF                                                                                                       
{
  "took" : 2,
  "timed_out" : false,
  ...
  "hits" : {
    ...
    "hits" : [
      {
        "_index" : "prodsys_rucio_ami",
        "_type" : "task",
        "_id" : "9786284",
        "_score" : 18.208021,
        "_source" : {
          ...
          "description" : "ESD level A3 Tune MC Samples for Tile Noise",
          ...
        }
      }
    ]
  }
}
```
same as:
```
[dkb@aiatlas171 ~]$ curl -XGET 'http://127.0.0.1:9201/prodsys_rucio_ami/task/_search?q=description:"A3+Tune+mc"&pretty&size=1'  
<...>
```

And yes, it returns data just as they are stored, so it`s gonna work right as we need it more or less automatically.

Utils/Elasticsearch/requests/dataset_aggregations.json
maria-grigorieva: I have to think about this field. I'll try to answer tomorrow ))

maria-grigorieva: I suppose we should split output_formats by dot in Stage 016! 
But we wouldn't make aggregations on this field, as this is just enumeration on output data formats for the task. 
If we need aggregation by data format we will use 'data_format' parameter from 'output_dataset' type. 
'output_formats' will be used only, for example, to search all tasks, returning 'DAOD_*' data format, or all tasks, returning both 'AOD' and 'HITS' data formats.

mgolosova:
@maria-grigorieva,
Please, would you mind updating PR with additional commits instead of rewriting the original one?
After all the review questions are addressed, they could be squashed together with `rebase -i` + `push --force` to the PR branch, so that in the end we had one beautiful commit -- but for now it is quite annoying to track which changes are already made and which are still required.

Your comments were helpful, but unfortunately not ultimate :(

---

Back to the review.
As far as I understand, now addressed (or half-addressed):
> Am I right and the first example (aggregated summary) aggregates tasks (and their datasets) by Task field "output_formats", and if there were more than one value for that field (say, "AAA" and "BBB"), we would count same output datasets more than once? I mean, would they give input into average value for type "AAA" as well as for "BBB"?
>
> And what I noticed: we need to fix "output_formats" detection, as in Oracle this field has a dot-separated value, while we, I believe, would like to see an array (added to ToDo for #88, Stage 016).

The result of the discussion is the following:
* the `output_formats` will be splitted by dot in Stage 016 (#88)
* **question: should the query be rewritten after this?**

> The [last query](https://github.com/PanDAWMS/dkb/pull/100/files#diff-d99e8bfca6942271fe4b81a3a2b04db9R96) in "dataset_aggregations" seems to return datasets records themselves, as well as the average dataset size.

Fixed.

> Do we need "match_all" part [here](https://github.com/PanDAWMS/dkb/pull/100/files#diff-f9605ab02d077574a678fd6140ee3ddd)?

Fixed.

> Found couple of typos:
> * [here](https://github.com/PanDAWMS/dkb/pull/100/files#diff-3fcd546a2fef35e74e6331a3d0bf3fb2R2): `...search IN index...`

Fixed.

> * and [here](https://github.com/PanDAWMS/dkb/pull/100/files#diff-e8d8ee67836f7ec8ce6f48f7c5fa35edR24): comma is missed in the end of line.

Was not easy to realize, where this link is pointing to, but, after all, I confirm -- it is fixed.

> **My question is: do we really need subfield `keyword` for the most of the fields?**

The last one is not directly related to the content of the current PR, yet it is quite close to it.
It was half-discussed. The results of the discussion:
* think carefully about using (or not using) the `keyword` field when construct query.

Questions left:
* shouldn\`t we change the mapping and remove excessive `keyword` subfield for those fields that do not need it, like `campaign`?
**OR**
* should we keep `keyword` subfield for all the fields just for some kind of uniformity?

---

NOT addressed (nor commented) yet:

> 2. Formatting: noticed, that in a couple of places indentation is a bit shifted.

> 3. I have a suggestion concerning query templates (where placeholders like `<CAMPAIGN>` are used): perhaps, we should list all the parameters in the heading comment?

maria-grigorieva:
Thank you for comments and reminder about 'data_format'. 
I've changed aggregation queries, according to this field and checked how it works on the current test_prodsys_rucio_ami index at CERN. It works as expected. 

Concerning mapping and 'keyword'. I think we can leave mapping "as-is", just should keep in mind about 'keyword'.

Utils/Elasticsearch/requests/dataset_aggregations.json
```diff
@@ -0,0 +1,138 @@
...
+     }
+   }
+ }
+
+/*
+  Real example:
+  Average dataset size
+  Search by datasets
+*/
+GET prodsys_rucio_ami/output_dataset/_search
```
mgolosova: Should this query return average dataset size for all datasets?
I ask because there is a `should` clause:
```
"should": [
  {"term": {"data_format.keyword": "AOD"}},
  {"term": {"data_format.keyword": "DAOD"}}
]
```
If we wanted to get average size for AOD and DAOD datasets, then it does not work (if I specify `"size": 10"` I see datasets with `"data_format" : ["NTUP_PILEUP", "NTUP"]` as well). If we didn`t -- then this clause should be removed.

And, by the way, there\`s no `data_format.keyword` subfield in the current version of mapping (in `master`); `data_format` itself is of type `keyword`.

Utils/Elasticsearch/requests/event_summary.json
```diff
@@ -0,0 +1,49 @@
...
+  "size": 0,
+   "query": {
+     "bool": {
+       "must": [
+         { "term": { "campaign.keyword": "<CAMPAIGN>" } },
+         { "term": { "status": "<STATUS>" } }
+       ],
+       "should": [
+         { "term": { "hashtag_list": "<HASHTAG>"} }
+       ]
```
mgolosova: ES documentation [says](https://www.elastic.co/guide/en/elasticsearch/reference/5.4/query-dsl-bool-query.html):
>If the `bool` query is in a query context and has a `must` or `filter` clause then a document will match the bool query even if none of the `should` queries match. In this case these clauses are only used to influence the score.

In other words, `should` is used to put documents with given <HASHTAG> on top of the search results -- but in the summary it is as good as nothing. If we want to search by hashtag, it should also be presented in `must` clause, if not -- removed.

mgolosova:
@maria-grigorieva,
Thank you! I believe the queries must work, as you have checked them (yet will try them myself a bit later, just to make sure I understand them properly ;) ).
What about formatting and description of placeholders?

About `keyword` subfield: the reason why we can not remove it for `campaign` etc. is the fact that it is required for aggregations (as long as we haven\`t enabled [`fielddata`](https://www.elastic.co/guide/en/elasticsearch/reference/5.4/fielddata.html) and ain\`t going to).
Thus, I think, we might want this field to be lower case and use in aggregations only... if it is possible (to be investigated).

Also, this part caught my attention:
https://github.com/PanDAWMS/dkb/blob/071f764635aae2b026fe81f2ef88fd704a0baa9d/Utils/Elasticsearch/requests/event_summary.json#L16-L26

As it was already mentioned:
> Say, if we look for the tasks from MC16 campaign:
> * `"term": {"campaign": "MC16"}` -- we will get nothing, as the "campaign" field is `text`-indexed so it contains only "normalized" terms: ["mc16"];
> * `"term": {"campaign": "mc16"}` -- it will give us both documents with "campaign": "MC16" and with "campaign": "mc16", as being `text`-indexed, they contain same set of terms;
> * `"term": {"campaign.keyword": "mc16"}` -- only documents with "campaign": "mc16"
> * `"term": {"campaign.keyword": "MC16"}` -- only documents with "campaign": "MC16"

I think, it should be one of the following (unless we really want case-sensitive search):
* `"term": {"campaign": "<CAMPAIGN>"}`
  (with a note in the preceeding comment that <CAMPAIGN> must be lowercase), 
* `"query_string": {"query": "campaign:<CAMPAIGN>"}`.
  And, by the way, if we use `query_string`, we can replace the above part of the request with:
  ```
  "query": {
    "query_string": "campaign:<CAMPAIGN> AND status:<STATUS> OR hashtag_list:<HASHTAG>"
  }
  ```
* `"match": {"campaign": "<CAMPAIGN>"}`


It is not that uniformity in the queries is something bad (I mean the tendency to use `term` query), but here we can chase two goals:
* give ready-to-use queries to retrieve this or that data, and thus:
  * if we leave space for user values, we must clearly specify how to fill it [1] and make sure that
  * the queries will return expected results (independent of how the data were written to ES -- in lower, upper or mixed case) [2];
* give examples of how to construct queries, and thus:
  * the queries must be readable (mind the comments about broken JSON formatting),
  * easy-to-understand and avoid unnecessary complications [3].

---
[1] e.g. when `term` query is used with placeholders for `text` fields, we should specify in the query description that user values must be in lower case
[2] e.g. when `term` query is used with placeholders for `keyword` fields (including `*.keyword` subfields), we should specify that the query is to be used for case-sensitive search only
[3] <here must be an example, but it will be added later>

Utils/Elasticsearch/requests/dataset_aggregations.json
maria-grigorieva: Thank you! It was fixed in commit #ac069fc

mgolosova: Thank you!

However, I have a couple of comments about this commit.

1. ac069fc description says:
```
ES/agg datasets query fixed

- 'Should' (OR) clause was moved inside the 'must' (AND) clause
- .keyword suffix removed from 'data_format'
```

How does this change correspond to it?
```
-            "query":
-              {"term": {"project.keyword": "mc15_13TeV"}}
+            "query":  {"term": {"project.keyword": "mc15_13TeV"}}
```

2. You change both formatting and query structure (and in the resulting query the indentation is a bit wierd), so it is very hard to get the essence of the change.

Look at 743fd7d -- it is perfectly clear, what and where was moved, as the query was uniformly formatted originally and follows the same style in the new version.

I would strongly recommend you to unify the formatting of all queries the way you take most readable (say, it is fair enough to place `"query":  {"term": {"key": "value"}}` on one line instead of 5 lines, that would be produced if we did it strictly); and then follow this formatting whatever changes you do.
Maybe even do it in the first commit and update next ones to follow the chosen variant.

I know it will take time and will be painfull. But isn`t it painful to me -- stumbling over this issue again and again, despite asking you to fix it? :(

3. I would also suggest more detailed description, that would explain why and where it was moved, something like:
```
We need the query to behave like "A and B and (C or D)"

Was: 
  "bool": {
    "must": [ A, B ],
    "should": [ C, D ],
  }

Here "should" clause affects only the order of the query results (<link to the ES docs>).
To make its conditions required, it must be moved inside the "must" clause:
  "bool": {
    "must": [
      A, B,
      "bool": { "should": [ C, D ] }
    ]
  }
```

It takes a bit more time to make this description, but it woulf significantly simplify the grasp of the commit.
I bet you yourself wouldn`t recall this logic a month from now, at least not without looking around the commit. Why make others to speculate? ;)

maria-grigorieva: [1] Should I fix it using rebase? 
[2] formatting was fixed in the last commit
[3] message about MUST&SHOULD usage was added to commit message - 165d008

mgolosova: [1] Yes, sooner or later it seems to require a rebase. I check the queries as a whole all the time (as for now I am not good enough in ES query logic and can\`t juts "update" it in memory after some changes :) ), so if you feel like rewriting the history -- go ahead any time you want :)

[2] Thank you. Thought some issues are left... I will add comments to the code a bit later.

[3] It was not exactly about this commit. In 165d008 we did have 
```
  "bool": {
    "must": [ A, B ],
    "should": [ C ],
  }
```
originally, but new version is 
```
  "bool": {
    "must": [ A, B, C ]
  }
```
, right?

PR #102, opened 2017-12-25
Added 093_datasetsFormat stage

maria-grigorieva:
The main purpose of this stage is to add dataset format parameter to the JSON string of each dataset.

091_datasetsRucio/output --> 090_datasetsFormat/output --> 092_datasetInfoAMI

Result:
   * added 'data_format' parameter, extracted  according to datasets nomenclature
   https://dune.bnl.gov/w/images/9/9e/Gen-int-2007-001_%28NOMENCLATURE%29.pdf()

- [x] Data format is splitted into 2 parts: general data format (AOD, DAOD, HITS, ...) and full data format (DAOD_SUSY5, ...)

- [x] Add 'data_format' field in ES mapping

- [ ] Add to master branch after [PR#85](https://github.com/PanDAWMS/dkb/pull/85)

---
*Update:*
- [x] Update Dataflow/README
- [x] Add stage README
- [x] rename to Stage 92 (and 92 -- to 93)?
  *Renamed to 93; and 92 will become 95 -- just to leave space between stages.*
- [x] join README commit with stage initial commit (0548f79 with 2d44c20)
  *Note "data_format" field data format, mentioned in README: update it according to the current state of the stage:*
  * *2d44c20 -- 'string'*
  * *137cdb8 -- 'string' or 'tuple'*
  * *db4bf30 -- 'list'*


maria-grigorieva:
About the storage of 'data_format' field I have the following:
I've tested the method of keeping both - general data format (like 'DAOD') and full data format ('DAOD_SUSY5'). And here are my results.
Mapping file for index 'test_array_agg'
```
{
  "mappings": {
    "my_type": {
      "properties": {
        "bytes": {
          "type": "integer"
          },
        "data_format": {
          "type":  "keyword"
        }
      }
    }
  }
} 
```
Create index:
`curl -XPUT 'http://localhost:9200/test_array_agg' --data-binary "@test_array_agg.mapping"; echo`
Insert data:
```
curl  -XPOST 'http://localhost:9200/test_array_agg/my_type/' -d '{
    "bytes": 1,
    "data_format": "AOD"
}'
curl -XPOST 'http://localhost:9200/test_array_agg/my_type/' -d '{
    "bytes": 2,
    "data_format": "DAOD"
}'
curl -XPOST 'http://localhost:9200/test_array_agg/my_type/' -d '{
    "bytes": 3,
    "data_format": ["DAOD", "DAOD_SUSY5"]
}'
curl -XPOST 'http://localhost:9200/test_array_agg/my_type/' -d '{
    "bytes": 4,
    "data_format": ["DAOD", "DAOD_SUSY0"]
}'
curl -XPOST 'http://localhost:9200/test_array_agg/my_type/' -d '{
    "bytes": 5,
    "data_format": ["DAOD", "DAOD_SUSY5"]
}'
curl -XPOST 'http://localhost:9200/test_array_agg/my_type/' -d '{
    "bytes": 6,
    "data_format": "HITS"
}'
curl -XPOST 'http://localhost:9200/test_array_agg/my_type/' -d '{
    "bytes": 7,
    "data_format": ["HITS", "HITS_SUSY5"]
}'
curl -XPOST 'http://localhost:9200/test_array_agg/my_type/' -d '{
    "bytes": 8,
    "data_format": ["HITS", "HITS_SUSY0"]
}'
```
Aggregation query example, getting the average 'bytes' for all 'data_formats': 
```
GET test_array_agg/my_type/_search
{
    "size": 0,
    "aggs": {
      "formats": {
        "terms": { "field" : "data_format" },
        "aggs": {
         "bytes_agg": {
           "avg": { "field": "bytes" }
          } 
        }
      }
    }
}
```
And the results are:
```
"buckets": [
        {
          "key": "DAOD",
          "doc_count": 4,
          "bytes_agg": {
            "value": 3.5
          }
        },
        {
          "key": "HITS",
          "doc_count": 3,
          "bytes_agg": {
            "value": 7
          }
        },
        {
          "key": "DAOD_SUSY5",
          "doc_count": 2,
          "bytes_agg": {
            "value": 4
          }
        },
        {
          "key": "AOD",
          "doc_count": 1,
          "bytes_agg": {
            "value": 1
          }
        },
        {
          "key": "DAOD_SUSY0",
          "doc_count": 1,
          "bytes_agg": {
            "value": 4
          }
        },
        {
          "key": "HITS_SUSY0",
          "doc_count": 1,
          "bytes_agg": {
            "value": 8
          }
        },
        {
          "key": "HITS_SUSY5",
          "doc_count": 1,
          "bytes_agg": {
            "value": 7
          }
        }
      ]
```
And test query with filter by 'data_format':
```
GET test_array_agg/my_type/_search
{
    "size": 0,
    "aggs": {
      "formats": {
        "filter" : { "term": { "data_format": "DAOD_SUSY5" } },
        "aggs": {
         "bytes_agg": {
           "avg": { "field": "bytes"}
          } 
        }
      }
    }
}

"aggregations": {
    "formats": {
      "doc_count": 2,
      "bytes_agg": {
        "value": 4
      }
    }
  }
```
So, I can say that we've got the results correctly, as expected. It means that keeping general data format with full data format is the right way to proceed.

Utils/Dataflow/090_datasetsFormat/datasets_format.py
```diff
@@ -0,0 +1,73 @@
...
+    and first part ('DAOD'), defining the general name of the data format.
+
+    :param datasetname:
+    :return:
+    """
+    ds_format = datasetname.split('.')[4]
+    if re.match("\w+_\w+", ds_format) is not None:
+        return (ds_format, ds_format.split('_')[0])
+    else :
+        return ds_format
```
mgolosova: I think we need to return `list` object in both cases.
`tuple` is OK, as via JSON format it will be transformed to `list`, but generally we use `list` in such cases, so to avoid possible misunderstanding I would use `list`.

Returning `list` in both cases will reduce possibility of mistake on further stages of data processing: if we have number of possible data formats, we got to check it carefully (to avoid mistaking string as a list of symbols, for example, as we do from time to time).

mgolosova:
Original version of the PR is saved in `pr-102-stage-90-original` branch for reference.

What has changed:
* rebased on current version of PR #85.

Original commits are saved; new commits are added.

@maria-grigorieva,
please check that I haven`t broken anything important in your code :)
I checked with the same input data sample and it seems like there`s nothing changed except square brackets around single-item values of `data_format` (as can be seen in db4bf30), but who knows...

mgolosova:
Added README file and updated information in the Dataflow/README.
I think the last commit (0548f79) should be joined with the first one (2d44c20); not joined yet now to avoid mess in history.

maria-grigorieva:
I've checked the code - it's OK. Thank you!

mgolosova:
(Hopefully) last rebase:
* renamed to Stage 093
* rebased on Master
* joined commits:
  * add README (0548f79) -> create new stage (2d44c20)
  * make executable (84ea03b) -> create new stage (2d44c20)
  * replace input sample with soft link (ff24740) -> create new stage (2d44c20)
  * fix codestyle (a0265af) -> create new stage (2d44c20) & split data type by '_' (137cdb8)

To see diff with the previous version run:
`git diff db4bf304`

PR #103, opened 2017-12-25
Filtering datasets by format

maria-grigorieva:
Added filtration of the datasets by their data format 'on-the-fly'.
Filtration is implemented by the AOD, EVNT and HITS data formats.

Also, added input and output samples.

P.S. this PR in depends on [PR#102](https://github.com/PanDAWMS/dkb/pull/102) in which we've added 'data_format' parameter.


mgolosova:
Splitted first commit (8083461) in two: filtering by data format and README file.

PR #109, opened 2018-02-09
Stage 009: reverse data load

mgolosova:
Adding possibility to load data in a backward direction, moving deeper into the past with every step.

Main changes:
* reverse data load introduced (initiated by corredponding parameters in `timestamps` section:
  * `final` < `initial`,
  * `step` < 0.

Minor changes:
* code refactoring;
* improved config parameters handling (verification, default values, etc.).

### DIFF
--- a/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
+++ b/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
@@ -17,8 +17,6 @@ SQUASH_POLICY = 'SQUASH'
 
 # Global variables
 # ---
-# Config file
-conf = None
 # Operating mode
 mode = None
 # Output stream
@@ -38,27 +36,12 @@ def main():
        Query: SELECT min(timestamp) from t_production_task;
     """
     args = parsingArguments()
-    global conf
-    conf = args.config
     global mode
     mode = args.mode
 
     # read initial configuration
-    config = ConfigParser.SafeConfigParser()
-    try:
-        config.read(conf)
-        # unchangeable data
-        dsn = config.get("oracle", "dsn")
-        step = config.get("timestamps", "step")
-        step_seconds = interval_seconds(step)
-        final_date = str2date(config.get("timestamps", "final"))
-        queries_cfg = config.items("queries")
-        queries = {}
-        for (qname, f) in queries_cfg:
-            queries[qname] = {'file': config_path(f)}
-    except (IOError, ConfigParser.Error), e:
-        sys.stderr.write('Failed to read config file (%s): %s\n'
-                         % (conf, e))
+    config = read_config(args.config)
+    if config is None:
         sys.exit(1)
 
     # Offset storage initialization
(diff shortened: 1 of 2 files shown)


Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -91,25 +90,57 @@ def read_config(config_file):
...
+    :param default: object (any type)
+    :return: param value from config or default value
+    :rtype: object
+    """
+    if not isinstance(config, ConfigParser.ConfigParser):
+        raise TypeError("config_get() expects first parameter to be"
+                        " an instance of 'ConfigParser' (get '%s')."
+                        % config.__class__.__name__)
+    if type(section) != str and type(section) != unicode:
+        raise TypeError("config_get() expects second parameter to be"
```
anastasiakaida: What if in `if..else` will be `or` instead of `and`? Can't catch the main idea of this conditional operator. What will be if just a single condition will satisfy?

mgolosova: Let me think...
I believe it was an attempt to pass over the fact that Python 3 doesn`t have `unicode` type, but a very lazy one. Or, at least, not very accurate.

The idea is that when `section` is of type `str` (works for both Python 2 and 3), the `if` check won\`t reach the second condition and we won\`t see this message: `NameError: name 'unicode' is not defined`. When it is *not* `str`, in v.2 we simply check if it is `unicode` or not, while in v.3 we have `NameError` exception... not excactly charactonym one (as we\`d expect to see `TypeError` with description, which is within the `if` clause), yet the under same conditions.

If there was `or` instead of `and`, we would never pass the check:
```
>>> for section in ('string', u'string', 123):
...   if type(section) != str or type(section) != unicode:
...     print "Not passed!"
... 
Not passed!
Not passed!
Not passed!
```

Will add `try/except` handler around the construction (or, better, check for the Python version, for readability).

But later we need to add type-check function for input parameters in pyDKB, I think... something like `check_types((set, of, parameters), (set, of, types))`, where "type" can be any Python type / class, and for strings -- `str`, `unicode`, or some inner pyDKB constant named `STRING`.
It would be better then checking the parameters by hand every now and then...

Thank you!

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -38,60 +36,140 @@ def main():
...
+                        % section.__class__.__name__)
+
+    try:
+        result = config.get(section, param)
+    except ConfigParser.Error:
+        if default is not None:
+            sys.stderr.write("(INFO) Config: parameter '%s' (section '%s')"
+                             " is not configured. Using default value: %s\n"
+                             % (param, section, default))
+        result = default
```
anastasiakaida: In PR#98 you said that for multiline log messages every line should be prepended with (ERROR|WARN|WARNING|INFO|DEBUG|INFO) (or only the first line, and all the others with (==)). I see (INFO) only on the first line. Should it be fixed? :)

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -143,9 +205,13 @@ def init_offset_storage(config):
...
-            sys.stderr.write('(INFO) Stored offset is less then configured'
-                             ' initial date. Using initial date instead.\n')
+        elif not reverse and current_offset < initial_date \
+                or reverse and current_offset > initial_date:
+            sys.stderr.write('(INFO) Stored offset is %s then configured'
+                             ' initial date (%s data loading). Using initial'
+                             ' date instead.\n'
+                             % ('greater' if reverse else 'less',
+                                'reverse' if reverse else 'normal'))
             current_offset = initial_date
```
anastasiakaida: The same situation with `(INFO)`. 
And one more question. Why in the previous block of code it was double quotes and here - single ones?

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: There\`s no `\n` in the end of first line, so it\`s just a matter of wrapping the code in 80 symbols ;)
See:
```
>>> sys.stderr.write("(INFO) Config: parameter '%s' (section '%s')"
...                  " is not configured. Using default value: %s\n"
...                  % ('str1', 'str2', 'str3'))
(INFO) Config: parameter 'str1' (section 'str2') is not configured. Using default value: str3
```

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: Still no `\n` in the middle of the message, so it will be output as a single line.
Single and double quotes... er... no idea :) 
I am trying to follow some conventions on this but usually fail...:(
The convention was, as far as I remember, the following:
* use single quotes for 'constants' (like dict keys);
* use double quotes for "strings" (like log messages);
* use double quotes when you need to escape single one and vice versa.

Will fix it. Thank you for pointing :)

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
anastasiakaida: Oh... Now I see. My apologies. :)

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: Done.

Also found missed `\n` in a couple of error messages -- also fixed.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: Forget to update this discussion...
Done (by adding the Python version check).

mgolosova:
Thank you!

PR #110, opened 2018-02-12
SQL: LEFT JOIN error fixed

maria-grigorieva:
The issue is that LEFT JOIN doesn't return all records
from the left table.

The solution is to change WHERE clause to AND
so that the condition is part of the join condition, not a filter applied after the join.
Now all rows in the left table will be present in the result.


maria-grigorieva:
Ok! I've just move AND by rebasing the commit. 
Thank you.

mgolosova:
Ok, thank you!

PR #111, opened 2018-02-15
Travis pep8 update rewrite

Evildoor:
My part of #96 history rewrite.
What was done:

- Removal of unused function containing E741/E722 moved ahead of two commits fixing these errors.
- Variables in xmltable.construct_row() renamed according to suggestion.
- (update) Deletion of comments moved from E741 commit to function deletion commit, message edited accordingly.
- (update) Added clarification for import removal commit.

What was not done:

- Replacement of some cycles with single-line constructs with map/[for]/whatever. This change, while sensible, does not belongs to pull request dealing with code pycodestyle since it actually affects program logic (even if result is the same). This will be done elsewhere.
- Everything else suggested in #96. Either discussion showed that there was no problem or I was not convinced.

### DIFF
--- a/Utils/Dataflow/030_PDFAnalyzer/manager.py
+++ b/Utils/Dataflow/030_PDFAnalyzer/manager.py
@@ -546,71 +546,10 @@ class Paper:
     def delete(self):
         """ Delete all files associated with paper. """
         shutil.rmtree(self.dir)
-# def find_title(self):
-# """ New title determining method, does not works ideally yet.
-# Titles consisting of several lines are problematic to
-# determine.
-# """
-#        lines = self.get_xml_page(1)
-#
-#        d = {}
-# for l in lines:
-#            m = re_xml_symbol.match(l)
-# if m:
-#                size = float(m.group(1))
-#                text = m.group(2)
-# if size in d.keys():
-#                    d[size] += text
-# else:
-#                    d[size] = text
-# elif re_xml_empty_symbol.match(l):
-#                d[size] += " "
-#        xml_title = False
-# print d
-# while True:
-#            size = max(d.keys())
-#            valid = True
-# try:
-# d[size].decode()
-# except:
-#                valid = False
-# if not "atlas note" in d[size].lower() and valid:
-#                xml_title = d[size]
-# break
-# else:
-#                del d[size]
-#
-# print xml_title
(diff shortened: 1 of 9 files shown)


Utils/Dataflow/030_PDFAnalyzer/pdfwork.py
```diff
@@ -45,9 +45,6 @@ def get_page_text(interpreter, page, tmp, rotation=0):
     interpreter.process_page(page)
     tmp.seek(0)
     text = remove_ligatures(tmp.read())
-#    newlines = []
-#    for l in lines:
-#        newlines.append(remove_ligatures(l))
```
mgolosova: I would move "remove commented code" action to another commit -- perhaps to the 1fc48fb (changing its description accordingly), or right after it, what do you think?

Utils/Dataflow/030_PDFAnalyzer/xmltable.py
```diff
@@ -413,54 +405,28 @@ def get_tables_from_text(text):
...
+                    row.append(nline)
+            row.sort(key=lambda nline: nline.center[0])
             rows.append(row)
             t += row
     rows.sort(key=lambda row: row_centery(row))
 
     return rows
-# for row in rows:
-# if row[0].text.isdigit():
-# print "NUM ROW"
```
mgolosova: Same as above, concerning "remove commented code".

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -67,8 +67,7 @@ def path_join(a, b):
 if __name__ == "__main__":
     try:
         import Tkinter
-        from tkFileDialog import askdirectory, askopenfilename,\
-            askopenfilenames, asksaveasfile
+        from tkFileDialog import askdirectory, askopenfilenames, asksaveasfile
```
mgolosova: I still would add to the commit description the information about "which import was removed".
It would be obvious if it was something like
```
- import sys
```
, but it`s not the case. I had to calculate the answer... twice, already ;)

Evildoor:
Done.

mgolosova:
Thank you!
Changes are cherry-picked to PR #96.

PR #115, opened 2018-02-19
DF/09: fix issue with read/write access to the config file.

mgolosova:
Originally we used this file to both read the configuration and to store
the offset value, so we needed to check for the read/write acces.
Currently it is used only to read the initial configuratuion, and thus
the only requirement is to be readable.


maria-grigorieva:
I've fixed 'access' )) can be approved.

mgolosova:
Sorry, I missed the comment of @anastasiakaida :(
@maria-grigorieva , thank you for fixing!

mgolosova:
Fixed 'acces' typo in the commit log message.
Please re-review :)

mgolosova:
Fixed log message about 'read/write access'.

mgolosova:
@anastasiakaida , thank you! Merged.

PR #116, opened 2018-02-19
DF/data4es: add `data4es-status` script.

mgolosova:
It checks current state of the `data4es` dataflow: if it is running
  or not.

---

Required for full automatization of `data4es` dataflow in CERN.


anastasiakaida:
I apologise, I've checked it two days ago and forgot to approve. Everything seems ok.

PR #119, opened 2018-04-10
Merge 'master' into 'pyDKB'

mgolosova:
Two main reasons for this merge:
* there were some changes to pyDKB lib in `master` branch,
* CI (continuous integration) checks has changed.

To go on with pyDKB development, I think it is better to make this merge in advance.

Due to the merge conflict in `pyDKB/dataflow/stage/AbstractStage.py`, the merge is done in a separate PR (instead of a direct PR from `master` to `pyDKB`).

ToDo:
- [x] update DF stages (remove `try/except` around pyDKB `stage`, as it became excessive for now).

PR #120, opened 2018-04-10
[override] pyDKB: separate input/output functionality realization from ProcessorStage.

mgolosova:
One day we want to add Connector stages to the pyDKB, and to avoid:
* re-implementation of data input/output functions from the Processor stage and
* multiple inheritance,

we need to isolate this functionality from Stages at all.

New classes are introduced:
* Stream -- takes care of the protocol support (handling of EOM, EOP, ...);
* Consumer -- takes care of the source detection (files, STDIN, HDFS files, ...);
* Producer -- takes care of the sink detection (files, STDOUT, HDFS files, ...).
---
Waits for #119 
Is splitted into pieces: #123, #124, #125, #126, #127, #134, #129, #131, #132, #133.


mgolosova:
There are some minor differences with what was merged to `pyDKB` with PRs, listed in the PR description, yet it seems like nothing was missed and this PR can be closed.

Issue #91, opened 2017-12-07
Logger for pyDKB


mgolosova:
We need to add some kind of the common logger for the Dataflow stages and the library itself to avoid manual formatting, writing to `sys.stderr`, etc.
The logger should:
* be aware of different log levels (fatal, error, warning, info, trace);
* accept one message or list of messages;
* coordinate formatting of message lines with Kafka ExternalProcessLogger [1];
* never forget about newline in the end of message (as we often do, writing then manually);
* output messages to the STDERR;
* add timing to the messages [2];
* add info about the program produced the message [3].

There are, of course, some loggers for Python that can be used (say, [logging](https://docs.python.org/2/library/logging.html)). They should be studied to see if there\`s a standard solution for we need (I think there must be, as we\`re not looking for something extraordinary), and if found -- introduced to the common workflow. If -- al of a sudden -- nothing appropriate was found, implement the custom logger.

---
[1] The suggestion is to leave already existing agreement about starting the message with smth like `(ERROR)` (or any other log level marker), but also teach Java part to take `(==)` as "same level as above" for multiline messages. Having a lot of `(ERROR)`s in the log looks ugly.

[2] Then we must teach Java part to remove this timing, as it will be added via Java logger as well.

[3] Then Java part must not add the information about the external command, form which STDERR the message was taken.

PR #123, opened 2018-04-20
[stack] pyDKB development (pt.1): unified logging for Stages.

mgolosova:
Logging added in the simpliest possible way -- as a method of parent class for Stages.

It would definitely be better to use some independant "logger" to avoid re-implementation of this method in any other class and for the library as a whole, but it is left for the future (see #91).


mgolosova:
@anastasiakaida , thank you for the appoving -- unfortuntely I can not merge it before #119 (merge of `master` into `pyDKB`). I have fixed that typo you found there, thank you again :)

PR #124, opened 2018-04-20
[stack] pyDKB development (pt.2): improve HDFS module

mgolosova:
Main changes:
* added `path` functionality (using `posixpath` for HDFS-related actions);
* added `File()` function (to simplify work with temporary local copy of HDFS files);
* added `movefile()` function (to simplify writing output files to HDFS).

### DIFF
--- a/Utils/Dataflow/pyDKB/common/hdfs.py
+++ b/Utils/Dataflow/pyDKB/common/hdfs.py
@@ -71,6 +71,17 @@ def putfile(fname, dest):
                             "Error message: %s\n" % (fname, err))
 
 
+def movefile(fname, dest):
+    """ Move local file to HDFS. """
+    if os.path.exists(fname):
+        putfile(fname, dest)
+        try:
+            os.remove(fname)
+        except OSError, err:
+            sys.stderr.write("(WARN) Failed to remove local copy of HDFS file"
+                             " (%s): %s" % (fname, err))
+
+
 def getfile(fname):
     """ Download file from HDFS.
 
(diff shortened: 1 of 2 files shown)


Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -152,3 +153,26 @@ def listdir(dirname, mode='a'):
...
+    if path is None:
+        path = ''
+    return posixpath.basename(path).strip()
+
+
+def dirname(path):
+    """ Return dirname without filename. """
+    if path is None:
+        path = ''
+    return posixpath.dirname(path).strip()
```
anastasiakaida: What will happen if these functions return 'None'?

Because if we set `s = '   '` and then call `s.split()`, calling `if s` returns `False` (just as an example).

mgolosova: It should never return None:
* if passed `path` was None, it is replaced with empty string `''`;
* `posixpath.dirname()` returns empty string `''` in case of empty string in path (and fails in case of non-string value);
* both `strip()` (which is used) and `split()` (typo?) should never return `None` (empty string `''` and empty list `[]` respectively when being applied to empty string).

About your example... I am not sure I understand you correctly. Did you mean the following construction?
```
s = ' '
s = s.split()
if s:
  print 'True'
else:
  print 'False'
```

If so, `if` clause here does the following: first it transforms `s` to `bool`, and then if result bool value is `True` outputs "True", if `False` -- then "False".
And it is not only `None`, that becomes `False` being casting to `bool` -- empty string, list, dict or False itself do the same:
```
>>> bool
<type 'bool'>
>>> bool('')
False
>>> bool(None)
False
>>> bool([])
False
>>> bool({})
False
>>> bool(False)
False
```

anastasiakaida: @mgolosova 
Thank you!
The question now is taken off. :)

PR #125, opened 2018-04-20
[stack] pyDKB development (pt.3): fixes

mgolosova:
Fixed:
* last output file was never moved to HDFS;
* missed logic step in output file definition;
* typo in docstring.

### DIFF
--- a/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
+++ b/Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
@@ -250,7 +250,7 @@ class AbstractProcessorStage(AbstractStage):
             # If something went wrong in `except` clause, we will still
             # get here and return, so the exceptions from there will never
             # reach the user
-            if not isinstance(err, Exception):
+            if err and not isinstance(err, Exception):
                 sys.exit(exit_code)
             return exit_code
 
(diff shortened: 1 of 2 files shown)


Utils/Dataflow/pyDKB/dataflow/stage/AbstractProcessorStage.py
```diff
@@ -450,6 +450,7 @@ def __out_files(self, t='l'):
                 if cf == self.__current_file_full:
                     yield fd
                     continue
+                cf = self.__current_file_full
                 output_dir = self.ARGS.output_dir
                 if not output_dir:
                     output_dir = os.path.dirname(self.__current_file_full)
```
anastasiakaida: Should `cf` be used here too?
(Or this improvement was applied in next PR?)

mgolosova: Instead of `self.__current_file_full`? Yes, it could be... breaks nothing, but a bit ugly, you`re right :)

Can`t say it was really fixed, but definitely gone in [this](https://github.com/PanDAWMS/dkb/pull/134/commits/2ffed05f396b675a0c1ed74c294e142efc4856be) commit of PR #134 (pt.6).

Will do my best to fix it before you\`re done with this PR, but if I can\`t make it -- please, don\`t wait for me :)

PR #126, opened 2018-04-20
[stack] pyDKB development (pt.4): Processor stage improvements.

mgolosova:
Main changes:
* XXXProcessorStage classes are removed in favor of basic `ProcessorStage` + `set_(in|out)put_message_type()`;
* EOP marker is to be output in any processing mode (not only for stream output);
* default output directory changed from the "directory with input files" to `./out/`;

*(update)*
* `flush_buffer()` function clears the buffer after output.

### DIFF
--- a/Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
+++ b/Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
@@ -156,8 +156,7 @@ class ProcessorStage(AbstractStage):
                           nargs='?',
                           help=u'Directory for output files '
                                 '(local or HDFS). ',
-                          default='',
-                          const='output/',
+                          default='out',
                           metavar='DIR',
                           dest='output_dir'
                           )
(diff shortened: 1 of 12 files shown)


Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
```diff
@@ -156,8 +156,7 @@ def defaultArguments(self):
                           nargs='?',
                           help=u'Directory for output files '
                                 '(local or HDFS). ',
-                          default='',
-                          const='output/',
+                          default='out',
                           metavar='DIR',
```
anastasiakaida: I've found a small typo in this commit -- '...ones by defau(l)t'. :)

mgolosova: Right, will fix. Thank you!

However, I have a comment too ;) it is not obvious, which commit you`re talking about, when looking at the "conversation" tab. Here it can be found easily by context (yet still should be found), but in some cases it might not work and will make one to scan through commits or check "blame" version of the file to see where this change belongs, or something.

And your comment is not about the change in the code, but cites some changes (as you used the option of leaving a comment to the code).. it makes one at the first glance to worry about the code and distructs from the real intent of the message :)

PR #130, opened 2018-04-22
[WIP] Connectors' markers -- improving stages 010, 060, 069.

anastasiakaida:
- [x] Stage 010
- [x] Stage 060
- [x] Stage 069

Also:
- [x] Solve an issue: stages' behavior with '\0'.
- [x] Unify code where it is possible.
- [x] Change EOB delimiter in 060/069.

### DIFF
--- a/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
+++ b/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
@@ -20,6 +20,7 @@ OUTPUT
 MARKERS
   -e, --eom      EOM            End-of-message marker.
   -E, --eop      EOP            End-of-process marker.
+
 KERBEROS
   -u, --username USER           Cern account login.
   -r, --retry                   Try to get new Kerberos ticket in case of
--- a/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
+++ b/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
@@ -17,6 +17,9 @@ OUTPUT
                                 from the pipe.
 				If neither FILE nor PIPE is specified, output
                                 goes to STDOUT.
+MARKERS
+  -e, --eom      EOM            End-of-message marker.
+  -E, --eop      EOP            End-of-process marker.
 KERBEROS
   -u, --username USER           Cern account login.
   -r, --retry                   Try to get new Kerberos ticket in case of
--- a/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
+++ b/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
@@ -53,6 +53,14 @@ do
     -t|--tmp)
       CLEAN="NO"
       ;;
+    -e|--eom)
+      EOM="$2"
+      shift
+      ;;
+    -E|--eop)
+      EOP="$2"
+      shift
+      ;;
     -*)
       echo "Unknown option: $key" >&2
       usage >&2
@@ -69,6 +77,16 @@ done
 if [ -z "$USR" ] ; then
   echo "Username is not specified." >&2
   usage >&2
+  exit 1
+fi
+
+if [ -z "$EOM" ] ; then
+  echo "EOM marker is not specified. Exiting." >&2
+  exit 1
+fi
+
+if [ -z "$EOP" ] ; then
+  echo "EOP marker is not specified. Exiting." >&2
   exit 1
 fi
  
--- a/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
+++ b/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
@@ -30,6 +30,8 @@ GENERAL
 "
 }
 
+EOMessage="\n"
+
 while [[ $# > 0 ]]
 do
   key="$1"
@@ -82,16 +84,6 @@ if [ -z "$USR" ] ; then
   usage >&2
   exit 1
 fi
-
-if [ -z "$EOM" ] ; then
-  echo "EOM marker is not specified. Exiting." >&2
-  exit 1
-fi
-
-if [ -z "$EOP" ] ; then
-  echo "EOP marker is not specified. Exiting." >&2
-  exit 1
-fi
  
 if [ -n "$PIPE" ]; then
   if ! ( [ -p "$PIPE" ] || mkfifo "$PIPE" 2>&1 > /dev/null ) ; then
@@ -99,6 +91,21 @@ if [ -n "$PIPE" ]; then
     exit 2
   fi
   FILE="$PIPE"
+fi
+
+if [ -n "$FILE" ] || [ -n "$PIPE" ]
+then
+  EOProcess=""
+else
+  EOProcess="\0"
+fi
+
+[ -n "$EOM" ] && EOMessage="$EOM"
+[ -n "$EOP" ] && EOProcess="$EOP"
+
+if [ -z "$EOMessage" ] ; then
+  echo "EOM marker is not specified. Exiting." >&2
+  exit 1
 fi
 
 GLANCE_REQUEST="https://glance-stage.cern.ch/api/atlas/analysis/papers"
@@ -132,7 +139,7 @@ scp $option $USR@$lxplus:$json $tmp >&2
 
 [ $? ] || { echo "Can not copy file from remote. Exiting." >&2; exit 4; }
 
-echo -ne "$EOM" >> $tmp
+echo -ne "$EOMessage" >> $tmp
 
 if [ -n "$FILE" ] ; then
   cat $tmp > $FILE
@@ -140,7 +147,7 @@ else
   cat $tmp
 fi
 
-echo -ne "$EOP"
+echo -ne "$EOProcess"
 
 [ "$xCLEAN" = "xYES" ] && rm $tmp >&2
 
--- a/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
+++ b/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
@@ -132,7 +132,7 @@ scp $option $USR@$lxplus:$json $tmp >&2
 
 [ $? ] || { echo "Can not copy file from remote. Exiting." >&2; exit 4; }
 
-echo -ne "\n" >> $tmp
+echo -ne "$EOM" >> $tmp
 
 if [ -n "$FILE" ] ; then
   cat $tmp > $FILE
@@ -140,6 +140,7 @@ else
   cat $tmp
 fi
 
+echo -ne "$EOP"
 
 [ "$xCLEAN" = "xYES" ] && rm $tmp >&2
 
--- a/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
+++ b/Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
@@ -67,7 +67,7 @@ do
       shift
       ;;
     -*)
-      echo "Unknown option: $key" >&2
+      echo "(ERROR) Unknown option: $key" >&2
       usage >&2
       exit 1
       ;;
@@ -80,14 +80,14 @@ done
 
 [ -z "$CLEAN" ] && CLEAN="YES"
 if [ -z "$USR" ] ; then
-  echo "Username is not specified." >&2
+  echo "(ERROR) Username is not specified." >&2
   usage >&2
   exit 1
 fi
  
 if [ -n "$PIPE" ]; then
   if ! ( [ -p "$PIPE" ] || mkfifo "$PIPE" 2>&1 > /dev/null ) ; then
-    echo "Can not create a FIFO named pipe: $PIPE. Exiting." >&2
+    echo "(ERROR) Can not create a FIFO named pipe: $PIPE. Exiting." >&2
     exit 2
   fi
   FILE="$PIPE"
@@ -104,7 +104,7 @@ fi
 [ -n "$EOP" ] && EOProcess="$EOP"
 
 if [ -z "$EOMessage" ] ; then
-  echo "EOM marker is not specified. Exiting." >&2
+  echo "(ERROR) EOM marker is not specified. Exiting." >&2
   exit 1
 fi
 
@@ -121,23 +121,23 @@ if ! krenew -H 60 ; then
   if [ -n "$RETRY" ]; then
     kinit "$USR@CERN.CH"
   else
-   echo "No Kerberos ticket for CERN.CH realm found. Exiting." >&2
+   echo "(ERROR) No Kerberos ticket for CERN.CH realm found. Exiting." >&2
    exit 3 
   fi
 fi
-klist | grep 'CERN.CH' >/dev/null || { echo "No Kerberos ticket for CERN.CH realm found. Exiting." >&2; exit 3; }
+klist | grep 'CERN.CH' >/dev/null || { echo "(ERROR) No Kerberos ticket for CERN.CH realm found. Exiting." >&2; exit 3; }
 
 # Lxplus address
 lxplus=lxplus.cern.ch
 
 option="-q -o StrictHostKeyChecking=no -o GSSAPIAuthentication=yes -o GSSAPIDelegateCredentials=yes -o GSSAPITrustDNS=yes"
 ssh $option -K $USR@$lxplus "( $cmd1; $cmd2; ) 2>&1 >/dev/null" 2>&1 > /dev/null
-[ $? ] || { echo "Can not execute remote commands. Exiting." >&2; exit 4; }
+[ $? ] || { echo "(ERROR) Can not execute remote commands. Exiting." >&2; exit 4; }
 
 tmp=/tmp/list_of_papers.json
 scp $option $USR@$lxplus:$json $tmp >&2
 
-[ $? ] || { echo "Can not copy file from remote. Exiting." >&2; exit 4; }
+[ $? ] || { echo "(ERROR) Can not copy file from remote. Exiting." >&2; exit 4; }
 
 echo -ne "$EOMessage" >> $tmp
 
@@ -152,4 +152,4 @@ echo -ne "$EOProcess"
 [ "$xCLEAN" = "xYES" ] && rm $tmp >&2
 
 ssh $option -K $USR@$lxplus "$cmd4;" 2>&1 > /dev/null 
-[ $? ] || { echo "Can not execute remote commands. Exiting." >&2; exit 4; }
+[ $? ] || { echo "(ERROR) Can not execute remote commands. Exiting." >&2; exit 4; }
--- a/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
+++ b/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
@@ -71,7 +71,8 @@ OPTIONS:
 }
 
 upload_files () {
-
+  EOProcess=""
+  
   if [ -z "$1" ] ; then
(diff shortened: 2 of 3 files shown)


anastasiakaida:
@mgolosova 
I chose '\4' for EOB, but in ASCII there is no escape-sequence for 'END OF TRANSMISSION'. Is it ok if I assign '\4' as EOB marker?

anastasiakaida:
@mgolosova 
I made all changes till the checkpoint. Please, check the results. I'm not sure about EOM implementation, but it will be better to check it now. :)

mgolosova:
@anastasiakaida,
sorry I have somehow missed your question about EOB marker.
In general, it is OK to use whatever we choose (as it is our internal protocol and we are free to do whatever we want).
But to make default EOB marker "self-explanatory", I think it would be good to use ETB ("end of transmission block") ASCII symbol (\x17 (hex)).

I will do my best to check the code today and will let you know if I have any comments.

Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
```diff
@@ -93,34 +122,35 @@ if ! krenew -H 60 ; then
...
-echo -ne "\n" >> $tmp
+echo -ne "$EOMessage" >> $tmp
 
 if [ -n "$FILE" ] ; then
   cat $tmp > $FILE
 else
   cat $tmp
 fi
 
+echo -ne "$EOProcess"
```
mgolosova: Remind me about this, please.
Have we decided to send EOP to STDOUT independedntly of the destination of output data (file, pipe, STDOUT, whatever), separating so-called "command channel" from "data channel"?

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -158,14 +170,15 @@ upload_stream () {
...
-    while read -r -d "$delimiter" line; do
+    while read -r -d "$delimiter" line || [[ -n "$line" ]]; do
       n=`ps ax | grep 'curl' | grep "$HOST:$PORT" | grep -v 'grep' | wc -l`
       while [ $n -gt $CURL_N_MAX ]; do
         sleep $SLEEP
         n=`ps axf | grep 'curl' | grep "$HOST:$PORT" | grep -v 'grep' | wc -l`
       done
       echo "$line" | $cmd &>/dev/null || { echo "(ERROR) An error occured while uploading stream data." >&2; continue; } &
-      echo -n $'\6'
+      echo -ne EOMessage
```
mgolosova: `EOMessage` here looks a bit excessive -- Output Connector Stage (here -- Stage 60) sends no data back to supervisor program, so there\`s no messages to be separated.

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -158,14 +170,15 @@ upload_stream () {
...
+      echo -ne EOProcess
```
mgolosova: What happens here when cURL command execution (two lines above) fails?

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -114,16 +123,19 @@ upload_files () {
         ;;
     esac
 
-    eval "$cmd" || { echo "(ERROR) An error occured while uploading file: $INPUTFILE" >&2; continue; }
-
+    eval "$cmd" || { echo "(ERROR) An error occured while uploading file: $INPUTFILE" >&2; continue; } &
+    echo -ne EOMessage
```
mgolosova: EOMessage here looks a bit excessive -- Output Connector Stage (here -- Stage 60) sends no data back to supervisor program, so there`s no messages to be separated.

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -231,8 +252,10 @@ done
 [ -z "$HOST" ] && echo "(ERROR) empty host value." >&2 && exit 2
 [ -z "$PORT" ] && echo "(ERROR) empty port value." >&2 && exit 2
 [ -z "$GRAPH" ] && GRAPH=http://$HOST:$PORT/$GRAPH_PATH
-[ -z "$DELIMITER" ]  && DELIMITER=$'\0'
-[ "x$DELIMITER" = "xNOT SPECIFIED" ] && DELIMITER=$'\n'
+[ -z "$DELIMITER" ]  && DELIMITER=$'\4'
+[ "x$DELIMITER" = "xNOT SPECIFIED" ] && DELIMITER=$'\4'
```
mgolosova: What it special about the `DELIMITER` variable being `"NOT SPECIFIED"` at this point?
I mean, in the earlier versions it used to mean "no batch mode". In this case the script had to handle messages one by one -- that is why  `\n` (hardcoded EOM marker) was used as the delimiter.

If now `"NOT SPECIFIED"` has no special effect, I believe it should be removed at all.

But I think we should keep the possibility to work in "no batch mode". Talking about this -- I think it would be more readable:
* to have some flag option to specify "batch mode on/off" (say, `--batch`);
* to use "empty delimiter" as force "no batch mode" (say, `--eob ''`);
* to use EOM as the delimiter for `read` command in "no bach mode".

It gives us something like this:
```
./uploadTTL.sh <...> --batch                # batch mode with default EOB
./uploadTTL.sh <...>                             # no batch mode
./uploadTTL.sh <...> --eob ''                # no batch mode
./uploadTTL.sh <...> --batch --eob ''   # no batch mode (empty EOB to cancel --batch option)
./uploadTTL.sh <...> --batch '\0'          # batch mode with custom EOB
./uploadTTL.sh <...> --eob '\0'             # batch mode with custom EOB (equivalent to above)
```

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -24,8 +56,10 @@ ES_PORT='9200'
 
 CURL_N_MAX=10
 SLEEP=5
-DELIMITER=`echo -e -n "\x00"`
-EOProcess=`echo -e -n "\x06"`
+EOBatch=$"\4"
```
mgolosova: Why use `$"\4"`?
[Dollar sign + single-quoted string](https://www.gnu.org/software/bash/manual/html_node/ANSI_002dC-Quoting.html) (`$'some string'`) makes escape sequences in the string to be interpreted (that\`s why you have used it in Stage 60 to place code `4` into a variable: `DELIMITER=$'\4'`).
[Dollar sign + double-quoted string](https://www.gnu.org/software/bash/manual/html_node/Locale-Translation.html) (`$"some string"`) will cause the string to be translated according to the current locale. Is that what you mean?

Seems like pure `"\4"` would give exactly the result we need here, and save a reader from looking for some hidden  meaning in it ;)

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -35,25 +69,41 @@ load_files () {
   log "Putting data to ES"
   for INPUTFILE in $*;
   do
-    ${cmd}${INPUTFILE} || exit 3
+    [ ${cmd}${INPUTFILE} ] && echo -ne "$EOMessage" || exit 3
```
mgolosova: We do not need to output EOMessage in Sink Connector stages (just as in both previous comments about this).

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -35,25 +69,41 @@ load_files () {
...
+  else
+    log "Switched to the stream mode."
+    while read -r -d `echo -ne "$EOBatch"` line; do
       n=`ps axf | grep '[c]url' | grep "$HOST:$PORT" | wc -l`
+      while [ $n -gt $CURL_N_MAX ]; do
+        sleep $SLEEP
+        n=`ps axf | grep '[c]url' | grep "$HOST:$PORT" | wc -l`
+      done
+      echo "$line" | ${cmd}- &
+      echo -ne "$EOMessage"
```
mgolosova: EOMessage for Sink Connectors (same comment as for other stages).

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -12,6 +12,38 @@ PARAMETERS:
...
+      EOP="$3"
+      shift
+      ;;
+    -B|--eob)
+      EOB="$3"
+      shift
+      ;;
+    -*)
+      echo "(ERROR) Unknown option: $key" >&2
+      usage >&2
```
mgolosova: As you have added command line parameters, it would be nice to reflect in the usage message. Or else user might stumble like this:
```
$ ./load_data.sh filename -b 1           
(ERROR) Unknown option: -b
USAGE:
load_data.sh [FILE...]

PARAMETERS:
  FILE -- file in NDJSON format for loading to Elasticsearch via bulk interface
```
See, the message that appears after error message does not explain how to use the script properly. The only way to get this information is to read the source code, which is not good.

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -12,6 +12,38 @@ PARAMETERS:
 "
 }
 
+while [[ $# > 1 ]]
+do
+  key="$2"
```
mgolosova: This looks a bit strange, aspecially because you use command `shift` later in the loop.

**About the general logic**

I guess you wanted to leave input filename in the first argument, but the thing is that FILENAME is a) an optional argument (which is reflected in usage massage via square brackets around the argument name "FILENAME") and b) there might be more than one file, which name should be passed to the script (which is reflected by dots after the argument name).
It means, that FILENAME simply can not go as the first argument -- only as the last (or else we will have problems locating and parsing other arguments).

The general rule is to specify first options and parameter arguments ("`--foo`" = option "foo" is on, "`--foo bar`" = parameter "foo" has value "bar"), and only then -- positional arguments, last of which can be optional and/or have multiple values.

That is why in other scripts we have this usage form:
```
script_name [<options>] [FILENAME...]
```

FILENAME here is the last positional argument, that is why it can be omitted (to enter the "stream mode"), or multiple files can be passed to the "file mode".

**About `shift`**

`shift [N]` (with default N=1) makes command line arguments to "move left" on N positions.
Say, you run script with arguments:
```
./my_script arg1 arg2 arg3
```
First your arguments are: "arg1", "arg2", "arg3". Now move them "to the left" on one position. As there is no position "to the left" of the "arg1", it disappears. To the left of "arg2" was "arg1" -- "arg2" goes to its position. Same for "arg3" and position of "arg2". Now the arguments you have are: "arg2", "arg3".

```
$ cat shift_test.sh 
#!/bin/bash

while [ $# -gt 0 ]; do
  echo "Arguments: $*"
  shift
done

$ ./shift_test.sh arg1 arg2 arg3
Arguments: arg1 arg2 arg3
Arguments: arg2 arg3
Arguments: arg3
```

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
mgolosova: And, now it seems to me that it is a bad idea to send evaluation of `$cmd` to the background.
See what can happen:
```
$ sleep 3 && command_that_fails || echo "Failed!" & \
> echo "End of process!"                                                                                       
[1] 16485
(INFO) End of process!
mgolosova@#artois-kiae:~/work/svn/git.dkb-project/Utils/Dataflow/069_upload2es$ command_that_fails: command not found
(ERROR) Failed!
```

We see the message saying that the process is finished (though it still goes on), and when after 3 seconds it fails, we see another message -- saying that the process has failed.

Right, we need to send EOP even when the process failed; but should we send it earlier than the process is finished?
It is a question of parallelization... if we do not send EOP before process is finished, we get no new data to run next process. But if we do, we can miss some records. I need to think about this a little...

For now it is OK to leave things as is; but it should be fixed somehow when we add manual position commit to the supervisor.

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -114,16 +123,19 @@ upload_files () {
         ;;
     esac
 
-    eval "$cmd" || { echo "(ERROR) An error occured while uploading file: $INPUTFILE" >&2; continue; }
-
+    eval "$cmd" || { echo "(ERROR) An error occured while uploading file: $INPUTFILE" >&2; continue; } &
+    echo -ne EOMessage
+    echo -ne EOProcess
```
mgolosova: General comment: dollar sign is missed before variable names here and there (`echo EOProcess` will output `"EOProcess"` string, not the value of the `EOProcess` variable).

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -114,16 +123,19 @@ upload_files () {
         ;;
     esac
 
-    eval "$cmd" || { echo "(ERROR) An error occured while uploading file: $INPUTFILE" >&2; continue; }
-
+    eval "$cmd" || { echo "(ERROR) An error occured while uploading file: $INPUTFILE" >&2; continue; } &
```
mgolosova: If we send the process to the background, we have a risk of running too many cURL processes, sending data to the Virtuoso (each of them trying to send the content of a file, which can be big enough).

We should check a number of cURL processes before starting a new one -- or simply not put the process to the background and handle single file at a time.

To ensure EOP is sent even in case of a failure, it is better to send it before checking the exit code of the $cmd evaluation. For example, like this:
```
eval "$cmd"
res=$?    # Saving exit code
echo $EOProcess
# Check exit code and if it is not 0, output log message and continue with the next file
[ $res -ne 0 ] && { echo "(ERROR) An error occured while uploading file: $INPUTFILE" >&2; continue; }
```

By the way, I am not sure that "continue" is required here (as there is nothing after it)...

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
mgolosova: And, now it seems to me that it is a bad idea to send evaluation of `$cmd` to the background.
See what can happen:
```
$ sleep 3 && command_that_fails || echo "Failed!" & \
> echo "End of process!"                                                                                       
[1] 16485
(INFO) End of process!
mgolosova@#artois-kiae:~/work/svn/git.dkb-project/Utils/Dataflow/069_upload2es$ command_that_fails: command not found
(ERROR) Failed!
```

We see the message saying that the process is finished (though it still goes on), and when after 3 seconds it fails, we see another message -- saying that the process has failed.

Right, we need to send EOP even when the process failed; but should we send it earlier than the process is finished?
It is a question of parallelization... if we do not send EOP before process is finished, we get no new data to run next process. But if we do, we can miss some records. I need to think about this a little...

For now it is OK to leave things as is; but it should be fixed somehow when we add manual position commit to the supervisor.

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -133,7 +133,7 @@ upload_files () {
 upload_stream () {
   EOProcess="\0"
   
-  local delimiter=$'\n'
+  local delimiter=$'\4'
```
mgolosova: As you had to specify the default value multiple times, I think it would be good to move it into some global variable to reuse here and there. Next time we have to change the default value it will save us from a mistake (forgetting to change the value here or there) and will take less time ;)

Utils/Dataflow/069_upload2es/load_data.sh
mgolosova: And... does it work, actually? I have this error in a similar situation:
```
$ cmd="curl http://127.0.0.1:9200/_bulk?pretty --data-binary @"
$ filename="filename"
$ [ ${cmd}${filename} ] && echo 1
-bash: [: too many arguments
```

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
anastasiakaida: @mgolosova 
Thank you. I apologize for that. Will fix it!

Utils/Dataflow/069_upload2es/load_data.sh
mgolosova: As you use `echo -ne "$EOBatch"` later, it was not fatal ;) Yet with `$'\4'` it will work, too.

mgolosova:
The comments to the code above are mostly related to the final state of the code.
More general comments and comments to some commits are below.

db582bf
---

Not sure that I got the commit message, but I believe it has something to do with the issue discussed in [this comment](https://github.com/PanDAWMS/dkb/pull/130/files#r200144111).

19175cd
---
It allows to read the last "line" (remember, that the "line" is actually a batch of messages), but as it has not ended properly with EOB -- should we read it, or skip? It is the same question as we discussed concerning the `custom_readline()` in `pyDKB` library.

There are two points to consider:
* strict realization (that requires EOB in the end of the last batch) allows us to be sure that we have processed the whole batch or none of the batch, and no data were lost; but it also requires the supervisor (that sends the data) to wait for EOP from the worker before it commits the offset (in other words, saves information about the last message successfully processed);
* loose realization (introduced in this commit) allows us to avoid manual commiting of the offset, but can result in missed records (if the record was successfully passed to the worker, but the worker failed before it was processed -- passed to the final storage).

I think that the strict realization is better bor now -- though later we may add possibility to use 'loose' one, when missed records are less of a problem than the performance.

41f8f91
---
Maybe `DELIMITER` variable should be replaced with `EOBatch`, for uniformity and readability?

9fd0d12
---
Default value has changed from two-symbol escape sequence to the one-symbol result of its interpretation, yet the usage of the variable has not changed (`echo -ne "$EOBatch"` is used whenever we need to use this marker). Then what\`s the point of this change? _(see comment on special symbols handling below)_

Special symbols handling
---
There are two approaches to handle special symbols:
* put escape sequences to variables and then interpret them when marker is used;
* put marker symbol into the variable and use it directly.

__Variant 1.__
```
EOProcess="$1"   # Variable contains what was in the command line argument
EOBatch='\4'     # Variable contains two symbols: '\' and '4'
<...>
while read -d $(echo -ne "$EOBatch") line; do
  <...>
  echo -ne "$EOProcess"
done
```

__Variant 2.__
```
EOProcess=$(echo -ne "$1")   # Variable contains result of escape sequences interpretation
                             # in the command line argument
EOBatch=$'\4'     # Variable contains marker symbol itself
<...>
while read -d "$EOBatch" line; do
  <...>
  echo -n "$EOProcess"
done
```

Using both of them, we risk to make a mistake forgetting about `echo` argument `-e`, or about `echo` at all (e.g. in `read -d "$EOBatch"`) where it is required. It also make things harder to read: one has to check how the variable was declared (and re-declared) every now and then to make sure it is used properly.

I think it would be better to choose one approach and stick to it everywhere.
And as the second approach does not work for `'\0'`, `'\n'`, `'\t'` and other whitespace symbols (as shell trims these symbols when command substitution is used (``` `command` ``` or `$(command)`), we should decide on the first of them.

Utils/Dataflow/010_glancePapers/papersFromGLANCE.sh
anastasiakaida: @mgolosova 

While we have only 1 message (so-called JSON-array) EOM and EOP should be just sent after that message. You said that I can write both of them into $tmp-file, (or) but it would be better to leave only EOM in $tmp and to output EOP explicitly (just to be prepared for a case when we'll divide somehow "data channel" and "command channel" . :)

mgolosova: Right, now I remember it. Thank you!

PR #135, opened 2018-05-18
[WIP] pyDKB: common logging system

mgolosova:
Motivation
---
We have individual implementation of logging functionality for some
classes in the library, but there are some disadvantages:
* same functionality is already re-implemented number of times
* ...and must be re-implemented for every module and base class
  in the library;
* hardly extendable to provide convenient interface of 'info()', 'error()',
  etc. methods;
* does not provide filtering by log level (and can hardly be extended
  to be appliable library-wide);
* ...

In other words, there are too many things to be done before current
implementation can become a sound logging system; and as they done,
we might find ourself ending up with something resembling the standard
Python library 'logging'.

However this library does not provide some functionality we would like
to have in our logging system: say, avoiding bare lines (with no
identifiers) -- to simplify parsing its log messages on the supervisor
side -- or TRACE log level.

Implementation
---
Here is introduced a wrapper around standard 'logging' module,
which allows us to implement what we miss in the original module -- and still
use all its advantages.

To have logging configuration in the main library file, we have to include
'common' submodule and configure logging before including any other module;
and to cope with `pycodestyle` we should wrap imports into `try/except` clause
(see E402, module level import not at top of file)

Recommended way to use logging
---
1. Within the library.
```
from pyDKB.common import logging
logger = logging.getLogger(__name__)

...

logger.error("My error message.")
```

2. Outside the library.
```
from pyDKB.common import logging
logger = logging.getLogger("myProgramName")

...

logger.error("My error message.")
```

---
ToDo:
- [x] rethink pyDKB import errors (when to raise ImportError);
- [x] replace logging implementation in base classes with new logging system;
- [ ] update existing Dataflow modules to use stage logger instead of `sys.stderr.write()`.

---
Waiting for #151 (pyDKB stage basic logic fix).

### DIFF
--- a/Utils/Dataflow/pyDKB/common/_logging.py
+++ b/Utils/Dataflow/pyDKB/common/_logging.py
@@ -102,34 +102,32 @@ class MultilineFormatter(logging.Formatter, object):
             format = fmt
         return format, suffix
 
-    def formatExtra(self, lines, suffix=None, prefix="  (==) ", align=False):
+    def formatSuffix(self, record):
+        """ Return formatted suffix. """
+        return self._suffix % record.__dict__
+
+    def formatExtra(self, lines, prefix="  (==) ", suffix='', align=False):
         """ Format extra lines of the log message (traceback, ...).
 
         Parameter 'align' shows whether the suffix should be aligned
         to the right (by the longest line), or to the left (as for normal
         log messages).
         """
-        if suffix is None:
-            suffix = self._suffix
         if isinstance(lines, list) and len(lines):
             max_len = len(max(lines, key=len))
+            # Need to add prefix len (in case that the longest line is
+            # among those that will be prefixed).
+            max_len += len(prefix)
             if suffix and align:
                 line_fmt = "%%(line)-%ds" % max_len
             else:
                 line_fmt = "%(line)s"
-            extra = prefix + line_fmt % {'line': lines[0]} + suffix
+            extra = line_fmt % {'line': lines[0]} + suffix
             for line in lines[1:]:
-                extra += "\n" + prefix + line_fmt % {'line': line} + suffix
+                extra += "\n" + line_fmt % {'line': prefix + line} + suffix
         else:
             extra = ""
         return extra
-
-    def formatException(self, ei):
-        """ Format traceback as extra lines. """
-        s = super(MultilineFormatter, self).formatException(ei)
-        lines = s.splitlines()
-        exc_text = self.formatExtra(lines, align=True)
-        return exc_text
 
     def format(self, record):
         """ Format multiline message.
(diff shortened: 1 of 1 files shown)


Utils/Dataflow/pyDKB/common/_logging.py
```diff
@@ -0,0 +1,286 @@
+"""
+pyDKB.common._logging
+
+A wrapper for standard 'logging' module.
+"""
+
+import logging
+from logging import CRITICAL, FATAL, ERROR, WARNING, WARN, INFO, DEBUG, NOTSET
```
anastasiakaida: @mgolosova 
What is the main difference between "critical" and "fatal" errors (especially in pyDKB)?

mgolosova: @anastasiakaida,
I guess [this](https://github.com/python/cpython/blob/2.7/Lib/logging/__init__.py#L124) is the answer ;)

Utils/Dataflow/pyDKB/common/_logging.py
```diff
@@ -0,0 +1,318 @@
...
+        a true value, e.g.
+
+        logger.trace("Houston, we have a %s", "interesting problem",
+                     exc_info=1)
+        """
+        if self.isEnabledFor(TRACE):
+            self._log(TRACE, msg, args, **kwargs)
+
+    def traceback(self, **kwargs):
+        """ Log traceback without additionat messages with severity 'DEBUG'.
```
Evildoor: "additionaL".

Utils/Dataflow/pyDKB/common/_logging.py
```diff
@@ -0,0 +1,318 @@
...
+        logger.traceback()
+        """
+        if self.isEnabledFor(DEBUG):
+            if not (kwargs.get('exc_info')):
+                kwargs['exc_info'] = 1
+            self.debug('Traceback info:', **kwargs)
+
+
+class RootLogger(Logger):
+    """ Same as Logger, but must must have `Level` and be the only one. """
```
Evildoor: Double "must".

Utils/Dataflow/pyDKB/common/_logging.py
```diff
@@ -0,0 +1,318 @@
...
+        return result
+
+
+# --------------------------
+# Module top-level functions
+# --------------------------
+
+
+def isInitialized():
+    """ Checks if the module (namely, root logger) is initialized. """
```
Evildoor: Maybe "Check if ...", to conform with the other definitions and "Do X" template?

Evildoor:
Per DKB meeting on 15.08:
This pull request was a first attempt on standardizing logging, at some point the work on it was postponed and later #274 was started as a more practical approach. Therefore, #274 should be treated as the up-to-date work on logging and this PR would be left as reminder. In case of #274 being dismissed and/or insufficient, further work on logging may or may not start from this PR.

PR #137, opened 2018-06-07
DF: data4es config lookup

mgolosova:
Sometimes we need to run two or more `data4es` processes with different
configurations (say, regular updates every hour, reverse update for data
repair, weekly update to get updated data from original sources, where
timestamp was not updated properly). To make it possible, we need to at
least be able to set "where to read configuration from" without changing
code.


mgolosova:
@anastasiakaida , thank you for fast response with the review. I`m running final checks and will merge it as they are finished.

PR #138, opened 2018-06-07
data4es: individual home directory for a process.

mgolosova:
To have possibility to run two or more different processes for `data4es`
dataflow process, we need to separate their home directories (where the
process stores temp files, including pidfile).

---
~~Waiting for: #137~~

### DIFF
--- a/Utils/Dataflow/run/data4es-start
+++ b/Utils/Dataflow/run/data4es-start
@@ -5,21 +5,37 @@ BATCH_SIZE=100
 DEBUG=
 
 base_dir=$( cd "$(dirname "$(readlink -f "$0")")"; pwd)
-pidfile=$base_dir/$(basename "$0" | cut -d- -f1).pid
 
 # Directories with configuration files
 [ -n "$DATA4ES_CONFIG_PATH" ] && \
     CONFIG_PATH="$DATA4ES_CONFIG_PATH" || \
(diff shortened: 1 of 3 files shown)


Utils/Dataflow/run/data4es-start
```diff
@@ -5,21 +5,38 @@ BATCH_SIZE=100
...
-pipe_dir=$base_dir/.pipe
+pipe_dir=$HOME_DIR/.pipe
 mkdir -p $pipe_dir
 
-buffer_dir=$base_dir/.buffer
+buffer_dir=$HOME_DIR/.buffer
 mkdir -p $buffer_dir
 
+pidfile="$HOME_DIR/pid"
+echo "$$ main" > $pidfile
```
anastasiakaida: @mgolosova 
Could you explain, please, the main idea of the last string? Can't catch the main idea. :)

mgolosova: @anastasiakaida ,
It is meant to output line with current process ID (see [here](https://www.tutorialspoint.com/unix/unix-special-variables.htm)) and tag word "main" to the file with PIDs.
And I think that you have just saved me from lazyness -- this line is better go to a separate commit (or even another PR), as it was not here before and does not actually fit this PR topic..

mgolosova:
Moved output of main process ID to a separate branch (`data4es-main-pid`), as it is not needed in this PR (and not even used yet). When it is used by status / stop scripts, this change will appear in another PR.

PR #140, opened 2018-07-09
Remove remains of XXXProcessorStages

mgolosova:
Fix two faults of d8a0ba0:
- [x] stages 30, 54 and 55 kept on using obsolete `XXXProcessorStage` classes;
- [x] file with `XXXProcessorStage` classes was not removed.


mgolosova:
@anastasiakaida , thank you for the fast response!

PR #143, opened 2018-07-13
Data4es process for analysis tasks

mgolosova:
How it is different from ordinar `data4es` process:
* analysis tasks are tasks with `pr_id` = 300 (thus different files with SQL queries);
* data should be uploaded to another ES index (thus different ES config file).

~NOTE: if merged after #145 or #146, mapping should be updated.~ (fixed by changing the index names).

ToDo:
- [x] parametric SQL query for 'production' and 'analysis' cases.

---
~Waiting for: #142~


mgolosova:
Sorry for dismissing the review. I realized that the easiest way to make things work (by copying mapping and SQL query files with minimal changes) is not a very good way: it makes one to fix two files when some changes are needed.
So now I`m trying to make things a bit more "universal"...

mgolosova:
Now it must be all. Looking forward for a review! :)

PR #146, opened 2018-07-18
data4es: metadata from Rucio for primary input dataset

mgolosova:
Stage 91 now is able to get metadata from Rucio for both output datasets and primary input dataset.
I have decided not to create a separate stage, as -- logically -- it can be a single stage indeed, but splitting the data flow in two: flow with output datasets metadata and flow with task metadata. It could save us prom transmitting its input data two times, but this bright future can be reached only with new version of communication protocol between stages (pyDKB) and supervisors (Kafka).

For now, Stage 91 is added to the `data4es` process for the second time, as a node of task metadata chain.

~NOTE: if merged after #143, mapping for analysis tasks to be fixed.~ *(fixed in #143)*

---
~Waits for #145~ *(merged)*


Utils/Dataflow/091_datasetsRucio/datasets_processing.py
```diff
@@ -40,16 +40,39 @@
...
 
 def main(argv):
     """ Program body. """
     stage = pyDKB.dataflow.stage.JSONProcessorStage()
+    stage.add_argument('-t', '--dataset-type', action='store', type=str,
+                       help=u'Type of datasets to work with: (i)nput'
+                             ' or (o)utput',
+                       nargs='?',
+                       default=OUTPUT,
+                       choices=[INPUT, OUTPUT],
```
anastasiakaida: As I understood, to set dataset_type we should do something like:

{{{
$ ./datasets_processing.py -t 'i'
}}}

Am I right?

mgolosova: Right. And without quotes it also must work properly :)

PR #153, opened 2018-07-27
ES query for prodtask-dev DKB web-interface

mgolosova:
Suggested query returns total input / output dataset size and events number for tasks by parameters:
* project name;
* current AMI tag of the task;
* output dataset data format.

First it selects all documents of type 'task':
* status: 'done',
* primary_input: 'aod' (dataset name contains this term),
* project: parametric value,
* AMI tag: parametric value

...with following conditions:
* primary_input dataset size > 0,
* among output datasets there is a dataset with data in required format (parametric value), and its size is > 0.

Then found documents are aggregated:
* total input datasets size (sum),
* total input events sumber (sum),
* total output datasets size and events number for:
  * not deleted datasets
  * with data in given format.

### DIFF
--- a/Utils/Elasticsearch/requests/deriv-ratio-q.json
+++ b/Utils/Elasticsearch/requests/deriv-ratio-q.json
@@ -15,7 +15,8 @@ GET tasks_production/task/_search
       "must": [
         {"term": {"primary_input": "aod"}},
         {"term": {"project": %%PROJECT_LOWERCASE%%}},
-        {"term": {"ctag": %%AMI_TAG_LOWERCASE%%}}
+        {"term": {"ctag": %%AMI_TAG_LOWERCASE%%}},
+        {"term": {"status": "done"}}
        ]
      }
    },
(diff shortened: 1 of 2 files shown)


anastasiakaida:
@mgolosova I've tried to use this query (inserting it to the terminal). It works, but I got another results. 

I used values provided in comments. What values were used?

mgolosova:
@anastasiakaida,
Interesting, I have just checked with same parameters and got same result.
What results did you have?

*(By the way, it is useful to specify what "different results" did you have and the exact way you used the query. It will make a long comment, right -- but the question like "what exactly did you do" and "what exactly did you see" are the most popular ones after comments like "I get different result" or "I tried but got an error" ;) So answering them in advance is never a bad idea).*

anastasiakaida:
@mgolosova 

I apologize a lot. That was my fault -- the difference is only in time taken for query processing. Everything else seems the same and works fine.

PR #154, opened 2018-07-30
Stage 030 update

Evildoor:
Improvements for:
- Table detection.
- Year detection.
- Export.
- Interface.
- Help file.

Also update for simple cycles per suggestion in #96 and some other minor changes.


Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -355,19 +353,14 @@ def organize_intervals(intervals):
...
                         # TO DO: improve this.
                         ni2 = []
                         for i1 in ni1:
-                            add_zeros = maxlen - len(i1)
-                            i2 = ""
-                            for j in range(0, add_zeros):
-                                i2 += "0"
-                            i2 += i1
+                            i2 = "".join(["0" for j in
+                                          range(0, maxlen - len(i1))]) + i1
```
mgolosova: Wouldn\`t `"0" * (maxlen - len(i1)) + i1` be more readable?
```
>>> "".join("0" for i in range(0,3))
'000'
>>> "0" * 3
'000'
```

Evildoor: I wasn't aware that such multiplication is possible. True, your suggestion looks better, will do.

Evildoor:
Updated b875f27:
- Improved one cycle per suggestion above.
- Removed a few unnecessary cycles, added this to commit message.

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1826,23 +1826,27 @@ def export_all(self, quick=False, n=None, n_p=None, errors=None, attr=None,
...
                             other = "1"
-                        if not tables and c.startswith("table"):
-                            attr["dataset_tables"].append(p.fname)
-                            tables = "1"
+                        if c.startswith("table"):
+                            tables.append(int(c[len("table_"):]))
+                    if other:
+                        attr["other_datasets"].append(p.fname)
+                    tables.sort()
+                    tables = str(tables).replace(",", "").strip("[]")
```
mgolosova: What about `tables = " ".join(tables)`?

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1826,23 +1826,27 @@ def export_all(self, quick=False, n=None, n_p=None, errors=None, attr=None,
...
                     for a in Paper.attributes_general:
                         if outp["content"]["plain_text"][a]:
                             attr[a].append(p.fname)
-                            if a == "atlas_name":
-                                s += "%s," % outp["content"]["plain_text"][a]
-                            else:
-                                s += "1,"
+                            add = outp["content"]["plain_text"][a]
+                            if a == "campaigns":
+                                add = str(add).replace(",", "").strip("[]")
```
mgolosova: What about `add = " ".join(add)`?

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -1826,23 +1826,27 @@ def export_all(self, quick=False, n=None, n_p=None, errors=None, attr=None,
...
                     s += "%s,%s," % (other, tables)
                     for a in Paper.attributes_general:
                         if outp["content"]["plain_text"][a]:
                             attr[a].append(p.fname)
-                            if a == "atlas_name":
-                                s += "%s," % outp["content"]["plain_text"][a]
-                            else:
-                                s += "1,"
+                            add = outp["content"]["plain_text"][a]
+                            if a == "campaigns":
```
mgolosova: Is it only for "campaigns", or for any `list`?
Maybe use `if isinstance(a, list)` or `if type(a) is list`?
Second variant is more readable, of course, but less universal (it can be another class, inheriting from `list` -- though it is not so often in our code).

mgolosova:
First set of comments.

97d93ed
---
This commit made me think of more DRY code (dont-repeat-yourself).
Maybe not now, but how about move this part of code:
```
if errors:
     keys = errors.keys()
     keys.sort(cmp=cmp_papernames)
     msg = "These papers were not added for some reason:\n\n"
     for e in keys:
         msg += "%s : %s\n\n" % (e, errors[e])
>    with open(ERRORS_FILE, "w") as f:
>        f.write(msg)
?    scrollable_warning(self.window, msg,
?       "Unable to add papers")
```
into a separate method?
Not sure if it must show warning; `ERRORS_FILE` can be replaced with parameter `dest` or `logfile` or whatever.

ff171bb
---
I would split this commit in two: one fixes logic errors, another -- header of STAT_FILE.

8ff3cb6
---
Not sure I see what changed in the logic of `other_datasets` processing.
It matters not how many `other_dataset` were found, `attr["dataset_tables"]` is appended with current paper name and only at the first hit -- in previous version, and in the new one.
There\`s `if not other` condition in the `if` clause, that saves from appending current filename to `attr["dataset_tables"]` more than once.

What have I missed?

87bbe98
---

Shouldn\`t then `"possible_project_realdata"` attr change? If there are two years mentioned, then two suggestions can be done; and if it is a range, like "2015-2017", then for all years in the range?


987803e
---
Is it OK that the new function thinks that 'a-2-1' is greater than 'a-10'?
The example is baseless, of course -- paper names are slightly different -- yet if one paper has extra field, it is still string comparisson.

---

More comments are (possibly) coming; stuck for a while at 9abebea...

mgolosova:
9abebea
---
Quite a large commit... was not easy to get through it, so I think it would be a good idea to split it at least in two: 'header' -> 'caption' change (there\`re a lot of lines in diff related to this changhe, and they distruct reader`s attention from the changes in logic) and logic changes themselves.

* Talking about changes in the logic, there\`s a piece of changes related to the font information taking, but it seems like it is not used anywhere.     I\`d move it to a separate commit with a couple of words of the intention of this change.
```
030_PDFAnalyzer$ git grep '\.font'
xmltable.py:                        self.font = f
```

* In `xmltable.py` two functions: `get_tables_from_text()` and `analyze_page()` have swapped places. Due to this it looks like `get_tables_from_text()` was fully rewritten... but to make sure one has to compare old and new versions side-by-side. Would be better to leave functions at the same order as before, so that there were no doubt in what has happened. If there\`s a reason fow this swap, it is better to do in a separate commit; if not, I think it should be fixed by rebase.

* `analyze_page()` now throws away first row, if there\`s a textline containing `"DRAFT"`. Is it really related to the tables handling?
I mean, it might be required because since now this line can erroneously become a table name, or something -- but it is not obvious from the commit description or the diff, so maybe do it in a separate commit -- after this one (as "fix issue with first DRAFT row"), or before (as "remove non-informative row"), or something like this?

* In `get_tables_from_text()`: what is the magical number "76.0"? It appears again in `show_paper_visual()` method of `Manager` class, but still looks mystical for the one who didn\`t dive deep into the PDF structure and its XML representation (just like the rest of this part of code):
https://github.com/PanDAWMS/dkb/blob/9abebea78149155215e9dc43c8ecc41566398ac5/Utils/Dataflow/030_PDFAnalyzer/xmltable.py#L432-L437

* And again about `get_tables_from_text()`: maybe not now, but I would recommend to split it into smaller pieces of code. For example:
  * mark rows (with 'caption', 'text', 'table')
  * get caption position
  * extract (single) table (or maybe plain and reversed extract in two different functions, if it is not possible to unify the algorithm for both cases)
  * construct caption (from caption rows)

  Maybe even smaller, if needed.
It will make things easier to read and understand -- and easer to test and debug.
These smaller pieces could also be reused by `Manager.show_paper_visual()`: now there\`re two very similar pieces of code, and comparing them between two files to make sure they follow same logic, is not as easy as could be.

* What\`s the point in introducing `caption_rows` in `Manager.show_paper_visual()`? What was wrong, when they were marked with 'red' on the fly?

* Again about `Manager.show_paper_visual()`: red/blue/black marking was changed to red/green/black, but I can hardly guess what it was and what it became. The `blue` variant originally came from the Subversion repo, and -- in theory -- the answer can be found there, but a couple of words in the commit log (or in comments) would make life much easier ;)

Utils/Dataflow/030_PDFAnalyzer/manager.py
```diff
@@ -846,17 +833,16 @@ def export(self, quick=False, outf=False):
...
             (text, datasets) = self.find_datasets()
             for category in datasets:
-                d = []
-                for [name, special] in datasets[category]:
-                    d.append(name)
                 outp["content"][category_export_dict[category]
-                                + "_datasets"] = d
+                                + "_datasets"] = [name for
+                                                  [name, special] in
+                                                  datasets[category]]
```
mgolosova: Isn\`t it too much for a single line of code?..

Maybe leave temporary variable `d`...
```
            for category in datasets:
                d = [name for [name, special] in datasets[category]]
                outp["content"][category_export_dict[category]
                                + "_datasets"] = d
```
...or create another for `"XXX_datasets"`...
```
            for category in datasets:
                content_type = category_export_dict[category] + "_datasets"
                outp["content"][content_type] = \
                    [name for [name, special] in datasets[category]]
```
...or, if temp variables are unacceptable, this formatting seem to work for `pycodestyle` as well...
```
            for category in datasets:
                outp["content"][category_export_dict[category]
                                + "_datasets"] = \
                    [name for [name, special] in datasets[category]]
```

...or, more globally: define "content" as a separate variable `content = {}` and use this variable instead of `outp["content"]` everywhere:
```
        content = {}
        outp["content"] = content
        ...
            for category in datasets:
                content[category_export_dict[category] + "_datasets"] = \
                    [name for [name, special] in datasets[category]]
```
As both `content` and `outp["content"]` are, actually, just links to same `dict` object, and as `dict` in Python is "mutable" object -- when `content` is changed, `outp` is changed as well:
```
>>> c = {}
>>> outp = {}
>>> outp['c'] = c
>>> c['a'] = 123
>>> print outp
{'c': {'a': 123}}
```

Evildoor:
>## 987803e
>Is it OK that the new function thinks that 'a-2-1' is greater than 'a-10'?
The example is baseless, of course -- paper names are slightly different -- yet if one paper has extra field, it is still string comparisson.

Overall - no, it's not OK and this function still has room for improvements. But for now it serves its purpose, since the paper names that I work with are not that diverse, so I think I'll leave it be for now.

>## 9abebea

>Quite a large commit... was not easy to get through it, so I think it would be a good idea to split it at least in two: 'header' -> 'caption' change (there`re a lot of lines in diff related to this changhe, and they distruct reader`s attention from the changes in logic) and logic changes themselves.

Makes sense, will do.

> In xmltable.py two functions: get_tables_from_text() and analyze_page() have swapped places. Due to this it looks like get_tables_from_text() was fully rewritten... but to make sure one has to compare old and new versions side-by-side. Would be better to leave functions at the same order as before, so that there were no doubt in what has happened. If there`s a reason fow this swap, it is better to do in a separate commit; if not, I think it should be fixed by rebase.

They were swapped because get_tables_from_text() now uses analyze_page() - IIRC, in Python you can ignore the rule "define f1() before f2() if f2() uses f1()", but I'm still adhering to it. OK, I'll move the swap to a separate commit then.

>analyze_page() now throws away first row, if there`s a textline containing "DRAFT". Is it really related to the tables handling?
I mean, it might be required because since now this line can erroneously become a table name, or something -- but it is not obvious from the commit description or the diff, so maybe do it in a separate commit -- after this one (as "fix issue with first DRAFT row"), or before (as "remove non-informative row"), or something like this?

Yes, it is related: 

- In old logic, we were simply picking a caption and working with it - if "DRAFT" was encountered here, it was discarded by the mechanism that separates table from the text above it.
- In new logic, we are trying to understand something before looking for captions. And during this research, "DRAFT" can interfere, so it is dealt with explicitly.

I'll add something like this explanation to the commit log.

> In get_tables_from_text(): what is the magical number "76.0"?

Will add a comment to the code.

>And again about get_tables_from_text(): maybe not now, but I would recommend to split it into smaller pieces of code.

Will look into it.

> Everything related to Manager.show_paper_visual()
> including font unused anywhere

As I've said somewhere, including at least one paper, interactive mode of the analyzer is intended for further development and debugging. Now, this function (together with Manager.show_paper_page_tables(), for example) is the quintessence of this point. It's a temporary tool used while blundering about in the mess which are PDF documents in (usually vain) hope of finding some connections and correlations to base extraction logic upon. Hence, such tools are unreliable, prone to constant changes and aren't intended for reasonable usage by others - my apologies that I've not made this clear before and you had to endure the process of trying to judge one of them. What can we do to improve the situation here?

1. Fix the code (I mean code of this function and similar ones), eliminate inconsistencies, etc, and keep doing it in the future. Usually this is the way to go, but in the light of explanation above, I'm not so sure in this case.
2. Remove the code from master and keep it in separate branch(es).
3. Remove the code from github and only upload results.
4. Something else?

All other suggestions and comments not related to 9abebea look justified and I will look further into them.

Once again, thanks for dedicating time and effort to find these problems. Tell me if I've forgotten to reply to something.

Evildoor:
- [ ] Remove debug and research related things from this pull request
- [ ] Swap get_tables_from_text() and analyze_page() in separate commit
- [ ] Move 'header' -> 'caption' change into separate commit
- [ ] Replace complicated constructions with str.join()
- [ ] Split  ff171bb
- [ ] Make sure that extra check in 8ff3cb6 is unnecessary and remove it
- [ ] Change 'possible' attributes in 87bbe98
- [ ] Add an explanation for magical number in text row detection
- [ ] Check long code line in Paper.export()

Optional:
- [ ] Split 9abebea further to improve readability if this remains a problem after changes above
- [ ] Move similar error message related code in Manager.add_paper() and Manager.export_all() into separate function
- [ ] Explore dict[key] = dict.get(key, 0) +1

Evildoor:
This will be done in a separate branch and another pull request, but I'll use and update this checklist.

PR #156, opened 2018-08-08
es/query: add query to get statistics by task steps for hashtag.

mgolosova:
Currently query returns:
* total size of input datasets;
* total number of requested events;
* total size of output datasets;
* total number of events in output datasets;
* number of tasks by task status.

for:
* tasks with given hashtag;
* with all statuses but:
  * aborted;
  * failed;
  * broken;
  * obsolete;
* aggregated by step;
* for not deleted output datasets.

The example response is given for "gammajets" hashtag.

### DIFF
--- a/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
+++ b/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
@@ -36,6 +36,9 @@ GET <index>/task/_search
         "cpu": {
           "sum": {"script": {"inline":"doc['hs06'].value * doc['total_events'].value"}}
         },
+        "walltime": {
+         "sum": {"script": {"inline":"doc['end_time'].value - doc['start_time'].value"}}
+        },
         "output": {
           "children": {"type": "output_dataset"},
           "aggs": {
--- a/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
+++ b/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
@@ -37,7 +37,7 @@ GET <index>/task/_search
           "sum": {"script": {"inline":"doc['hs06'].value * doc['total_events'].value"}}
         },
         "walltime": {
-         "sum": {"script": {"inline":"doc['end_time'].value - doc['start_time'].value"}}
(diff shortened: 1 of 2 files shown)


mgolosova:
Waiting for PR #185.

mgolosova:
Changes were cherry-picked from #185.

mgolosova:
Last changes, requested by M.Borodin, are done.
Please review the PR.

anastasiakaida:
@mgolosova 
Checked! Everything works fine! Thank you!

anastasiakaida:
@mgolosova 
I apologize, but one more thing:

I think (if you agree, of course) it would be nice to add in that sample one more square brackets
```
{
  "size": 0,
  "query": {
    "bool": {
      "must": [
        {"terms": {"hashtag_list": %%HASHTAGS_LOWERCASE%%}},
        {"bool": {"must_not": [{"terms": {"status": ["aborted", "failed", "broken", "obsolete"]}}]}}
      ]
    }
  },
```
The reason is that the query doesn't work without square brackets now at all even there will be only one hashtag.

So it can look like this.
```
{
  "size": 0,
  "query": {
    "bool": {
      "must": [
        {"terms": {"hashtag_list": [%%HASHTAGS_LOWERCASE%%]}},
        {"bool": {"must_not": [{"terms": {"status": ["aborted", "failed", "broken", "obsolete"]}}]}}
      ]
    }
  },
```

mgolosova:
@anastasiakaida,
I thought about it (and even went this way first), however I thought then that these brackets do belong to the 'list' object, not to the query syntax. Just like quote marks belong to the 'string' object (and thus they are never presented around the `%%PLACEHOLDER%%`).

Also, the comment at the top of the query file explicitly says:
```
Parameter values:
  HASHTAGS_LOWERCASE -- list of hashtags (["gammajets", "mc16c_cp"])
```
I count on the fact that those who are going to use this query will read the comment carefully (at least when something goes wrong), or maybe even know something about ES queries (or JSON) in advance.
I may be too optimistic, though...

anastasiakaida:
@mgolosova 

Thank you!

I just tried to simulate the worst scenario. If there are no modifications, the branch can be merged, I think.

PR #157, opened 2018-08-10
[absolete]Connectors' markers implementation -- 010, 060, 069

anastasiakaida:
Sink-connectors: 060, 069 -- support only EOP/EOB.
Source-connector: 010.

Batch-mode option added. That option allows to switch between two states when batch mode is enabled or disabled.
Default value for EOB: ETB (End-of-Transmission-Block character -- '\x17').

### DIFF
--- a/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
+++ b/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
@@ -69,7 +69,8 @@ OPTIONS:
 }
 
 upload_files () {
-
+  EOProcess=""
+  
   if [ -z "$1" ] ; then
     echo "(ERROR) Input file is not specified." >&2
     usage
@@ -119,7 +120,8 @@ upload_files () {
         ;;
     esac
 
-    eval "$cmd" || { echo "(ERROR) An error occured while uploading file: $INPUTFILE" >&2; continue; }
+    eval "$cmd" || { echo "(ERROR) An error occured while uploading file: $INPUTFILE" >&2; } &
+    echo -ne "$EOProcess"
 
   done
 
@@ -128,6 +130,8 @@ upload_files () {
 }
 
 upload_stream () {
+  EOProcess="\0"
+  
   local delimiter=$'\n'
 
   [ -z "$TYPE" ] && { echo "(ERROR) input data format is not specified. Exiting." >&2; return 2;}
@@ -170,7 +174,7 @@ upload_stream () {
         n=`ps axf | grep 'curl' | grep "$HOST:$PORT" | grep -v 'grep' | wc -l`
       done
       echo "$line" | $cmd &>/dev/null || { echo "(ERROR) An error occured while uploading stream data." >&2; continue; } &
-      echo -n $'\6'
+      echo -ne "$EOProcess"
     done
   done
 }
@@ -209,6 +213,10 @@ do
       DELIMITER=`echo -e $2`
       shift
       ;;
+    -E|--eop)
+      EOP="$2"
+      shift
+      ;;
     -u|--user)
       USER="$2"
       shift
@@ -238,6 +246,7 @@ done
 [ -z "$GRAPH" ] && GRAPH=http://$HOST:$PORT/$GRAPH_PATH
 [ -z "$DELIMITER" ]  && DELIMITER=$'\0'
 [ "x$DELIMITER" = "xNOT SPECIFIED" ] && DELIMITER=$'\n'
+[ -n "$EOP" ] && EOProcess="$EOP"
 
 cmdTTL="curl --retry 3 -s -f -X POST --digest -u $USER:$PASSWD -H Content-Type:text/turtle -G http://$HOST:$PORT/sparql-graph-crud-auth --data-urlencode graph=$GRAPH"
 cmdSPARQL="curl --retry 3 -s -f -H 'Accept: text/csv' -G http://$HOST:$PORT/sparql --data-urlencode query"
--- a/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
+++ b/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
@@ -9,7 +9,8 @@ PORT='8890'
 GRAPH=
 GRAPH_PATH='DAV/ATLAS'
 MODE="f"
-DELIMITER="NOT SPECIFIED"
+BATCHMODE="d"
+EOBatch='\x17'
 
 # File .credentials may contain variable definition for USER and PASSWD
 if [ -f ".credentials" ]; then
@@ -132,8 +133,7 @@ upload_files () {
 upload_stream () {
   EOProcess="\0"
   
-  local delimiter=$'\n'
-
+  local delimiter=''
   [ -z "$TYPE" ] && { echo "(ERROR) input data format is not specified. Exiting." >&2; return 2;}
   while [[ $# > 0 ]]
   do
@@ -152,6 +152,8 @@ upload_stream () {
       esac
       shift
   done
+
+  [ -z "$delimiter" ] && { echo "(ERROR) Delimiter is not specified. Exiting." >&2; return 2;}
 
   case $TYPE in
     t|ttl)
@@ -209,8 +211,12 @@ do
       MODE="${2,,}"
       shift
       ;;
-    -d|--delimiter)
-      DELIMITER=`echo -e $2`
+    -b|--batch
+      BATCHMODE="${2,,}"
+      shift
+      ;;
+    -B|--eob)
+      EOBatch=`echo -ne $2`
       shift
       ;;
     -E|--eop)
@@ -244,9 +250,23 @@ done
 [ -z "$HOST" ] && echo "(ERROR) empty host value." >&2 && exit 2
 [ -z "$PORT" ] && echo "(ERROR) empty port value." >&2 && exit 2
 [ -z "$GRAPH" ] && GRAPH=http://$HOST:$PORT/$GRAPH_PATH
-[ -z "$DELIMITER" ]  && DELIMITER=$'\0'
-[ "x$DELIMITER" = "xNOT SPECIFIED" ] && DELIMITER=$'\n'
+
+[ -n "$EOB" ] && EOBatch="$EOB"
 [ -n "$EOP" ] && EOProcess="$EOP"
+
+case $BATCHMODE in
+  e|enabled)
+    [ -z "$EOB" ] && EOBatch="\n"
+    ;;
+  d|disabled)
+    ( [ -z "$EOB" ] || [ $EOB == "\x17" ] || [ $EOBatch == "\x17" ] ) && EOBatch="\n"
+    ( [ -n "$EOB" ] && [ ! $EOB == "\x17" ] ) && EOBatch="$EOB"
+    ;;
+  *)
+    echo "(ERROR) Unexpected batch-mode parameter." >&2 && { [ -n "$BATCHMODE" ] && usage && return 2; }
+    break
+    ;;
+esac
 
 cmdTTL="curl --retry 3 -s -f -X POST --digest -u $USER:$PASSWD -H Content-Type:text/turtle -G http://$HOST:$PORT/sparql-graph-crud-auth --data-urlencode graph=$GRAPH"
 cmdSPARQL="curl --retry 3 -s -f -H 'Accept: text/csv' -G http://$HOST:$PORT/sparql --data-urlencode query"
@@ -256,7 +276,7 @@ case $MODE in
     upload_files $*;
     ;;
   s)
-    upload_stream -d "$DELIMITER";
+    upload_stream -d "$EOBatch";
     ;;
   *)
     echo "(ERROR) $MODE: unsupported mode."  >&2
--- a/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
+++ b/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
@@ -55,10 +55,15 @@ OPTIONS:
                                 * in a (s)tream mode waits for input from STDIN,
                                   sending data to Virtuoso;
                                   messages are to be delimited by <delimiter>.
-  -d, --delimiter <delimiter>   Specifies the delimiter between sets of input
+  -b, --batch {e[abled]|d[isabled]} Specifies batch-mode: (e)nabled|(d)isabled.
+  -B, --eob <EOB>               Specifies the delimiter between sets of input
                                 data in the stream mode.
                                 Default: \n
-                                Kafka-style: \0
+                                Kafka-style: \x17
+  -E, --eop <EOP>               Specifies end-of-process marker.
+                                Default:
+                                * in a (f)ile mode: none
+                                * in a (s)tream mode: \0
   -h, --help                    Print this message and exit.
 "
 }
--- a/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
+++ b/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
@@ -163,7 +163,7 @@ upload_stream () {
   esac
 
   while true; do
-    while read -r -d "$delimiter" line; do
+    while read -r -d "$delimiter" line || [[ -n "$line" ]]; do
       n=`ps ax | grep 'curl' | grep "$HOST:$PORT" | grep -v 'grep' | wc -l`
       while [ $n -gt $CURL_N_MAX ]; do
         sleep $SLEEP
--- a/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
+++ b/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
@@ -10,14 +10,14 @@ GRAPH=
 GRAPH_PATH='DAV/ATLAS'
 MODE="f"
 BATCHMODE="d"
-EOBatch='\x17'
+EOB='\x17'
+EOBatch="$EOB"
 
 # File .credentials may contain variable definition for USER and PASSWD
 if [ -f ".credentials" ]; then
   USER=`sed -n 1p .credentials`
   PASSWD=`sed -n 2p .credentials`
 fi
-
 
 SLEEP=1
 CURL_N_MAX=10
@@ -59,8 +59,8 @@ OPTIONS:
   -b, --batch {e[abled]|d[isabled]} Specifies batch-mode: (e)nabled|(d)isabled.
   -B, --eob <EOB>               Specifies the delimiter between sets of input
                                 data in the stream mode.
-                                Default: \n
-                                Kafka-style: \x17
+                                Default: \x17
+                                No batch-mode: \n
   -E, --eop <EOP>               Specifies end-of-process marker.
                                 Default:
                                 * in a (f)ile mode: none
@@ -169,7 +169,7 @@ upload_stream () {
   esac
 
   while true; do
-    while read -r -d "$delimiter" line || [[ -n "$line" ]]; do
+    while read -r -d $(echo -ne "$delimiter") line || [[ -n "$line" ]]; do
       n=`ps ax | grep 'curl' | grep "$HOST:$PORT" | grep -v 'grep' | wc -l`
       while [ $n -gt $CURL_N_MAX ]; do
         sleep $SLEEP
@@ -216,7 +216,7 @@ do
       shift
       ;;
     -B|--eob)
-      EOB=`echo -ne $2`
+      EOB="$2"
       shift
       ;;
     -E|--eop)
@@ -255,10 +255,10 @@ done
 [ -n "$EOP" ] && EOProcess="$EOP"
 
 case $BATCHMODE in
-  e|enabled)
+  "e"|"enabled")
     [ -z "$EOB" ] && EOBatch="\n"
     ;;
-  d|disabled)
+  "d"|"disabled")
     ( [ -z "$EOB" ] || [ $EOB == "\x17" ] || [ $EOBatch == "\x17" ] ) && EOBatch="\n"
     ( [ -n "$EOB" ] && [ ! $EOB == "\x17" ] ) && EOBatch="$EOB"
     ;;
--- a/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
+++ b/Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
@@ -211,12 +211,12 @@ do
       MODE="${2,,}"
       shift
       ;;
-    -b|--batch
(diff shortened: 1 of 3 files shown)


anastasiakaida:
@mgolosova 

I've met some collisiong in #130, so I decided to close the previous one and open a new one. Please, notice me if I forgot to fix something.

PS: I still don't know that to do with the last line fix in 19175cd. I would like to use 'strict realization', but I'm not sure what to do firstly. Should I leave it as it is now or make some improvements?

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -209,8 +211,12 @@ do
       MODE="${2,,}"
       shift
       ;;
-    -d|--delimiter)
-      DELIMITER=`echo -e $2`
+    -b|--batch
+      BATCHMODE="${2,,}"
```
mgolosova: Why `"${2,,}"`, not simply `"$2"`?

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -244,10 +250,24 @@ done
 [ -z "$HOST" ] && echo "(ERROR) empty host value." >&2 && exit 2
 [ -z "$PORT" ] && echo "(ERROR) empty port value." >&2 && exit 2
 [ -z "$GRAPH" ] && GRAPH=http://$HOST:$PORT/$GRAPH_PATH
-[ -z "$DELIMITER" ]  && DELIMITER=$'\0'
-[ "x$DELIMITER" = "xNOT SPECIFIED" ] && DELIMITER=$'\n'
+
+[ -n "$EOB" ] && EOBatch="$EOB"
```
mgolosova: What the `EOB` variable is used for? It is not assigned anywhere in the code...

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -209,8 +211,12 @@ do
...
       shift
       ;;
-    -d|--delimiter)
-      DELIMITER=`echo -e $2`
+    -b|--batch
+      BATCHMODE="${2,,}"
+      shift
+      ;;
+    -B|--eob)
+      EOBatch=`echo -ne $2`
```
mgolosova: I believe we have quite a long discussion about this mistake I\`ve done when wrote this script :)
What happens when `echo -ne $2` is used?
1. If `$2` was `"\0"` or `"\n"`, the value of `EOBatch` will be the empty string, as shell trims all empty/space symbols from command substitution result. See in the example with `\n`:
```
[mgolosova@bamboo dkb]$ echo -e "abcde\n\n\n"                                                                                                                                                                
abcde



[mgolosova@bamboo dkb]$ echo `echo -ne "abcde\n\n\n"`
abcde
[mgolosova@bamboo dkb]$ 
```

2. Even if `$2` was not a code of "empty/space" symbol: here you get the interpreted value, while the default value is not interpreted: `EOBatch='\x17'`. What do you pass to `upload_stream`? Interpreted value or not? If not -- what will be used as a delimiter in `read -r -d "$delimiter"`?
See:
```
[mgolosova@bamboo dkb]$ EOBatch='\x17'                                                                                                                                                                             
[mgolosova@bamboo dkb]$ echo -ne "message_1\nmessage_2\n${EOBatch}message_3\nmessage_4\n${EOBatch}" \
>   | hexdump -C
00000000  6d 65 73 73 61 67 65 5f  31 0a 6d 65 73 73 61 67  |message_1.messag|
00000010  65 5f 32 0a 17 6d 65 73  73 61 67 65 5f 33 0a 6d  |e_2..message_3.m|
00000020  65 73 73 61 67 65 5f 34  0a 17                    |essage_4..|
0000002a
[mgolosova@bamboo dkb]$ echo -ne "message_1\nmessage_2\n${EOBatch}message_3\nmessage_4\n${EOBatch}" \
>   | while read -r -d "$EOBatch" line; do echo $line; done
[mgolosova@bamboo dkb]$ echo -n "message_1\nmessage_2\n${EOBatch}message_3\nmessage_4\n${EOBatch}" \
>   | while read -r -d "$EOBatch" line; do echo $line; done                                                         
message_1
nmessage_2
n
x17message_3
nmessage_4
n
[mgolosova@bamboo dkb]$ echo -ne "message_1\nmessage_2\n${EOBatch}message_3\nmessage_4\n${EOBatch}" \
>   | while read -r -d "`echo -ne $EOBatch`" line; do echo $line; done                                             
message_1 message_2
message_3 message_4
[mgolosova@bamboo dkb]$ 
```

Resume:
* variable should always contain interpreted or not interpreted value, whenever it was (re)assigned;
* to `read` should be passed interpreted value;
* we have to do something with "empty/space" symbols: `read` uses `\0` as delimiter if empty string is passed, but I have no idea right now how to pass it `\n` :)

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -44,36 +67,65 @@ ES_PORT='9200'
...
+    ;;
+  d|disabled)
+    ( [ -z "$EOB" ] || [ $EOB == "\x17" ] || [ $EOBatch == "\x17" ] ) && EOBatch="\n"
+    ( [ -n "$EOB" ] && [ ! $EOB == "\x17" ] ) && EOBatch="$EOB"
+    ;;
+  *)
+    log ERROR "Unexpected batch-mode parameter." && { [ -n "$BATCHMODE" ] && usage && return 2; }
+    break
+    ;;
+esac
```
mgolosova: Not sure I\`m getting this `case` statement.
What makes me puzzled:
* `BATCHMODE` is just assigned with `"d"` (rewriting user value); why use `case $BATCHMODE`, when we already know the value for sure?
* `EOBatch` is just assigned with default value `'\x17'`; why compare it with `"\x17"` again?
* after this `case`, `EOBatch` will stay assigned with `'\x17'` only when batchmode is `enabled` and `EOB` is not empty;
* after this `case`, `EOBatch` will be assigned with value of `EOB` only when batchmode is `disabled` and `EOB` is not empty and not `"\x17"`.

What did you mean to do?

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -231,8 +250,23 @@ done
...
+[ -n "$EOB" ] && EOBatch="$EOB"
+[ -n "$EOP" ] && EOProcess="$EOP"
+
+case $BATCHMODE in
+  e|enabled)
+    [ -z "$EOB" ] && EOBatch="\n"
+    ;;
+  d|disabled)
+    ( [ -z "$EOB" ] || [ $EOB == "\x17" ] || [ $EOBatch == "\x17" ] ) && EOBatch="\n"
+    ( [ -n "$EOB" ] && [ ! $EOB == "\x17" ] ) && EOBatch="$EOB"
```
mgolosova: Not sure I\`m getting this `case` statement properly... It is similar to the one puzzled me in `069_upload2es/load_data.sh`, but here we have slightly different initial conditions...

First, what does it say now (suggesting, that `EOB` should be used as `EOP`, for custom value passed with `--eob`):
"If batch mode enabled, EOBatch is `\n` unless EOB is defined. If batch mode is disabled, EOBatch is `\n` unless EOB is defined with anything but `\x17` -- else EOBatch is EOB." 

Concerning conditions... as we have this lines above:
```
...
    -B|--eob)
      EOB=`echo -ne $2` # allowed myself to replace EOBatch with EOB
...
[ -n "$EOB" ] && EOBatch="$EOB"
```
within the `case`:
* if `EOB` is empty, `EOBatch` should be `"\x17"` (the default value);
* if `EOB` is `"\x17"` (or, better to say, if user passed `--eob \x17` in the command line), then `EOB` is not *string* `"\x17"`, but ASCII code ETB;
* if, after all, `EOB` *is* string `"\x17"`, then `EOBatch` is `"\x17"` too.

So, in plain words -- what should be done during this `case`?

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -19,14 +30,26 @@ ES_CONFIG="${base_dir}/../../Elasticsearch/config/es"
...
+      usage
+      exit;;
+    -E|--eop)
+      EOP="$2"
+      shift;;
+    -B|--eob)
+      EOB="$2"
+      shift;;
+    -b|--batch)
+      BATCHMODE="${2,,}"
```
mgolosova: Again, why `"${2,,}"`, not simply `"$2"`?
There seem to be no difference:
```
[mgolosova@bamboo dkb]$ f () {
>   echo "'${1,,}'"
>   echo "'${1}'"
> }
[mgolosova@bamboo dkb]$ f enabled
'enabled'
'enabled'
[mgolosova@bamboo dkb]$ f e
'e'
'e'
[mgolosova@bamboo dkb]$ f
''
''
```

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -44,36 +67,65 @@ ES_PORT='9200'
...
+    log ERROR "Unexpected batch-mode parameter." && { [ -n "$BATCHMODE" ] && usage && return 2; }
+    break
+    ;;
+esac
 
 cmd="curl $ES_AUTH http://$ES_HOST:$ES_PORT/_bulk?pretty --data-binary @"
 
 load_files () {
-  [ -z "$1" -o ! -f "$1" ] && log $(usage) && exit 1
+  [ -z "$1" -o ! -f "$1" ] && log NOTSET $(usage) && exit 1
```
mgolosova: Why `NOTSET`? I\`d say `INFO` or `ERROR` or, if the idea is to output `usage` without log level -- use `usage >&2`, plain `usage`, or maybe even teach `log` to work with skipped level parameter...

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -55,10 +55,15 @@ OPTIONS:
                                 * in a (s)tream mode waits for input from STDIN,
                                   sending data to Virtuoso;
                                   messages are to be delimited by <delimiter>.
-  -d, --delimiter <delimiter>   Specifies the delimiter between sets of input
+  -b, --batch {e[abled]|d[isabled]} Specifies batch-mode: (e)nabled|(d)isabled.
+  -B, --eob <EOB>               Specifies the delimiter between sets of input
                                 data in the stream mode.
                                 Default: \n
-                                Kafka-style: \0
+                                Kafka-style: \x17
```
mgolosova: I would say that as "batch mode" is formalized and properly defined, there\`s no such thing as "Kafka style" any longer. Kafka style was informal description of "batch mode", as Kafka-supervised processing was the only case when batch mode was used.

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -163,7 +163,7 @@ upload_stream () {
   esac
 
   while true; do
-    while read -r -d "$delimiter" line; do
+    while read -r -d "$delimiter" line || [[ -n "$line" ]]; do
```
mgolosova: Concerning your question: I think, if we want to work in "strict" mode (no EOM -- no message, no EOB -- no batch), then this change is not required.

Yet thank you for noticing: I have missed this issue when wrote the code.

By the way: if this commit was to be kept, there must be issue description in the commit log, right? ;)

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -69,7 +69,8 @@ OPTIONS:
 }
 
 upload_files () {
-
+  EOProcess=""
+
```
mgolosova: `EOProcess` is explicitly defined in both `upload_files` and `upload_stream`. How user value is supposed to be used?

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
```diff
@@ -209,8 +211,12 @@ do
       MODE="${2,,}"
       shift
       ;;
-    -d|--delimiter)
-      DELIMITER=`echo -e $2`
+    -b|--batch
```
mgolosova: Bracket is missed:
`-b|--batch)`

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -10,6 +10,17 @@ $(basename "$0") [-c CONFIG] [FILE...]
 PARAMETERS:
   CONFIG -- configuration file
   FILE   -- file in NDJSON format for loading to Elasticsearch via bulk interface
+
+OPTIONS:
+  -b, --batch {e[abled]|d[isabled]} Specifies batch-mode: (e)nabled|(d)isabled.
```
mgolosova: I think it could be handy to use this option as a flag: presented = enabled, missed = disabled.
Yet to make it possible to use it with value, we need a bit of invention in parsing:
```
BATCHMODE='disabled'
case $1 in
  ...
  -b|--batch)
    if [ -z "$2" ] || [[ "$2" == -* ]]; # it is last command line argument, or the next one starts with '-'
    then
      BATCHMODE='enabled'
    else
      BATCHMODE="$2"
      shift
    fi
    ;;
  ...
esac
```

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -19,14 +30,26 @@ ES_CONFIG="${base_dir}/../../Elasticsearch/config/es"
 
 while [ -n "$1" ]; do
   case "$1" in
-    --config|-c)
+    -c|--config)
```
mgolosova: This change must be done in a separate commit: it has nothing to do with custom EOB/EOM, it is just formatting.

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -19,14 +30,26 @@ ES_CONFIG="${base_dir}/../../Elasticsearch/config/es"
...
+      shift;;
+    -b|--batch)
+      BATCHMODE="${2,,}"
+      shift;;
     --)
       shift
       break;;
     -*)
-      echo "Unknown option: $1" >&2
+      log ERROR "Unknown option: $key" $(usage)
```
mgolosova: 1. This change has nothing to do with custom EOB/EOP.
2. `$key` variable is not defined.

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -35,7 +58,7 @@ while [ -n "$1" ]; do
   shift
 done
 
-log "Loading defaults and config $ES_CONFIG if any"
+log INFO "Loading defaults and config $ES_CONFIG if any"
```
mgolosova: This change has nothing to do with custom EOB/EOP.

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -44,36 +67,65 @@ ES_PORT='9200'
...
+esac
 
 cmd="curl $ES_AUTH http://$ES_HOST:$ES_PORT/_bulk?pretty --data-binary @"
 
 load_files () {
-  [ -z "$1" -o ! -f "$1" ] && log $(usage) && exit 1
+  [ -z "$1" -o ! -f "$1" ] && log NOTSET $(usage) && exit 1
 
-  log "Putting data to ES"
+  log INFO "Putting data to ES"
```
mgolosova: This change has nothing to do with custom EOB/EOP.

Utils/Dataflow/069_upload2es/load_data.sh
anastasiakaida: @mgolosova 

Should it appear in another PR (I'm thinking about to move everything related with log() into another PR and formalize all mentioned connectors)?

mgolosova: Hm, separate PR is a great idea.
I always miss the moment when a set of small changes becomes not so small and deserves personal branch...

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
anastasiakaida: @mgolosova 

I found an implicit mistake in my code because of that question. Thank you!

The thing is that I named all variables which keeps parameters from a comand line like EOM/EOP/EOB (my mistake is that I named it as "EOBatch"). In 010 connector I saw that it would be better to pass markers from another variables EOMessage/EOProcess, because it was a little bit inconvenient  for me (according to the code's logic) to use variables EOM/EOP.

I hope, after renaming it'll become much more clear. I'll create an extra commit with explanation of my decision.

Utils/Dataflow/069_upload2es/load_data.sh
anastasiakaida: @mgolosova 

I couldn't decide which level (`ERROR` or `INFO`) that case belongs to. I decided to assign to it level 0 (NOTSET). I took into account PR #135, so why not to mention NOTSET directly and allow to handle log with skipped parameter as level 0?

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
anastasiakaida: @mgolosova 

Got it. I suppose, it would be better to represent it as:

`Dafault: \x17
No batch-mode: \n`

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
anastasiakaida: @mgolosova 

I supposed that `case` allows here to check the following conditions:

if batch mode is disabled (by hands or by default):
1. no EOB = `-b ''` -->  `$EOB='\n'`
2. we have EOB, but \x17 (hex) --> `$EOB='\n'` (it might be unnecessary and too strict, but I'm not sure)
3. we have any other EOB --> `$EOB='EOB'`

if batch mode is enabled: 
1. only if we type `--eob ''` --> `$EOB='\n'`

Utils/Dataflow/069_upload2es/load_data.sh
anastasiakaida: @mgolosova 
That looks very useful. I think, I should apply it right now. Thank you! :)

Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh
mgolosova: Alright. I think there\`s a little bit of mess happened with enabled/disabled batch mode etc...

---

First, about this (just to make sure we do understand things same way).
@anastasiakaida wrote:
> we have EOB, but \x17 (hex) --> $EOB='\n' (it might be unnecessary and too strict, but I'm not sure)

This checks:
```
[ ! $EOB == "\x17" ]
[ $EOB == "\x17" ]
```
are string comparison. Works this way:
```
mgolosova@artois-K:~$ a_str="\x17"
mgolosova@artois-K:~$ a_ascii=$'\x17'
mgolosova@artois-K:~$ echo "str: '$a_str'; ASCII: '$a_ascii'"
str: '\x17'; ASCII: ''
mgolosova@artois-K:~$ [ "$a_str" == "\x17" ] && echo "True" || echo "False"
True
mgolosova@artois-K:~$ [ "$a_ascii" == "\x17" ] && echo "True" || echo "False"
False
 ```
Now `EOB` is assigned with string value, so it\`s OK; just wanted to make things clear.

---

Next, about the general logic.
I have tested it with attached script: [test_batchmode.txt](https://github.com/PanDAWMS/dkb/files/2289531/test_batchmode.txt)

Here\`s the results:
```
   -b   ||  X   |  X   |   X    |   X    || 'e'  | 'e'  |  'e'   |  'e'   |
------- || ---- | ---- | ------ | ------ || ---- | ---- | ------ | ------ | 
   -B   ||  X   |  ''  | '\x17' | '\x11' ||  X   |  ''  | '\x17' | '\x11' |
======= || ==== | ==== | ====== | ====== || ==== | ==== | ====== | ====== |
EOBatch || '\n' | '\n' |  '\n'  | '\x11' || '\n' | '\n' | '\x17' | '\x11' |
```

How do I read it: batch mode is actually enabled in three cases:
* `-B 'non-default marker'`;
* `-b -B 'default marker'`;
* `-b -B 'non-default marker'`.

While I would expect the following table:
```
   -b   ||  X   |  X   |   X    |   X    ||  'e'   | 'e'  |  'e'   |  'e'   |
------- || ---- | ---- | ------ | ------ || ------ | ---- | ------ | ------ | 
   -B   ||  X   |  ''  | '\x17' | '\x11' ||   X    |  ''  | '\x17' | '\x11' |
======= || ==== | ==== | ====== | ====== || ====== | ==== | ====== | ====== |
EOBatch || '\n' | '\n' | '\x17' | '\x11' || '\x17' | '\n' | '\x17' | '\x11' |
```
Two cases added:
* `-b`;
* `-B 'default marker'`.

In other words:
* it shouldn\`t matter if user passed default marker or not: if it is passed implicitly, it is preferable value;
* if user said "enable batch mode", then we should use default EOB for batch mode.

Comment (mgolosova) [reply Utils/Dataflow/060_upload2virtuoso/uploadTTL.sh]: *) And about forgotten quotes: it was totally OK; shell takes non-quoted stuff as a string unless it is a keyword or the string is followed by `=` (meaning that it is variable assignment). Quotes are required for multi-word strings (though they could be written without quotes too, if spaces are escaped) and to escape special characters.
See:
```
mgolosova@artois-K:~$ a=abcdef
mgolosova@artois-K:~$ echo $a
abcdef
mgolosova@artois-K:~$ echo abcdef
abcdef
mgolosova@artois-K:~$ a=abc def
def: command not found
mgolosova@artois-K:~$ a=abc\ def
mgolosova@artois-K:~$ echo $a
abc def
mgolosova@artois-K:~$ echo abc\def
abcdef
mgolosova@artois-K:~$ echo abc\ def
abc def
mgolosova@artois-K:~$ echo "abc\ def"
abc\ def
mgolosova@artois-K:~$ echo "abc def"
abc def
```

So in `case` it is OK to write strings without quotes (as values, not keywords or something are expected); moreover, if a wildcard '*' is used, quotes will escape it and it will be taken as an asterisk character, not a wildcard:
```
mgolosova@artois-K:~$ a=abcd
mgolosova@artois-K:~$ case $a in
> "a*")
>   echo 'Quoted case: "a*"';;
> a*)
>   echo 'Non quoted case: a*';;
> *)
>   ;;
> esac
Non quoted case: a*
```

Yet in this case quotes are also OK to use, as no wildcard or something is used.

anastasiakaida:
This PR is the older one version of PR#179.
This PR will be closed and saved till PR#179 will be undone or closed without merge.

PR #158, opened 2018-08-17
Documentation for DKB project.

mgolosova:
We have docstrings in code, but never tried to build some documentation from them.

Now docstrings for pyDKB are transformed to HTML and PDF versions of the documentation (let me know if other formats look useful for you). Current version is full of artefacts due to the wrong formatting and can be not very useful right now, but gives us starting point to fix it and make docstrings more informative.

To rebuild documentation, run in `/Docs` directory:
```
$ make clean
$ make all
```

To rebuild only PDF or HTML version, replace "all" with "pdf" or "html".

Previous content of the `/Docs` directory is moved to `/Docs/misc`.
I believe it is to be stored at least for history; new documents can also be added there.

---

NOTE:
 * to generate docs, Sphinx is required:
     `pip install -U Sphinx`
 * to generate PDF version, pdflatex is required

---

Although the documentation is still to be improved, I suggest to merge it into `master` -- so that we could go on the work basing on the [most recent version of the `pyDKB` library](https://github.com/PanDAWMS/dkb/tree/pyDKB): there are a lot of changes comparing to the `master` version, so even if we have `master` version fully documented, there will be a lot of things removed and requiring update immediately after merge. So let\`s not vaste our time on them :)


Utils/Dataflow/pyDKB/__init__.py
```diff
@@ -1,5 +1,120 @@
...
+(transformation) and *sink connector* (load to internal DKB storage).
+Currently ``pyDKB`` library can be used only for *processor* stages, but
+in future versions *connector* stages will also be supported.
+
+Quickstart guide
+----------------
+
+To create simple processor stage application first decide input and
+output data format. In following examples we will work with data in JSON
+format (for the full list of suppurted formats check
```
anastasiakaida: typo: 'suppOrted"

Utils/Dataflow/pyDKB/__init__.py
```diff
@@ -1,5 +1,120 @@
...
+In our example, messages without key ``'name'`` will produce no output
+messages, so ``stage.output()`` will not be called at all.
+In terms of data flow it means that the input message is filtered out
+and will not reach the *sink connector*..
+
+Boolean return value of ``my_process()`` indicates if the processing was
+successful or not.
+If processing failed (``False`` is returned), produced output messages
+will be dropped to avoid loading into the DKB storages sketchy
+information.
```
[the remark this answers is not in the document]
mgolosova: You mean 
> If processing **fails (and returns ``False``)**, produced output...

?

Utils/Dataflow/pyDKB/__init__.py
```diff
@@ -1,5 +1,120 @@
...
+``JSONMessage(out_data)`` statements are used to decode/encode message
+to/from Python ``dict`` object.
+Message, passed to the function, is taken from the input data flow; to
+write new message(s) to the output data flow,
+``stage.output(out_message)`` is used. It can be used as many times as
+many output messages were generated (or once with the list of messages).
+In our example, messages without key ``'name'`` will produce no output
+messages, so ``stage.output()`` will not be called at all.
+In terms of data flow it means that the input message is filtered out
+and will not reach the *sink connector*..
```
Evildoor: Double dot?

Utils/Dataflow/pyDKB/__init__.py
Evildoor: Word order: "... will be dropped to avoid loading sketchy information into the DKB storages".

Utils/Dataflow/pyDKB/__init__.py
```diff
@@ -1,5 +1,120 @@
...
+  if __name__ == '__main__':
+      stage = JSONProcessorStage()
+      stage.process = my_process
+      stage.parse_args(sys.argv[1:])
+      stage.run()
+
+First we create stage object and indicate that input and output message
+format is JSON: ``stage = JSONProcessorStage()`` (for full list of
+processors check :ref:`pyDKB.dataflow.stage` section of this documentation).
+Then set stage processing function to our function ``my_process()``,
```
Evildoor: Maybe 'Then we set stage ...'?

mgolosova: Maybe this way?

> First we create stage object and indicate ...; then set ... function to ... ``my_process()``, parse .. arguments and start ... execution.

mgolosova: Just wanted to avoid another 'we' :)

Evildoor: I'm afraid that second 'we' is required because of 'First' in the front and I'm not sure whether removing 'First' will make things better. One option is to rewrite this, for example: 'The first line does A, the second line does B, ...'. Another one is to make this into a list or something.

Utils/Dataflow/pyDKB/__init__.py
mgolosova: Fixed.

Utils/Dataflow/pyDKB/__init__.py
mgolosova: Yes, fixed.

Utils/Dataflow/pyDKB/__init__.py
```diff
@@ -1,5 +1,120 @@
...
+
+Dataflow
+  ETL process (extract-transform-load) for populating internal DKB
+  storages and keeping them up to date
+
+Dataflow stage
+  Logical step of ETL process, implemented as standalone executable
+  program (worker)
+
+Dataflow stages are standalanoe programs, but can be combined into a
```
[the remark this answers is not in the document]
mgolosova: Fiwxed, thank you.

Utils/Dataflow/pyDKB/__init__.py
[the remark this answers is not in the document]
mgolosova: Fixed with the version suggested above.

Utils/Dataflow/pyDKB/__init__.py
mgolosova: Fixed, thank you.

Utils/Dataflow/pyDKB/__init__.py
mgolosova: Still do not see why the version with "First we create, then set" is not good enough.
Yet will try to rephrase :)

mgolosova:
Fixed typos and other issues mentioned above (except with respect to [this one](#discussion_r213266424)).
Sorry for multiple force update, somehow two of the commits were too eager to stick together...

mgolosova:
@Evildoor ,

>Sorry, but 'standalone' in
#158 (comment)
is still wrong.

Yeah, but in another manner now... it\`s me to be sorry for making typo during fixing. Thank you for checking!
Now fixed and double-checked...

> Two code snippets in the guide look like separate files, while logical form of the file would be:
  ```
  import sys
  import pyDKB stuff 1
  import pyDKB stuff 2
  processing function
  main
  ```

Improved this part, please check.

> 'We create then set' - I can't provide a good enough explanation for my comment, so I'm probably wrong and you may ignore it.

Ok, then I leave it be for a while.
Working with the docs we can always change it if decide that it catches eye of a readed.

mgolosova:
In the last update I have just reworded commit logs, using `docs/pyDKB` prefix for changes related to the pyDKB documentations.

PR #162, opened 2018-08-29
Stage 009: query update

mgolosova:
Improvements in SQL query of Stage 009.

Query changes:
* JSON field parsing algorithm simplified;
* removed excessive escape symbols;
* improved `--parameter[ =]"value"` pattern;
* refactoring: replace repeating field lists with equivalent constructions (at least, with equivalent result).

Other changes:
* data samples are updated;
* added test util to compare NDJSON files (without it one can hardly say what has changed in sample file and is it question of ordering or a real change);
* added timings to Stage 009.


Utils/Dataflow/009_oracleConnector/dbConnection.py
```diff
@@ -151,7 +152,10 @@ def execute_saved(self, qname, **params):
             return False
 
         try:
+            t1 = time.time()
             c.execute(None, params)
+            t2 = time.time()
+            sys.stderr.write("(TIMING) (%s) (sec): %s\n" % (qname, t2 - t1))
```
anastasiakaida: @mgolosova 

Why we should output info about execution time (if there is no error) to stderr?

anastasiakaida: @mgolosova 
we have just two ways to output something we need: stderr and stdout. If we send a message to stdout (just an "in-mind" experiment) -- and even create a pipeline --, a message will be among all sended data, so all system logs should be sended to stderr.
if I understood it right,

PR #167, opened 2018-08-31
data4es: new derived value `input_events`

mgolosova:
Add new field `input_events` to ES.

Value of this field is calculated basing on other fields:
* `processed_events` (for old EVNT tasks);
* `primary_input_events`, `requested_events` (for non-EVNT tasks);
* `n_files_per_job`, `n_events_per_job`, `n_files_to_be_used` (for more recent EVNT tasks, for which all these parameters can be queried from ProdSys database).

---
~Waits for #166.~
*As it was nearly impossible to complete this task without working Stage 091, this branch is built upon that PR and waits till it will be merged.*


Utils/Dataflow/016_task2es/task2es.py
```diff
@@ -129,6 +129,59 @@ def get_category(row):
     return categories
 
 
+def input_events(data):
+    """ Calculate derived value 'input_events'.
+
+    For EVNT tasks:
+     * n_files_to_be_used  -> nFilesToBeUsed * (nEventsPerJob|1) / nFilesPerJob
+     * !n_files_to_be_used -> processed_events
```
Evildoor: This formatting is somewhat puzzling. Is there some reason for it, for example, a documentation constructor or some other non-human interpreter demanding such format? Otherwise:

1. Why is name format different in the right part of line 136? Commit message and the code itself use `some_var_name` format, yet here it's `someVarName`.
2. What's `|1`? I know that "!x" = "not x" and with such interpretation line 137 correlates with the code and the commit message. However, if I presume "x|y" = "x or y (under some condition)", then it contradicts the code:

            try:
                events_per_job = int(data.get('n_events_per_job'))
            except TypeError:
                events_per_job = 0
 
If this is a typo and should be `|0`, then shouldn't `nFilesPerJob` be `(nFilesPerJob|1)`?

mgolosova: @Evildoor , sorry for the late reply.

> Why is name format different in the right part of line 136? Commit message and the code itself use `some_var_name` format, yet here it's `someVarName`.

`some_var_name` is an exact name of document field in ES -- the value, that is derived from some other values. 
`someVarName` etc. -- I believe I just wanted to save some space. The line is long enough, and usage of underscores would make 10 extra symbols. But if it would look better I can change it.

> What's ` |1`? I know that "!x" = "not x" and with such interpretation line 137 correlates with the code and the commit message. However, if I presume "x|y" = "x or y (under some condition)", then it contradicts the code: <...>
> If this is a typo and should be `|0`, then shouldn't `nFilesPerJob` be `(nFilesPerJob|1)`?

Right, it should be read as "x or y". Hm, there must be a typo, as you said: should be `(nFilesPerJob|1)` indeed. Will check again and fix it.

`|0` part, seen in code, was added "just in case", as we can never be sure there is no NULL in Oracle data. If this field is NULL -- then it is taken for 0; I don\`t think it should be explicitly specified in docs, it is just a (more or less) standard way to convert NULL to int.

Evildoor: > some_var_name is an exact name of document field in ES -- the value, that is derived from some other values.
> someVarName etc. -- I believe I just wanted to save some space. The line is long enough, and usage of underscores would make 10 extra symbols. But if it would look better I can change it.

I do believe that this change is necessary, because the current, non-unified version is rather puzzling.

> Right, it should be read as "x or y". Hm, there must be a typo, as you said: should be (nFilesPerJob|1) indeed. Will check again and fix it.
> |0 part, seen in code, was added "just in case", as we can never be sure there is no NULL in Oracle data. If this field is NULL -- then it is taken for 0; I don`t think it should be explicitly specified in docs, it is just a (more or less) standard way to convert NULL to int.

So it should be `(nEventsPerJob|0)` and `(nFilesPerJob|1)` like I supposed, right? If yes, then i have no further objections.

mgolosova: Fixed now.

mgolosova:
[Разрешение конфликтов в PR #167](https://trello.com/c/AZokUxhx/44-%D1%80%D0%B0%D0%B7%D1%80%D0%B5%D1%88%D0%B5%D0%BD%D0%B8%D0%B5-%D0%BA%D0%BE%D0%BD%D1%84%D0%BB%D0%B8%D0%BA%D1%82%D0%BE%D0%B2-%D0%B2-pr-167)

Utils/Dataflow/009_oracleConnector/dbConnection.py
```diff
@@ -210,3 +215,6 @@ def results(self, qname, arraysize=1000, rows_as_dict=False):
                 else:
                     yield row
             results = c.fetchmany(arraysize)
+
+        rc2 = c.rowcount
+        sys.stderr.write("(TIMING) (%s) (n_rows): %s\n" % (qname, rc2 - rc1))
```
Evildoor: Pardon my possible stupidity, but why is change in amount of rows logged as TIMING?

mgolosova: @Evildoor, good question, indeed ))
How long the query execution took ([first record logged as TIMING](https://github.com/PanDAWMS/dkb/pull/162/commits/331e6f1ef81add3099b559835760d58575537596#diff-410871e02bb7fd9f1844d212d852fbe7R158)) makes no sense by itself -- we need to know, how many records were retrieved via this query. It is obvious that querying all tasks for 1 month will take longer than that for 1 hour, right? So the actual metric of interest here is seconds per row, not seconds or rows themselves.

You are right, it was not the most accurate decision to get the metrics like this, marking them all as `(TIMING)`, but if we want to change it -- it is a matter for another PR (as this change came form `master`).

Evildoor: Fair enough.

PR #169, opened 2018-09-06
Added documentation for pyDKB.common.json_utils.

Evildoor:
Basic documentation for pyDKB.common.json_utils.


mgolosova:
After (or with) every change in docs I would add new version of HTML and PDF documentation; after all, it is the destination of any changes in docstrings ;)

Utils/Dataflow/pyDKB/common/json_utils.py
```diff
@@ -30,10 +72,16 @@ def valueByKey(json_data, key):
...
         STRING key -- dot-separated list of nested keys.
-                      If a key contains dot itself, the key must be put between
-                      quotation marks.
+                      If a key contains dot itself, the key must be put
+                      between quotation marks.
+
+    **Examples:**
+        Transform STRING ``\'1.2.3'`` into LIST ``[\'1', \'2', \'3']``.
+
+        Transform STRING ``\'1.\"2.3".4'`` into
```
mgolosova: Talking about examples — I would suggest example in form of Python code, something like:
```
>>> nestedKeys('1.2.3')
['1', '2', '3']
```
Having less words it looks more illustrative for me.

Utils/Dataflow/pyDKB/common/json_utils.py
```diff
@@ -30,10 +72,16 @@ def valueByKey(json_data, key):
 def nestedKeys(key):
     """ Transform STRING with nested keys into LIST.
 
-    Parameters:
+    **Parameters:**
         STRING key -- dot-separated list of nested keys.
-                      If a key contains dot itself, the key must be put between
-                      quotation marks.
+                      If a key contains dot itself, the key must be put
```
mgolosova: Please do not forget that originally it was written with no concern to Sphinx, reStructuredText and building documentation from this comment. Thus newline in the comment does not mean newline or new paragraph in the built version — check how it looks now. Is it what you expected to get there?

Utils/Dataflow/pyDKB/common/json_utils.py
```diff
@@ -1,14 +1,56 @@
...
+
 """
 
 
 def valueByKey(json_data, key):
     """ Return value by a chain (list) of nested keys.
 
-    Parameters:
+    It is common for JSON objects to contain many layers of dictionaries
+    nested in other dictionaries - this function extracts the data from
```
mgolosova: Single hyphen produces short dash, double — long one. In this case I guess you meant to use the long one, right?

Utils/Dataflow/pyDKB/common/json_utils.py
```diff
@@ -30,10 +72,16 @@ def valueByKey(json_data, key):
 def nestedKeys(key):
     """ Transform STRING with nested keys into LIST.
 
-    Parameters:
+    **Parameters:**
         STRING key -- dot-separated list of nested keys.
-                      If a key contains dot itself, the key must be put between
-                      quotation marks.
```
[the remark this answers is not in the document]
mgolosova: I think I am going to add a “howto” to the repo, so that we had a common guide for this work.

Utils/Dataflow/pyDKB/common/json_utils.py
Evildoor: It means a newline in built version when [source] button is clicked.
image
[docs] version is unaffected.

Utils/Dataflow/pyDKB/common/json_utils.py
Evildoor: Makes sense. However, per our discussion yesterday - tests will be used as examples, so I'll remove examples I've made here.

Utils/Dataflow/pyDKB/common/json_utils.py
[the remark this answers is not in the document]
Evildoor: Sure thing, I'll take a look at custom_readline().

Utils/Dataflow/pyDKB/common/json_utils.py
```diff
@@ -1,14 +1,56 @@
 """
-Utils to work with JSON (dict) objects.
+Utils to work with JSON objects.
+
+In context of python, JSON [#]_ objects may be considered as structures
```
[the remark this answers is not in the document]
Evildoor: According to my knowledge, JSON is not tied to python in any way - python is just another language which can work with this format. Other languages may not even have things such as 'dictionary' and use different terms to define the same object. Therefore, I believe that saying "you see, JSON file is usually just a big dictionary which may contain other dictionaries, lists, ..." is wrong without indicating that you are talking in terms of python.

Utils/Dataflow/pyDKB/common/json_utils.py
Evildoor: You are right, thanks. Will fix.

Utils/Dataflow/pyDKB/common/json_utils.py
mgolosova: Sorry, didn\`t mention that I was talking about PDF version.
HTML looks fine -- couldn\`t check it from tablet ;)

Utils/Dataflow/pyDKB/common/json_utils.py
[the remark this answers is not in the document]
mgolosova: Huh, bold symbols are not that bold when you don\`t know they are here...
I meant "shouldn\`t Python be capitalized?"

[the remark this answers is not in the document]
Evildoor: Ha, good one! Sure, will do.

Evildoor:
Updated the docs according to comments.

@mgolosova, due to software reasons known to you I cannot make PDF version at the moment. Should I make and upload HTML version or leave this issue to you?

mgolosova:
@Evildoor 
I believe I finally manage to configure TeX on the bamboo server. Please check, and if there are any problems (except that `pdflatex` claims about `fancyvrb.cfg`) please let me know.

Evildoor:
@mgolosova 
Thank you, the PDF builds now. Also made the changes according to your comments.

Utils/Dataflow/pyDKB/common/json_utils.py
```diff
@@ -57,6 +57,8 @@ def nestedKeys(key):
 
     String should contain keys separated by dot. If a key contains
     dot itself, the key must be put between matching quotation marks.
+    Quotation marks inside the keys (not preceding or following a dot)
+    are ignored.
```
mgolosova: By "ignored" you mean "treated as ordinary symbols", right?
If I read the description without knowing how the function works (or supposed to), for a moment I might think that "ignored" means "removed from the output", like:
```
>>> nested_keys('ab.c"d.e')
['ab', 'cd', 'e']
```

Utils/Dataflow/pyDKB/common/json_utils.py
```diff
@@ -28,12 +53,18 @@ def valueByKey(json_data, key):
...
-    """ Transform STRING with nested keys into LIST.
+    """ Transform string with nested keys into list.
+
+    String should contain keys separated by dot. If a key contains
+    dot itself, the key must be put between matching quotation marks.
+    Quotation marks inside the keys (not preceding or following a dot)
+    are ignored.
+
+    :param key: nested keys
+    :type key: str
```
mgolosova: > :type key: str, list

If passed `key` is of type `list`, it is returned without changes.

Utils/Dataflow/pyDKB/common/json_utils.py
```diff
@@ -1,14 +1,39 @@
...
-        DICT   json_data -- to search in
-        STRING key       -- dot-separated list of nested keys
+    It is common for JSON objects to contain many layers of dictionaries
+    nested in other dictionaries -- this function extracts the data from
+    such constructions according to given string or list with keys.
+
+    :param json_data: to search in
+    :type json_data: dict
+    :param key: nested keys
+    :type key: str
```
mgolosova: > :type key: str, list

If passed `key` is of type `list`, it is treated as "already parsed" nested keys.

Evildoor:
Done.

Utils/Dataflow/pyDKB/common/json_utils.py
```diff
@@ -1,14 +1,39 @@
...
 
 def valueByKey(json_data, key):
     """ Return value by a chain (list) of nested keys.
 
-    Parameters:
-        DICT   json_data -- to search in
-        STRING key       -- dot-separated list of nested keys
+    It is common for JSON objects to contain many layers of dictionaries
+    nested in other dictionaries -- this function extracts the data from
+    such constructions according to given string or list with keys.
```
mgolosova: The description does not explain how keys should be passed. If usage of list version can be guessed, string is not clear at all.
What about something like this?

```
...according to a list of keys. 
<...>
:param keys: sequence of nested keys (dot-separated string or list object)
```
?

mgolosova:
Everything seems to be fine; now waiting for the `pyDKB` fix...

Utils/Dataflow/pyDKB/common/json_utils.py
Evildoor: Makes sense, will do.

PR #165, opened 2018-08-31
pyDKB: merge 'master' branch

mgolosova:
Background:

* This PR is aimed in bringing documentation stuff to the `pyDKB` branch, to make it possible to revise and improve documentation directly in the latest version of the library.

* Documentation was requested by @Evildoor for pyDKB unit tests development.

* Unit tests were requested by @mgolosova to simplify further development of the library (as development without proper testing may sometimes lead to situations when things are broken and nobody knows about it for a long time; early fixes are much easier than late ones).

---

ToDo:
- [x] check conflicts resolution:
  - [x] Stage 091;
  - [x] `pyDKB.common.custom_readline`;
- [x] update pyDKB docs according to the code in destination branch;
- [x] check `--config` parameter in pyDKB and 009.

---
*~Can not be merged to `pyDKB` unless `pyDKB` itself is fixed (see PR #170).
After the merge `pyDKB` branch should be merged here again.~*
Everything that could be fixed was fixed.

PR #171, opened 2018-09-06
pyDKB: Improve stage usage/help message.

mgolosova:
Did my best to make things clear and avoid ambiguity.
With unclear help message there was no guide to check if the stage behavior is correct or not after changes.

ToDo:
 - [ ] add comment / function about flag options detection


mgolosova:
Added 'docs' label to emphasise the fact that this PR belongs to the chain of PRs that holds back PR #165, which, in its turn, holds back almost everything related to the `pyDKB` docs.

Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
```diff
@@ -115,53 +115,61 @@ def defaultArguments(self):
...
-                          nargs='?',
-                          help=u'Base directory in local file system '
-                                'or in HDFS (for relative FILE names). '
-                                'If no FILE specified, all files from the '
-                                'directory will be taken.',
-                          default='',
-                          const='',
+                          help=u'base directory in local file system '
+                          'or in HDFS (for relative FILE names). '
+                          'If no FILE specified, all files with '
```
Evildoor: Is specified? Was specified?

Evildoor:
I can't test the new code without the documentation for reworked pyDKB, so my input on this PR will be limited and probably pointless.

### 6038393
1) Is it possible for the stage author to add their own actions?
2) This change has a condition `startswith('store_')`, but the code has action without the underscore. Is this okay?
`self.add_argument('-e', '--end-of-message', action='store', type=str,`

Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
mgolosova: Both. Any.
It seems to be quite common to say “no file specified” in error messages, so it is just stuck in my mind.
It is something about “omission of ‘to be’ in passive” or something like this.
[Here](
https://forum.wordreference.com/threads/passive-voice-omission-of-be.2554621/#post-12874978) is one of the first hits of what people say about it.

mgolosova:
@Evildoor,

> Is it possible for the stage author to add their own actions?

If you mean own parameters, then the answer is yes. Please check [Stage 091](https://github.com/PanDAWMS/dkb/blob/b74ffd76bb1c31725a7fbfd6959ae9536681bdb0/Utils/Dataflow/091_datasetsRucio/datasets_processing.py#L61) -- parameter `--dataset-type|-t` is added from the stage script. 
However, currently it is not possible to override pre-defined parameters. It can be done (technically), but I am not sure it would be safe enough.

If you mean the `action` that is passed to `add_argument()`, then technically it is "yes" as well:
```
You may also specify an arbitrary action by passing an Action subclass
or other object that implements the same interface. The recommended
way to do this is to extend Action, overriding the __call__ method and
optionally the __init__ method.
```
But I never needed to use this option, so I don\`t know how it works in the real life.

> This change has a condition `startswith('store_')`, but the code has action without the underscore. Is this okay?
> `self.add_argument('-e', '--end-of-message', action='store', type=str,`

`ArgumentParser.add_argument()` has different types of [actions](https://docs.python.org/2.7/library/argparse.html#action), including:
* 'store';
* 'store_const';
* 'store_true';
* 'store_false'.

First means "store to the parameter what was passed", while others -- "store to the parameter <this_value>" (`True`, `False` or what is passed to `add_argument()` via `const`).
It is used for "flag" options, which work for the use like "turn smth on/off". And in this case it doesn\`t matter what value is stored when the option is on or off -- so the idea of `startswith('store_')` was to avoid output of `DEFAULT: <const>` in these cases (`DEFAULT: False` for `--hdfs` looked a bit ugly ;) ).

mgolosova:
@Evildoor (and whoever will review this PR),

to check the usage/help output (which is the subject of this PR) one can run any stage with `--help` argument:
```
$ 016_task2es/task2es.py -h
usage: task2es.py [-h] [-m MODE] [-c CONFIG] [-e EOM] [-E EOP] [-s {f,s,h}]
                  [-i DIR] [-d {f,s,h}] [-o DIR] [--hdfs]
                  [FILE [FILE ...]]

DKB Dataflow data processing stage.

positional arguments:
  FILE                  source data file
                        NOTE: required in (f)ile, ignored in other modes


optional arguments:
  -h, --help            show this help message and exit
  -m MODE, --mode MODE  processing mode: (f)ile, (s)tream or (m)ap-reduce.
                        Processing mode is defined as a combination of four
                        parameters: "-s SRC -d DEST -e EOM -E EOP", where:
                        ...
```

The main question of this PR is as follows: is it clear for you, after reading the help message, how things are supposed to work and what will happen if you specify this or that parameter (or combination of parameters)?
\`Cause right now things do not work as described (shame on me...). But to bring it back to work, first we need to define how should it work and what we expect from it. And this is it -- some kind of definition; the top-level documentation, from which we will start to rebuild the `pyDKB` functionality.

Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
Evildoor: I've read the forum thread, but I still believe that something should be changed here. I agree with you - short forms "No X specified", "missing X" and some others are pretty common in error messages. However, here we are talking about a complete sentence, not "something to say", as the thread defines it. There is a possibility to make it look normal, but that means changing the whole message and all other messages to similar style, i.e. "Source directory. If no FILE specified, take all files from X which match input extension".

mgolosova:
@Evildoor,
this check *is* essentially the check for flag/non-flag options. If action is `store_smth`, then no value can be passed to it and, concequently, it is a flag option.

However you are right, it is not readable unless you are aware about `argparse` module details. I better add a comment there (at least), if not encapsulate it into a separate function.

Thank you!

Evildoor:
> @Evildoor (and whoever will review this PR),
> 
> to check the usage/help output (which is the subject of this PR) one can run any stage with `--help` argument:
> 
> ```
> $ 016_task2es/task2es.py -h
> usage: task2es.py [-h] [-m MODE] [-c CONFIG] [-e EOM] [-E EOP] [-s {f,s,h}]
>                   [-i DIR] [-d {f,s,h}] [-o DIR] [--hdfs]
>                   [FILE [FILE ...]]
> 
> DKB Dataflow data processing stage.
> 
> positional arguments:
>   FILE                  source data file
>                         NOTE: required in (f)ile, ignored in other modes
> 
> 
> optional arguments:
>   -h, --help            show this help message and exit
>   -m MODE, --mode MODE  processing mode: (f)ile, (s)tream or (m)ap-reduce.
>                         Processing mode is defined as a combination of four
>                         parameters: "-s SRC -d DEST -e EOM -E EOP", where:
>                         ...
> ```

Thanks, this helps.

> The main question of this PR is as follows: is it clear for you, after reading the help message, how things are supposed to work and what will happen if you specify this or that parameter (or combination of parameters)?
> `Cause right now things do not work as described (shame on me...). But to bring it back to work, first we need to define how should it work and what we expect from it. And this is it -- some kind of definition; the top-level documentation, from which we will start to rebuild the `pyDKB` functionality.

Some things are somewhat unclear in MODE description:
- 'Processing mode is defined as a combination of four parameters: "-s SRC -d DEST -e EOM -E EOP", where:'
Somehow, word "defined" makes me ask questions such as "defined where?" and "defined by who?", while actually it is defined right here, right now. Maybe "Processing mode is a combination ...", "Using processing mode is similar to using a combination ..." or something else?
- 'Any of these parameters can be rewritten by its explicit specification.'
I believe that this part should more strictly describe the program logic, which means something like this: "If both MODE and an individual parameter are used, the individually specified value will override the MODE value"
- '--hdfs' is a third way for reaching the same goal, shouldn't it be mentioned here?

mgolosova:
@Evildoor,

> Some things are somewhat unclear in MODE description:
> 
> * 'Processing mode is defined as a combination of four parameters: "-s SRC -d DEST -e EOM -E EOP", where:'
> Somehow, word "defined" makes me ask questions such as "defined where?" and "defined by who?", while actually it is defined right here, right now. Maybe "Processing mode is a combination ...", "Using processing mode is similar to using a combination ..." or something else?

Changed.

> * 'Any of these parameters can be rewritten by its explicit specification.'
> I believe that this part should more strictly describe the program logic, which means something like this: "If both MODE and an individual parameter are used, the individually specified value will override the MODE value"

Both sound good enough for me, so changed to your version.

> '--hdfs' is a third way for reaching the same goal, shouldn't it be mentioned here?

That\`s not true. `--hdfs` is an equivalent to `--source h --dest h`, specifying that base file system to work with (for any `MODE`) is not the local one, but the HDFS. It has nothing to do with `--end-of-message` and `--end-of-process`.

`--mode` is meant to be "processing mode" -- and all three available modes are used in specific situations:
* (f)ile -- debuggung, testing, sample generation, etc.;
* (s)tream -- in the chain of processors;
* (m)ap-reduce -- to massively process a lot of data (e.g., all the PDF documents).

`--hdfs` was originally introduced to simplify the (m)ap-reduce mode implementation (as data for MR processing are usually stored in HDFS and can not be taken as "local" files), but can be used for file processing as well.

Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
mgolosova: Reworded.

mgolosova:
@Evildoor said:
> Ok, then it may be useful to implement adding or not adding DEFAULT by checking whether or not the option is a flag one instead of its action starting with "store_".

@mgolosova said:
> you are right, it is not readable unless you are aware about `argparse` module details. I better add a comment there (at least), if not encapsulate it into a separate function.

Moved to a separate method.

mgolosova:
Seems like I\`ve addressed all the comments; please review.

anastasiakaida:
@mgolosova

I'll review it tomorrow.

Evildoor:
> > '--hdfs' is a third way for reaching the same goal, shouldn't it be mentioned here?
> 
> That`s not true. `--hdfs` is an equivalent to `--source h --dest h`, specifying that base file system to work with (for any `MODE`) is not the local one, but the HDFS. It has nothing to do with `--end-of-message` and `--end-of-process`.

True, but it has with "--source" and "--dest", right? If I understand the logic correctly, then "--source" and "dest" can be defined in any way out of three, and the priority is hdfs > individual > mode. The second ">" is mentioned in mode description, while hdfs description states that it overrides both other ones. Is it okay that one has to deduce this priority by reading both mode and hdfs descriptions? (I'm not insisting, it probably is.)

mgolosova:
@Evildoor ,
I believe that if one decides to use `--hdfs`, he/she will read its description.
While if `--hdfs` is not used, then it does not matter if it is "stronger" than individual `--source` and `--dest` or not.

PR #174, opened 2018-09-10
Fixed quote-processing mechanism in nestedKeys().

Evildoor:
- Incrementing i1 advances the cycle (it is endless with i).
- Dot is removed when split() is made, so it is added back into
quoted keys.

### DIFF
--- a/Utils/Dataflow/pyDKB/common/json_utils.py
+++ b/Utils/Dataflow/pyDKB/common/json_utils.py
@@ -37,24 +37,36 @@ def nestedKeys(key):
     """
     if type(key) == list:
         return key
-    nested_keys = []
 
-    splitted_key = key.split('.')
-    skip = []
-    for i in range(len(splitted_key)):
-        if i in skip:
-            continue
-        k = splitted_key[i]
-        if k[0] in ("'", '"'):
-            i1 = i
-            while k[-1] not in ("'", '"'):
-                i1 += 1
-                if i1 >= len(splitted_key):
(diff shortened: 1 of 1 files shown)


mgolosova:
First I want to say thank you for bringing up this issue.
However I think the function still does not work as it was supposed to (sorry for the poor realization :( ).

1. It is better to remove quote marks from the keys. Now it is:
  ```
  >>> nestedKeys('a."b.c"')
  ['a', '"b.c"']
  ```
  and should be:
  ```
  >>> nestedKeys('a."b.c"')
  ['a', 'b.c']
  ```

2. We should detect unpaired quote marks as a syntax error, not as a correct pair of quotes. Now it is:
  ```
  >>> nestedKeys('a."b.c\'.d')
  ['a', '"b.c\'', 'd']
  ```
  and should be something like:
  ```
  >>> nestedKeys('a."b.c\'.d')
    File "pyDKB/common/json_utils.py", line ???, in nestedKeys
      "configuration key: %s" % key)
    ValueError: Failed to decode dot-splitted configuration key: a."b.c'.d
  ```

And following this logic, another example should work this way:
```
>>> nestedKeys('a.\'b."c".d\'.e')
['a', 'b."c".d', 'e']
```

Evildoor:
Makes sense. I can take a look, but I'll probably have to rewrite the function in such case - is that ok?

mgolosova:
> I'll probably have to rewrite the function in such case - is that ok?

Yes, sometimes it is the only way to make things right.

Evildoor:
Rework is done.

PR #178, opened 2018-09-20
pyDKB: common/hdfs.py documentation

anastasiakaida:
Changed according to the template.

Note: there are still some empty :return: and :rtype: lines when 
a function don't return anything explicitly.

### DIFF
--- a/Utils/Dataflow/pyDKB/common/hdfs.py
+++ b/Utils/Dataflow/pyDKB/common/hdfs.py
@@ -23,14 +23,14 @@ def check_stderr(proc, timeout=None, max_lines=1):
     if MAX_LINES == None, output all the STDERR.
 
     Return value is the subprocess` return code.
-    
+
     :param proc: subrocess
     :type proc: subprocess
     :param timeout: time limit for operation
     :type timeout: number (int)
     :param max_lines: maximum quantity of lines
     :type max_lines: number (int)
-    
+
     :return: subprocess return code, set by poll()
     :rtype: int
     """
@@ -52,12 +52,12 @@ def check_stderr(proc, timeout=None, max_lines=1):
 
 def makedirs(dirname):
     """ Try to create directory (with parents).
-    
+
     :param dirname: a name of a created directory
     :type dirname: string
-    
-    :return: 
-    :rtype: 
+
+    :return:
+    :rtype:
     """
     cmd = ["hadoop", "fs", "-mkdir", "-p", dirname]
     try:
@@ -75,12 +75,12 @@ def makedirs(dirname):
 
 def putfile(fname, dest):
     """ Upload file to HDFS.
-    
+
     :param fname: file name
     :type fname: string
     :param dest: destination for uploaded file
     :type dest: string
-    
+
     :return:
     :rtype:
     """
@@ -100,12 +100,12 @@ def putfile(fname, dest):
 
 def movefile(fname, dest):
     """ Move local file to HDFS.
-    
+
     :param fname: file name
     :type fname: string
     :param dest: destination for moved file
     :type dest: string
-    
+
     :return:
     :rtype:
     """
@@ -120,10 +120,10 @@ def movefile(fname, dest):
 
 def getfile(fname):
     """ Download file from HDFS.
-    
+
     :param fname: file name of a downloaded file
     :type fname: string
-    
+
     :return: file name without its directory
     :rtype: string
     """
@@ -145,10 +145,10 @@ def getfile(fname):
 
 def File(fname):
     """ Get and open temporary local copy of HDFS file
-    
+
     :param fname: name of a temporary file
     :type fname: string
-    
+
     :return: open file object (TemporaryFile)
     :rtype: file
     """
@@ -181,12 +181,12 @@ def listdir(dirname, mode='a'):
         mode    -- 'a': list all objects
                    'f': list files
                    'd': list subdirectories
-    
+
     :param dirname: name of HDFS directory
     :type dirname: string
     :param mode: type of a list of files
     :type mode: string
-    
+
     :return: a list of files' names/subdirectories inside HDFS
     :rtype: list
     """
@@ -247,10 +247,10 @@ def listdir(dirname, mode='a'):
 
 def basename(path):
     """ Return file name without path.
-    
+
     :param path: path to a file
     :type path: string
-    
+
     :return: file name without path
     :rtype: string
     """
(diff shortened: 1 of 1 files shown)


mgolosova:
Not reviewed yet as #165 should go first.

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -1,5 +1,6 @@
 """
 Utils to interact with HDFS.
+
```
mgolosova: Why a new line?

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -22,6 +23,16 @@ def check_stderr(proc, timeout=None, max_lines=1):
...
     Return value is the subprocess` return code.
+    
+    :param proc: subrocess
+    :type proc: subprocess
+    :param timeout: time limit for operation
+    :type timeout: number (int)
+    :param max_lines: maximum quantity of lines
+    :type max_lines: number (int)
+    
+    :return: subprocess return code, set by poll()
```
mgolosova: The return code is not *set* by `poll()` -- it is returned by the subprocess, and can be *accessed* by `poll()`.
I believe it will be Ok to simply say that it is a "subprocess return code".

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -84,8 +120,12 @@ def movefile(fname, dest):
 
 def getfile(fname):
     """ Download file from HDFS.
-
-    Return value: file name (without directory)
+    
+    :param fname: file name of a downloaded file
```
mgolosova: I\`d rather say "path to file", then "file name".

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -84,8 +120,12 @@ def movefile(fname, dest):
...
+    :type fname: string
+    
+    :return: file name without its directory
```
mgolosova: "Without path"?

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -105,8 +145,12 @@ def getfile(fname):
 
 def File(fname):
     """ Get and open temporary local copy of HDFS file
-
-    Return value: open file object (TemporaryFile).
+    
+    :param fname: name of a temporary file
```
mgolosova: Ahem.. please check again ;)

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -105,8 +145,12 @@ def getfile(fname):
...
+    :type fname: string
+    
+    :return: open file object (TemporaryFile)
```
mgolosova: ...or None in case of error.

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -137,6 +181,14 @@ def listdir(dirname, mode='a'):
         mode    -- 'a': list all objects
                    'f': list files
                    'd': list subdirectories
+    
+    :param dirname: name of HDFS directory
+    :type dirname: string
+    :param mode: type of a list of files
+    :type mode: string
+    
+    :return: a list of files' names/subdirectories inside HDFS
```
mgolosova: ...in HDFS directory.
"Inside HDFS" looks a bit too broad and can make one to believe it is about all files/subdirectories.

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -194,21 +246,44 @@ def listdir(dirname, mode='a'):
 
 
 def basename(path):
-    """ Return file name without path. """
+    """ Return file name without path.
+    
+    :param path: path to a file
+    :type path: string
```
mgolosova: If `path` is `None`, it is taken as empty string -- I believe it should be mentioned somewhere...

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -194,21 +246,44 @@ def listdir(dirname, mode='a'):
...
         path = ''
     return path.basename(path).strip()
 
 
 def dirname(path):
-    """ Return dirname without filename. """
+    """ Return dirname without filename.
+    
+    :param path: path to a file
+    :type path: string
```
mgolosova: Same as above about `None`.

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -194,21 +246,44 @@ def listdir(dirname, mode='a'):
...
 
 
 def join(path, filename):
-    """ Join path and filename. """
+    """ Join path and filename.
+    
+    :param path: path to a file
+    :type path: string
+    :param filename: file name
+    :type filename: string
```
mgolosova: Same as above about `None`.

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -40,7 +49,17 @@ def check_stderr(proc, timeout=None, max_lines=1):
 
 
 def makedirs(dirname):
-    """ Try to create directory (with parents). """
+    """ Try to create directory (with parents).
+
+    Create directory and check status of process
+    with subprocess.Popen.
+
+    :param dirname: a name of a created directory
```
Evildoor: Maybe "name of the directory" or something else?
1) "a" before "name" - similar instances of "name" in this file come without "a".
2) "a" before "directory" - it's not some directory, it's a specific directory which was already mentioned in lines 52 and 54. By the way, in those lines it can be "create a directory".
3) "created" - we are talking about arguments, so nothing was created yet. "to create" or "to be created" can be used if necessary.

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -85,7 +128,14 @@ def movefile(fname, dest):
 def getfile(fname):
     """ Download file from HDFS.
 
-    Return value: file name (without directory)
+    Get file from HDFS and check status of process with
+    subprocess.Popen.
+
+    :param fname: path of a downloaded file
```
Evildoor: "Path to the file (to download/to be downloaded)". Apart from the past tense in function arguments (see comment to line 57, point 3), it is possible to read this as "path to the local copy of the downloaded file" - which is incorrect, if I understand correctly.

PR #179, opened 2018-09-20
[absolete] DF/10, 69: Custom markers

anastasiakaida:
Custom markers for sink/source connectors.

EOProcess: -E|--eop
EOMessage: -e|--eom (in 010 Stage)
Batch mode: -b|--batch [(e)enabled|(d)isabled], Default: (d)isabled
EOBatch: -B|--eob (in 069 Stage)

'\x17' - End Of Transmission Block (ASCII)

060 will be performed separately

### DIFF
--- a/Utils/Dataflow/069_upload2es/load_data.sh
+++ b/Utils/Dataflow/069_upload2es/load_data.sh
@@ -10,6 +10,17 @@ $(basename "$0") [-c CONFIG] [FILE...]
 PARAMETERS:
   CONFIG -- configuration file
   FILE   -- file in NDJSON format for loading to Elasticsearch via bulk interface
+
+OPTIONS:
+  -b, --batch {e[abled]|d[isabled]} Specifies batch-mode: (e)nabled|(d)isabled.
+  -B, --eob <EOB>                   Specifies the delimiter between sets of input
+                                    data in the stream mode.
+                                    Default: \x17
+                                    No batch-mode: \n
+  -E, --eop <EOP>                   Specifies end-of-process marker.
+                                    Default:
+                                    * in a (f)ile mode: none
+                                    * in a (s)tream mode: \0
 "
 }
 
--- a/Utils/Dataflow/069_upload2es/load_data.sh
+++ b/Utils/Dataflow/069_upload2es/load_data.sh
@@ -44,8 +44,8 @@ ES_PORT='9200'
 
 CURL_N_MAX=10
 SLEEP=5
-DELIMITER=`echo -e -n "\x00"`
-EOProcess=`echo -e -n "\x06"`
+DELIMITER='\x00'
+EOProcess='\x06'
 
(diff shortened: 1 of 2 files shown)


Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -57,23 +99,34 @@ load_files () {
   do
     ${cmd}${INPUTFILE} || exit 3
   done
+  echo -ne "$EOProcess"
 }
 
 load_stream () {
   log "Switched to the stream mode."
-  while read -r -d "$DELIMITER" line; do
+  while read -r -d $(echo -ne "$EOBatch") line; do
```
mgolosova: Still not working with "space"-like characters, including the newline character...
When we use `$(command)`, the space-like characters (newline, zero byte, etc) are stripped form the `command` output. We discussed this fact many times but I never noticed that here it is the same case...

I apologize for the mislead.

Seems like there\`s no other way but passing the already interpreted EOBatch symbol to the script, so that the stage is run like this:
```
  ./load_data.sh -E $'\n' -B $'\0'
```

Here is a modelled situation to check that it works:
```
$ f() {
>   EOBatch="$1"
>   while read -r -d "$EOBatch" line; do
>     echo "line: '$line'"
>   done
> }
$ echo -ne 'First line\x00Second line\x00' | f $'\x00'
line: 'First line'
line: 'Second line'
$
$ echo 'First line
Second line' | f $'\n'
line: 'First line'
line: 'Second line'
```

There may be a way to achieve what we need by `read`ing interpreted value from `echo`, but for now I do not see how to do it properly (simple pipe won\`t work). And again -- we can not write `\x00` into a variable.

---

Passing of "encoded" strings (like `'\n'` instead of `$'\n'` or `'

'`) is an appropriate option for markers used only in the output (like EOP); but for those used to interpret input, we want to pass them to some Shell command as parameters, and thus need to deal with the interpreted value... which is not a proper task for Shell, actually.

Utils/Dataflow/069_upload2es/load_data.sh
```diff
@@ -57,23 +99,34 @@ load_files () {
...
-    echo -n "$EOProcess"
+    echo -ne "$EOProcess"
   done
 }
 
 if [ -z "$1" ] ; then
+  if [ -z "$EOP" ] ; then
+    EOProcess="\0"
+  else
+    EOProcess="$EOP"
```
mgolosova: I\`m afraid that here will also appear that issue with empty string passed as EOP (`-E ''`).

anastasiakaida:
To prevent conflicts between branches I decided to close this PR and start the new one from scratch, trying to pay attention to all comments, issues and questions.

anastasiakaida:
To prevent conflicts between branches I decided to close this PR and start the new one from scratch, trying to pay attention to all comments, issues and questions.

PR #180, opened 2018-09-24
[WIP] Sample generator for 'data4es' stages.

mgolosova:
For all stages from 'data4es' process was added script `update_sample`, which updates output data sample for the given stage.

To avoid repeating same piece of code in every `update_sample`, it was moved into a separate function, placed in `Dataflow/shell_lib/update_sample`. `Dataflow/shell_lib` is meant to be a place for all common shell functions.

Another update script -- for the whole dataflow process (`run/data4es-update-samples`) -- was added to update samples for all stages in a proper order.

Samples were updated by means of the added script to have possibility to check that everything works as it should.


Utils/Dataflow/095_datasetInfoAMI/update_sample
```diff
@@ -4,6 +4,7 @@ base_dir=$(readlink -f $(dirname "$0"))
 
 sample=$base_dir/output/datasets.ndjson
 
+git checkout $sample
 mv $sample{,.bak}
```
anastasiakaida: @mgolosova 
As I remember, "git checkout" allows to switch to a remote existed branch. Can't figure out how it will work in that case.

anastasiakaida: Those strings were removed in d0af9f0, but I would like to ask anyway what was the main idea of c0b628e.

mgolosova: There are a lot of ways to use `git checkout`: https://git-scm.com/docs/git-checkout.
[One of them](https://git-scm.com/docs/git-checkout#git-checkout-emgitcheckoutemlttree-ishgt--ltpathspecgt82308203) allows to update files in the working tree:
```
git checkout [<tree-ish>] [--] <pathspec>…​
    Overwrite paths in the working tree by replacing with the contents
    in the index or in the <tree-ish> (most often a commit). When a <tree-ish>
    is given, the paths that match the <pathspec> are updated both in the index
    and in the working tree.
```

Utils/Dataflow/run/data4es-update-samples
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env bash
+
+# Utility to update data samples for dataflow stages
+#
+# data4es process overview:
+#
+# 009 --> 091(in) --> 025 --> 016 --> 019 --> 069
+#   \                                         /
+#    \----> 091(out) ----> 093 ----> 095 ----/
+
```
anastasiakaida: @mgolosova 
Pipeline 095 | 069 - looks a little bit strange, because -- according to our discussion -- 069 doesn't provide the data processing from 095. Could it be fixed?

mgolosova: Taking into account last changes (including, but not limited to the separation of Stage 16 into two stages, 16 and 17), I think I should mark this PR as `[WIP]`: its current version is not exactly corresponding the dataflow version found in master.

mgolosova:
Closing this one for `run/data4es-start --debug` with proper timestamp values do the trick; all that one needs to do is to move generated files with each stage's output to a proper location.

PR #181, opened 2018-09-24
DF: fix dataflow description in README (add stage 025).

mgolosova:


Evildoor:
Now there are two ins and two outs for *. Wouldn't it be better to use *, ** and *** instead of *, * and **?

mgolosova:
@Evildoor , good point. Fixed now.

PR #182, opened 2018-09-24
DF/data4es: processed events calculation improved.

mgolosova:
Value for `processed_events` used to be calculated as sum of `neventsused` field for all input datasets of a given task.
However for some tasks (e.g. tasks of EVNT step) there are no input datasets (as it is data generation step), so this value remained NULL.

To improve this situation, in such cases `processed_events` value is now taken from `total_events` field of ProdSys DB table `T_PRODUCTION_TASK`.

Before the change was made, data samples were actualized so that only the changes related to the last change were in the diff.


mgolosova:
[processed_events](https://trello.com/c/4sJH0EYY/23-processedevents)

PR #183, opened 2018-10-14
Stage 054 update

anastasiakaida:
Some changes were done according to changes in stage 30.

Note: input samples were taken directly from stage 30 output folder, but without "dkbID" field it wouldn't work.
For tests that field was added artificially (not added in that request).


mgolosova:
I think that old input data samples can simply be replaced with new ones: there seem to be no use of the old ones any more. For the 25th you even know the proper dkbID and can add it to the sample manually; for the 30th it can be a fake one, like "FAKE-ID-30".

Then the output data samples should be updated accordingly: data samples should be  kept up-to-date with the code that produces them, if possible.

anastasiakaida:
@mgolosova 
Done!
Please, continue your review. :)

mgolosova:
@anastasiakaida, thank you.

Now I would ask you to do the following.

1. Join commits:
* updated input files (cc4d462)
* delete unnecessary input samples (0797f88)
* add updated input data samples (e9db4b8)
* delete outstanding output samples (7ff4333)
2. Rewrite commit logs:
* in commit "stage changes "(c7a25a3):
    * most of the commit logs are started with verb; it makes reader to stumble over "stage", which can also be a verb meaning "put on", "present", "produce" etc.
* in all the commit logs:
    * please, follow the "imperative-mood" style [1]. It is not mandatory to follow this specific style (could be any other); but better to follow at least some. This one is the one I prefer to follow, and I see that most of the time you write logs this way too; so I believe it would not be difficult for you to follow it consciously and in all commits (not just the most of them).

Why to join commits (item 1 from the above):
1. It is not obvious, why e9db4b8 is needed when there already was cc4d462 (at least now and by log messages; I might knew it a month ago, now I can not remember).
2. When you add updated samples, there is no reason to keep obsolete ones in the repository.
3. When you remove input files the output ones are produced from, these output files become irrelevant.

The main idea is that after every commit the repository is left in a consistent state. When you have old and new samples in same directory, or when output files can not be produced from input samples -- it feels like an inconsistency. It is not good, to have no output sample at all -- but it is better then have irrelevant one; for it is an intermediate commit in a PR, it is excusable. As originally the input samples were a bit out of date, missing some changes from Stage 030, I believe it can be done in a sequential commits.
The "perfect" way, as I see it and try to follow whenever I can, would be to first make samples up to date (to the moment before `data_taking_year` became `data_taking_years`), both input and output; then perform changes related to the `data_taking_year(s)`: update input samples, code and output samples. But I understand that it requires some extra work to repair samples consistency (I hate it, personally) and thus do not insist :)

[1] https://git.kernel.org/pub/scm/git/git.git/tree/Documentation/SubmittingPatches?id=HEAD#n135
```
[[imperative-mood]]
Describe your changes in imperative mood, e.g. "make xyzzy do frotz"
instead of "[This patch] makes xyzzy do frotz" or "[I] changed xyzzy
to do frotz", as if you are giving orders to the codebase to change
its behavior.  Try to make sure your explanation can be understood
without external resources. Instead of giving a URL to a mailing list
archive, summarize the relevant points of the discussion.
```

anastasiakaida:
@mgolosova 
The required changes were made. Please, review. :)

mgolosova:
Everything looks fine now, thank you.
Merged, at last ;)

mgolosova:
[PR 177: stage 054](https://trello.com/c/3zSpSKBw/31-pr-177-stage-054)

PR #185, opened 2018-11-01
ES hashtag steps -- update

anastasiakaida:
New fields added to summary:

cpu (based on 'hs06')
walltime (end-start timestamp)

### DIFF
--- a/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
+++ b/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
@@ -37,7 +37,7 @@ GET <index>/task/_search
           "sum": {"script": {"inline":"doc['hs06'].value * doc['total_events'].value"}}
         },
         "walltime": {
-         "sum": {"script": {"inline":"doc['end_time'].value - doc['start_time'].value"}}
+         "avg": {"script": {"inline":"doc['end_time'].value - doc['start_time'].value"}}
         },
         "output": {
           "children": {"type": "output_dataset"},
--- a/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
+++ b/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
@@ -38,6 +38,9 @@ GET <index>/task/_search
         "cpu": {
           "sum": {"field": "hs06"}
         },
+        "walltime": {
+         "sum": {"script": {"inline":"doc['end_time'].value - doc['start_time'].value"}}
+        },
         "output": {
           "children": {"type": "output_dataset"},
           "aggs": {
--- a/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
+++ b/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
@@ -33,6 +33,9 @@ GET <index>/task/_search
         "processed_events": {
           "sum": {"field": "processed_events"}
         },
+        "cpu": {
+          "sum": {"field": "hs06"}
+        },
         "output": {
           "children": {"type": "output_dataset"},
           "aggs": {
--- a/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
+++ b/Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
@@ -11,6 +11,8 @@ GET <index>/task/_search
     "bool": {
       "must": [
         {"term": {"hashtag_list": %%HASHTAG_LOWERCASE%%}},
+        {"exists": {"field": "start_time"}},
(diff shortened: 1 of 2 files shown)


Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
```diff
@@ -11,6 +11,8 @@ GET <index>/task/_search
     "bool": {
       "must": [
         {"term": {"hashtag_list": %%HASHTAG_LOWERCASE%%}},
+        {"exists": {"field": "start_time"}},
+        {"exists": {"field": "end_time"}},
```
mgolosova: This condition will affect all the aggregations. Why should not we add value of "input_events" of a task with undefined start/end time to the total sum of input events number (for example)?

mgolosova:
@anastasiakaida ,
please take a look, data filtering can be performed within aggregations:
```
             "not_removed": {
               "filter": {"term": {"deleted": false}},
               "aggs": {
                 "bytes": {
                   "sum": {"field": "bytes"}
                 }
               }
             }
```

Here we filter out all the documents with field "deleted" set to `true` -- and sum field "bytes" of all the other documents (that passed the filter).

anastasiakaida:
@mgolosova 

Fixed! Hope, it will work better.

Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
```diff
@@ -38,9 +36,21 @@ GET <index>/task/_search
         "cpu": {
           "sum": {"field": "hs06"}
         },
-        "walltime": {
-         "sum": {"script": {"inline":"doc['end_time'].value - doc['start_time'].value"}}
-        },
+		"timestamp_defined": {
+			"filter": {
+             "bool": {
```
mgolosova: Just a comment: I believe it would also work without `"bool": {`, simply with its content: `"must": [...]`.

Not just a comment: indentation is broken. For such multi-level JSON it is very important for readability to keep it aligned properly, or else it is too easy to lose the context.

Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
```diff
@@ -38,9 +36,21 @@ GET <index>/task/_search
...
+               "must": [
+                 {"exists": {"field": "start_time"}},
+                 {"exists": {"field": "end_time"}}
+               ]
+             }
+           },
+          "aggs": {
+            "input_bytes": {
```
mgolosova: Should be `"walltime"`, I believe.

Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
```diff
@@ -38,9 +36,21 @@ GET <index>/task/_search
...
+              "sum": {"script": {"inline":"doc['end_time'].value - doc['start_time'].value"}}
```
mgolosova: And the last: please, add missed space after colon: `"inline":"doc['end_time']...`

Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
anastasiakaida: @mgolosova Sure. Thank you!

Utils/Elasticsearch/requests/hashtag-steps-statistics-q.json
anastasiakaida: @mgolosova 
I tried to check and, unfortunately, it doesn't work. :(
It seems that 'bool' is a mandatory part of this construction.

anastasiakaida:
@mgolosova required changes are done

mgolosova:
@anastasiakaida, thank you!
Now the query looks fine, yet response sample still looks questionable (number of documents matching same conditions from ~3000 became ~130000).

anastasiakaida:
@mgolosova 

Fixed and updated. :)

mgolosova:
@anastasiakaida, yeah, thank you!
One last thing left to do: reorder steps in the output JSON document, so that “Evgen Merge” was right after “Evgen” — it will make the diff more compact.

anastasiakaida:
@mgolosova Done!

mgolosova:
@anastasiakaida, perfect, thank you! Will include the changes into the original branch later today.

mgolosova:
@anastasiakaida, changes are cherry-picked into #156 (with a bit more detailed log messages). Please take a look, if everything is fine and nothing is missed -- I believe we can close this PR.

anastasiakaida:
Now all required works are done.
All commits were cherry-picked, so there is no need to keep this PR open.

PR #190, opened 2019-01-15
DF/025: standardize docstrings (params and return value descriptions).

mgolosova:


Utils/Dataflow/025_chicagoES/stage.py
```diff
@@ -59,7 +59,8 @@ def task_metadata(taskid, fields=[], retry=3):
 
     :param taskid: Task ID or None
     :type taskd: str, NoneType
-    :param fields: requested ES fields
+    :param fields: requested ES fields; if empty list or not passed, all
+                   fields available will be used
```
anastasiakaida: @mgolosova 
It seems that a construction "all available fields will be..." is more suitable.

mgolosova: @anastasiakaida, the difference (as I see it) is as between "поля, которые доступны" and "доступные поля". I meant the first, like a short form of "fields that are available".
Does it make sense? :)

Yet maybe it should be "all *the* fields"...

anastasiakaida: > @anastasiakaida, the difference (as I see it) is as between "поля, которые доступны" and "доступные поля". I meant the first, like a short form of "fields that are available".
> Does it make sense? :)
> 
> Yet maybe it should be "all _the_ fields"...

@mgolosova 
I supposed that when the meaning is the same after Russian-English translation (yeah, I think it's the same) it's better to use a "full" form without "that are..." construction, just my point of view. The final decision is up to you. :)

Just let me know if you think that it's still the best variant to leave as it is. I can approve it too.

mgolosova: @anastasiakaida, I left the `fields available` variant, yet slightly reworded the description. You made me re-read what I had wrote and, yeah, it wasn\`t a Good Writing :) Thank you!

Please proceed with the review.

mgolosova:
@anastasiakaida, it appeared that I made another typo in the docstrings; now it is fixed, but your approving was dismissed :(

Could you please re-approve?

mgolosova:
@anastasiakaida, thank you!

PR #191, opened 2019-01-18
Stage 025 update (total HS06)

mgolosova:
The HS06 we did take from Chicago ES before is actually an estimated
value for HS06*1sec per event.
But for already finished tasks it is possible to get the real value
(by aggregating task jobs metadata).

Additional metedata after this stage:
* totHS06_failed   -- CPU usage of all failed jobs;
* totHS06_finished -- CPU usage of all failed jobs;
* totHS06          -- 'finished' + 'failed'.

We could make it only 'totHS06' without distribution by status, but I
believe that both "what was really spent on this task" and "what would
be spent on this task if everything went perfect" can be useful metrics.

In Chicago ES there are individual index 'jobs_archive_YYYY-MM-DD' for
every day and it is quite a lot. To avoid hittin them all, we use task
start/end time to select indices we might need (since all the jobs
should be started and finished within this interval).

ToDo:
- [x] fix codestyle
- [x] add task status checks
- [x] update out data sample
- [x] update ES scheme
- [x] check in `data4es`
- [x] introduce to the real `data4es` process

---
~~Waits for #190~~ *(merged)*


mgolosova:
Not sure if I should leave artificial record in the input data:
```
{"taskid": 13319225, "start_time" : "27-02-2018 16:56:38", "end_time" : "07-03-2018 14:24:46", "status": "done"}
```

It helps to see what\`s going on, yet it has nothing to do with the general pattern of samples.

There\`s an idea of samples update (check [here](https://trello.com/c/zSy22PD4/57-samples-update-add-2018-data)), but this PR has nothing to do with it. So I see two options:
* it can be left "as is" in the perspective of the total samples update;
* the extra record can be removed in the last commit, like "changes are checked and the result is demonstrated, but to keep samples consistent this must not go to `master`".

mgolosova:
[Точное значение cputime](https://trello.com/c/CCuUcKKJ/48-%D1%82%D0%BE%D1%87%D0%BD%D0%BE%D0%B5-%D0%B7%D0%BD%D0%B0%D1%87%D0%B5%D0%BD%D0%B8%D0%B5-cputime)

anastasiakaida:
@mgolosova 
In 492b583 there is a typo `instead of the <> dict`, so now the description look a little bit odd. Please, correct.

mgolosova:
@anastasiakaida, fixed, thank you!

mgolosova:
[Stage 025: fix TooLongFrame](https://trello.com/c/dZKBDCx2/67-stage-025-fix-toolongframe)

PR #196, opened 2019-01-25
[frozen] pyDKB.dataflow.stage tests

Evildoor:
Tests for AbstractStage and ProcessorStage.

So far the tests are checking that the command line arguments are processed correctly.

### DIFF
--- a/Utils/Dataflow/pyDKB/dataflow/stage/tests/test_ProcessorStage.py
+++ b/Utils/Dataflow/pyDKB/dataflow/stage/tests/test_ProcessorStage.py
@@ -1,18 +1,16 @@
 #!/usr/bin/env python
 
 """
-Tests for pyDKB.dataflow.stage.ProcessorStage.
+Tests for pyDKB.dataflow.stage.ProcessorStage.configure().
 Usage: 'python -m unittest discover' from ..
 (directory with pyDKB.dataflow.stage code).
 """
 
-
-import cStringIO
 import os
 import sys
-import tempfile
 import unittest
 
+from lib import isolate_function_error
 
 # Relative import inside of pyDKB prevents the use of simple 'import pyDKB'.
 try:
@@ -26,36 +24,6 @@ try:
 except Exception, err:
     sys.stderr.write("(ERROR) Failed to import pyDKB library: %s\n" % err)
     sys.exit(1)
-
-
-def isolate_function_error(f, *args):
-    """ Silence and retrieve the function's error message.
-
-    The function is expected to throw a SystemExit when run with
-    specific arguments. Error stream is redirected into a string during the
-    function's execution, and the resulting messages can be analyzed.
-
-    :param f: function to execute
-    :type f: function
-    :param args: arguments to execute function with
-    :type args: list
-
-    :return: list with two members, first one is the error message,
-             second one is the function's return
-    :rtype: list
-    """
-    buf = cStringIO.StringIO()
-    temp_err = sys.stderr
-    sys.stderr = buf
-    try:
-        result = f(*args)
-    except SystemExit:
-        result = None
-    sys.stderr = temp_err
-    buf.seek(0)
-    msg = buf.read()
-    buf.close()
-    return [msg, result]
 
 
 args_to_add = {
@@ -78,7 +46,7 @@ modes = {
 }
 
 
-class ProcessorStageArgsTestCase(unittest.TestCase):
+class Case(unittest.TestCase):
     expected_args = {
         'mode': 'f',
         'config': None,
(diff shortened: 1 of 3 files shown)


Utils/Dataflow/pyDKB/dataflow/stage/tests/test_ProcessorStage.py
```diff
@@ -170,6 +170,24 @@ def f(self):
         setattr(ProcessorStageArgsTestCase, 'test_%s_%s' % (arg, val), f)
 
 
+def add_arg_incorrect(arg, short=False):
+    if short:
+        val = 'i'
```
mgolosova: Maybe it would be better to use here `'incorrect`' as value instead of `'i'`? 
I have mentioned it earlier, but the [comment](https://github.com/PanDAWMS/dkb/commit/a5e428b8f4ffcc4e28964f48a29b5e31ecdb9cbf#r32009968) was buried after `push --force` :) -- values for the parameters being tested are more likely to be one-char length, so using the longer word we are almost guaranteed from hitting a new valid value by chance.

Utils/Dataflow/pyDKB/dataflow/stage/tests/test_ProcessorStage.py
```diff
@@ -55,81 +55,74 @@ class ProcessorStageArgsTestCase(unittest.TestCase):
 
     def setUp(self):
         self.stage = pyDKB.dataflow.stage.ProcessorStage()
+        self.args = dict(self.default_args)
```
mgolosova: Maybe name it `expected_args` (or `exp_args`)? Longer, right, but more readable... it\`s just now it makes me remid myself every now and then how it is used.

mgolosova:
[test: pyDKB/dataflow/stage/ ProcessorStage.py parse_args](https://trello.com/c/9jiVDw2S/50-test-pydkb-dataflow-stage-processorstagepy-parseargs)

Utils/Dataflow/pyDKB/dataflow/stage/tests/test_ProcessorStage.py
Evildoor: Done.

Utils/Dataflow/pyDKB/dataflow/stage/tests/test_ProcessorStage.py
Evildoor: Done.

Evildoor:
> And maybe there should be two tests (with raw and interpreted strings passed); I don`t know what can go wrong, but it doesn`t mean it can`t :)

Done.

> This combination:

>     args['hdfs'] = True
>     args['source'] = 'h'
>     args['dest'] = 'h'

> appears here and there; maybe a shourtcut can be used (as with mode-dependent parameters)? Kind of:

>     hdfs_args = {'hdfs': True, 'source': 'h', 'dest': 'h'}
>     <...>
>     self.stage.parse_args(['--hdfs'])
>     self.args.update(hdfs_args)

Done.

Evildoor:
> So, unless there are objections or comments, then I'll split the file into two, simplify the case-processing mechanism, set a fixed name for a test case, and change the "how-to:tests" accordingly.

Done.

@mgolosova, all comments seem to be addressed. I believe that we should follow your suggestion in trello - merge this pull request into pyDKB (if no further changes are required, of course) before proceeding with further tests.

mgolosova:
@Evildoor, I am sorry for putting this PR aside for such a long time.

Tried to restore what was going on here, but when I try to run tests I see things like this:
```
$ git fetch
$ git checkout origin/pyDKB-tests
HEAD is now at 7dad261... Divide test_ProcessorStage.py per how-to: tests.
$ cd pyDKB/dataflow/stage/
$ python -m unittest discover  -t ../../../
<...>
ERROR: test_override_dest_f_mode_m (pyDKB.dataflow.stage.tests.ProcessorStage.test_ProcessorStage.ProcessorStageArgsTestCase)
----------------------------------------------------------------------
  File "/home/dkb/dkb-dev.git/Utils/Dataflow/pyDKB/dataflow/stage/tests/ProcessorStage/test_ProcessorStage.py", line 222, in f
    self.stage.parse_args(['--' + arg, val, '--mode', mode_val])
  File "/home/dkb/dkb-dev.git/Utils/Dataflow/pyDKB/dataflow/stage/AbstractStage.py", line 204, in parse_args
    sys.exit(1)
SystemExit: 1
<...>
```

I don't think it is how the test supposed to react. Have I missed something, maybe tests should be run with some additional parameters?

The part with `sys.exit(1)` came here with merge from `pyDKB` branch (c81a096), so it was not here when tests were created -- but now it seems like the tests code requires some update to handle the `exit` in case of wrong arguments.

---

I believe this PR is still actuall since untit tests for the library would be a Good Thing, but for now I can't review it.

PR #198, opened 2019-02-04
Stage 009 config example's duplicate.

Evildoor:
Currently there are two examples of the config file for the stage - the one in `Utils/Dataflow/config/` is more up to date, so the one in stage 009's readme is replaced with a link to the former.

Noticeable difference between the two examples is the description of the change introduced in bef4178 -
`Utils/Dataflow/config/009.cfg.example` was updated, but 009's readme was not. Because of this, using the readme's config for launching the stage would lead to failure due to missing query parametric values.

### DIFF
--- a/Utils/Dataflow/009_oracleConnector/README
+++ b/Utils/Dataflow/009_oracleConnector/README
@@ -24,37 +24,7 @@ To run `prodsys2ES` query only, use PLAIN mode:
 
   ./oracle2JSON.py --config $CONFIG --mode PLAIN
 
-The configuration file:
-```
-[oracle]
-dsn: USER/PASSWORD@HOST:PORT/SERVICE_NAME
+A template for the configuration file can be obtained in Dataflow's
+corresponding directory:
 
-[queries]
-# relative path to tasks query
-tasks: query/prodsys2ES.sql
-# relative path to datasets query
-datasets: query/datasets.sql
-
-[process]
-# Query response processing mode:
-#  PLAIN   -- pass records to output independently of each other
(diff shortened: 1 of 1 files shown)


Utils/Dataflow/009_oracleConnector/README
```diff
@@ -24,37 +24,6 @@ To run `prodsys2ES` query only, use PLAIN mode:
...
-# Current offset file
-# (relative to the dir with the config file or absolute path)
-# offset_file = .offset
-```
-
-Current offset by default is stored in the same directory as the config file,
-  but it can be configured.
+A template for the configuration file can be obtained in Dataflow's corresponding directory:
+
+  dkb/Utils/Dataflow/config/009.cfg.example
```
mgolosova: I'd rather use git's notation for paths within repository: `Utils/Dataflow/config/009.cfg.example`, just as `git` shows when asked about full names:
```
$ pwd
/home/mgolosova/work/DKB/git/Utils/Dataflow/config
$ git ls-files --full-name
Utils/Dataflow/config/009.cfg.example
Utils/Dataflow/config/095.cfg.example
Utils/Dataflow/config/README
```

And about formatting -- could you pleas:
* wrap the text at 80 characters;
* add newline at the end of file (see red symbol at the end of last line in the GitHub representation)?

Evildoor: Sure, will do.

Evildoor: Done.

mgolosova: Thank you!

PR #184, opened 2018-10-18
[WIP] Added a NDJSON compare script based on jsondiff and capable of summar…

gubinm:
Added a NDJSON compare script based on jsondiff and capable of summarizing the results.

PR #201, opened 2019-02-08
Task Chain

Evildoor:
- [x] Modify 009 to construct the chain_data for each task.
- [x] Modify 016 to extract the chain_id from chain_data and save the latter as array of integers.
- [x] Make sure that the modified stages work properly and produce the required results (fill ES on bamboo, perform the test run of the dataflow, etc).
- [x] Merge master here and update data samples. (waiting for #199) 
- [x] Properly state the reconstruction script's role in its description - according to my understanding, it was used as a proof of concept, and #232 was constructed basing on it, but it won't be a part of the actual working code.

---
~Waiting for #199~ *(merged)*


Evildoor:
Note: three commits regarding comments and rearrangement in process() are not related to the task chains directly, but I believe that they are fitting into this PR because the chain_data processing has to be put into process() and commented properly, and these changes help in doing so.

Utils/Dataflow/016_task2es/task2es.py
```diff
@@ -183,6 +183,29 @@ def input_events(data):
     return result
 
 
+def add_chain_id(data):
+    """ Add chain_id into data - taskid of the task chain's root.
+
+    :param data: data to be updated, must contain taskid and proper chain_data
+    (string of numbers separated by commas).
```
mgolosova: What about indent the second line of the `:paam data:` desription to be aligned with the description, not the `:param...:`?
```
    :param data: data to be updated, must contain taskid and proper chain_data
                 (string of numbers separated by commas).
```

Utils/Dataflow/016_task2es/task2es.py
```diff
@@ -183,26 +183,32 @@ def input_events(data):
...
-                         'chain_data "%s" seems to be incorrect.\n'
+        sys.stderr.write('(WARN) Task %s: cannot transform chain_data "%s", '
+                         'it seems to be incorrect and will be discarded.\n'
                          % (taskid, chain_data))
+        del data['chain_data']
         return False
-    data['chain_id'] = int(chain_data.split(',')[0])
+    chain_data = [int(i) for i in chain_data.split(',')]
+    data['chain_id'] = chain_data[0]
+    data['chain_data'] = chain_data
```
mgolosova: What do you think about additional logic, like "If there is no `chain_data`, then `chain_id` = `task_id` and `chain_data` = `[task_id]`"? We may be sure that 009 won\`t produce empty `chain_data` for now, but who said that:
* it will never change;
* this stage won\`t get data from somewhere else?

Utils/Dataflow/016_task2es/task2es.py
Evildoor: Done. I did some tests - actually, this is rather important: indent aligned with left side of `:param:` breaks the structure of the resulting documentation.

Utils/Dataflow/016_task2es/task2es.py
Evildoor: Done, assuming that you meant "if there is no VALID `chain_data`" - `chain_data` containing a jumble of random symbols does not look any more useful to me than the field missing outright.

Utils/Dataflow/test/utils/reconstruction.py
```diff
@@ -0,0 +1,99 @@
...
+
+
+def get_chain_data(es, index, chain_id):
+    ''' Get chain_data of all tasks that have given chain_id. '''
+    srch = {'query': {'match': {'chain_id': chain_id}}}
+    rtrn = []
+    fr = 0
+    sz = 100
+    while True:
+        results = es.search(index=index, body=srch, from_=fr, size=sz)
```
mgolosova: It is up to Mikhail, but I think if we\`re going to keep all the data in memory (`rtrn` variable) anyway, then it is OK to get them all at one gulp.
I mean of course we don\`t know how long the chain is, but still it must be OK to set `sz` to something like 2000 (used by Mikhail [here](https://github.com/PanDAWMS/panda-bigmon-atlas/blob/devel/atlas/dkb/views.py#L145) and [here](https://github.com/PanDAWMS/panda-bigmon-atlas/blob/devel/atlas/dkb/views.py#L643)).

Utils/Dataflow/test/utils/reconstruction.py
```diff
@@ -0,0 +1,99 @@
...
+    results = results['hits']['hits']
+    if not results:
+        return False
+    else:
+        return results[0]['_source'].get('chain_id')
+
+
+def get_chain_data(es, index, chain_id):
+    ''' Get chain_data of all tasks that have given chain_id. '''
+    srch = {'query': {'match': {'chain_id': chain_id}}}
```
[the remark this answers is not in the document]
mgolosova: And, again, you might want to use `"source": ["chain_data"]`; at least as long as you return only the IDs, not the whole set of task metadata.

Utils/Dataflow/test/utils/reconstruction.py
```diff
@@ -0,0 +1,99 @@
...
+        for hit in results['hits']['hits']:
+            chain_data = hit['_source'].get('chain_data')
+            if chain_data:
+                rtrn.append(chain_data)
+        fr += sz
+    return rtrn
+
+
+def construct_chain(data):
+    ''' Build a chain from the list of its tasks' chain_data. '''
```
mgolosova: The function looks fine (I mean, does what it is supposed to), yet it definitely lacks the description of input / output data.
The output structure is a bit complicated to construct and read... wouldn\`t something like nested dicts make it easier?

Utils/Dataflow/009_oracleConnector/README
```diff
@@ -23,3 +23,23 @@ A template for the configuration file can be obtained in Dataflow's
 corresponding directory:
 
   Utils/Dataflow/config/009.cfg.example
+
+3. Task Chain
+-------------
+The query prodsys2ES includes the subquery that obtains the task's chain_data:
```
mgolosova: `...includes...` --> `...contains...`?
I am not sure which of them is better in this context, so just a suggestion.

Utils/Dataflow/016_task2es/README
```diff
@@ -9,6 +9,8 @@ Transform task metadata to fit ES scheme:
 * split comma-separated 'hashtag_list' to the actual list structure
 * turn hashtags to lower case
 * split dot-separated 'output_formats' to the actual list structure
+* extract 'chain_id' from 'chain_data' and turn the latter into array
+of integers
```
mgolosova: Could you please indent the last line to be aligned with the rest of the text? If later there will be more items, it will make the list look more accurate and prevent stumbling on this line.
Compare:
```
* some item here
* extract 'chain_id' from 'chain_data' and turn the latter into array
  of integers
* yet another item describing the stage
```
vs.
```
* some item here
* extract 'chain_id' from 'chain_data' and turn the latter into array
of integers
* yet another item describing the stage
```

Utils/Dataflow/016_task2es/task2es.py
```diff
@@ -183,28 +183,65 @@ def input_events(data):
...
+                 (string of numbers separated by commas).
+    :type data: dict
+
+    :return: True if update was successful, False otherwise (chain_id and
+             chain_data will be updated so that the task is considered
+             independent)
+    :rtype: bool
+    """
+    if type(data) is not dict:
+        return False
```
mgolosova: Would be nice to have some notification here.

Utils/Dataflow/test/utils/reconstruction.py
```diff
@@ -0,0 +1,99 @@
...
+def es_connect():
+    ''' Establish a connection to elasticsearch.
+
+    TODO: take parameters from es config.
+    '''
+    return elasticsearch.Elasticsearch(['localhost:9200'])
+
+
+def get_chain_id(es, index, taskid):
+    ''' Get chain_id by given taskid. '''
```
mgolosova: Detailed docstrings are useful not only for automatic documentation production, but also for the cpde readability. I mean the missed parameters and return value descriptions (here and in the other functions) ;)

mgolosova:
Oh yes, totally forgot...my personal (un)favourite part of this kind of PRs: samples update.
This PRs changes behaviour of 009 and 016, so their output data are changed for sure; similarly output data of all the other stages in the `009 --> 091(in) --> 025 --> 016 --> 019` chain are changed as well.

If in #199 you will decide to change 2016's samples, first decide which of the PRs should be merged first -- another one will have to merge `master` (before being merged *to* `master`).

Evildoor:
Hm. I think we should act as following:

1. Don't change the 2016 samples, finish other #199's problems and merge it into master.
2. Merge master into this branch.
3. Update samples here.

2016's samples to be updated are 016, 025 and 019 (task part), and these (together with 009 and 091) have to be updated in this branch as well. Therefore, I believe that we can save ourselves some work and not update them in #199 separately, since they will be overwritten here - after all, the samples are more or less accurate: as you said in #199 (and I agree with you), a skipped stage here and there is distracting, but not critical.

Utils/Dataflow/009_oracleConnector/README
Evildoor: I can't find any solid reason to use any of the two options over another in this case, so I'll leave the text as is per my intuition.

Utils/Dataflow/test/utils/reconstruction.py
```diff
@@ -0,0 +1,99 @@
...
+    srch = {'query': {'match': {'_id': taskid}}}
```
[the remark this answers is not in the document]
Evildoor: Done. Thank you, I did not know that.

Utils/Dataflow/test/utils/reconstruction.py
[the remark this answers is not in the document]
Evildoor: Done. I've thought about `"term"`, but either forgot to use it or copy-pasted from wrong script. Thank you.

Utils/Dataflow/test/utils/reconstruction.py
Evildoor: Done.

Utils/Dataflow/test/utils/reconstruction.py
Evildoor: Added description.
Current version was approved by Mikhail in email, over the nested dicts that were suggested as well.

Utils/Dataflow/016_task2es/README
Evildoor: Done.

Utils/Dataflow/016_task2es/task2es.py
Evildoor: Done.

Utils/Dataflow/test/utils/reconstruction.py
Evildoor: Done. Yes, you are right and I agree with you. It wasn't done before due to the reason I've stated in email - possibility of changes requested by Mikhail that may mean tearing everything down and starting more or less from scratch.

mgolosova:
@Evildoor,
> 2016's samples to be updated (in #199) are 016, 025 and 019 (task part), and these (together with 009 and 091) have to be updated in this branch as well. Therefore, I believe that we can save ourselves some work and not update them in #199 separately, since they will be overwritten here

The thing is not only to have samples up to date, but also to demonstrate what has changed after this or that changes in code. This means that first you need make samples up-to-date (where the "date" is right before the changes in the code), and then -- update them again. If not, then there will be too many changes in the samples: added or removed fields (some -- due to the using of input data from different stage, some -- due to the changes in code performed in this PR); maybe new values (if data has changed in the orinal data sources, like Rucio/AMI/ProdSys); etc.

And whatever happens, there should be a "summary" (what happened to what fields) and expaination (why it happened -- changes in original data, in input data origin, in processing code) in the commit log. This summary is required for a simple reason: make sure yourself that you understand every changes and not missed anything that corrupts data. And the less items are to listed at once, the better -- I mean, easier to understand not to miss anything.
Annoying, I know. That\`s why "samples generator" (#180) and [compare_ndjson_files.py](https://github.com/PanDAWMS/dkb/blob/master/Utils/Dataflow/test/utils/compare_ndjson_files.py) (or #184) ;) And that\`s why they need improvements...

---

So I\`d suggest to (fully) update samples in #199; merge `master` with updated version here; update according to the new code.

Utils/Dataflow/test/utils/reconstruction.py
```diff
@@ -48,7 +78,23 @@ def get_chain_data(es, index, chain_id):
...
+    task has one of the previous ones as its parent (parent's output includes
+    child's input).
+
+    :param data: chain_data of all tasks in the chain
+    :type data: list
+
+    :return: constructed chain, format is
+             {
+                 ...
+                 taskidN: [taskids of all tasks which are children of taskidN]
```
mgolosova: Maybe:
```
taskidN: [ childN1_id, childN2_id, ...]
```
?

Or at least "`child task IDs`"? The shorter things are in the docs, the easier to "scan" them through when trying to figure out out what\`s going on in general -- or to find needed piece of information when need to clarify this or that moment.

Utils/Dataflow/test/utils/reconstruction.py
```diff
@@ -77,6 +123,15 @@ def check(chain, chain_data, task_id, indent=0):
...
     matching chain_data for each task and display it as well.
+
+    :param chain: constructed chain
+    :type chain: dict
+    :param chain_data: chain_data of all tasks in the chain
+    :type chain_data: list
+    :param task_id: task from which to start
+    :type task_id: int
+    :param indent: indent, also can be used to measure the chain's length
+    :type indent: int
```
mgolosova: Maybe specify that `indent` is some kind of the "recursion parameter"? I mean, to clarify that it is OK not to think about it when calling the function; or call it "initial indent" and say something about indentation in the description (or provide some example of the output).

I get that it is a debug function, yet as I try to understand the PR fully, I make everything to make my job easier and not to dig through the code every time I try to remember what is what ;)

Evildoor:
> ...
> 
> So I'd suggest to (fully) update samples in #199; merge `master` with updated version here; update according to the new code.

Fair enough, will do.

Utils/Dataflow/test/utils/reconstruction.py
Evildoor: Done.

Utils/Dataflow/test/utils/reconstruction.py
Evildoor: Right you are. Done.

mgolosova:
Updated PR title and description just to make clear that we are waiting for smth and what are we waiting for :)

Evildoor:
Updated the samples - as said in commit log, it seems that ProdSys/Rucio data corresponding to our 2018 interval have changed significantly.

Utils/Dataflow/016_task2es/task2es.py
```diff
@@ -183,6 +183,29 @@ def input_events(data):
...
+
+    :return: True if update was successful, False otherwise (data is unchanged)
+    :rtype: bool
+    """
+    if type(data) is not dict:
+        return False
+    taskid = data.get('taskid')
+    chain_data = data.get('chain_data')
+    if not chain_data or not chain_data.replace(',', '').isdigit():
+        sys.stderr.write('(WARN) Task %s: cannot obtain chain_id, '
```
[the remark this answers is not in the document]
Evildoor: Done.

Utils/Dataflow/test/utils/reconstruction.py
```diff
@@ -29,7 +43,23 @@ def get_chain_id(es, index, taskid):
...
 
 def get_chain_data(es, index, chain_id):
-    ''' Get chain_data of all tasks that have given chain_id. '''
+    ''' Get chain_data of all tasks that have given chain_id.
+
+    :param es: elasticsearch client
+    :type es: elasticsearch.client.Elasticsearch
+    :param index: index containing tasks
+    :type index: str
+    :param chain_id: chain_id - taskid of the chain's root
```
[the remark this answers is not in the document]
Evildoor: Looks like a careless copy-paste or something like that - as you correctly point out, there is no reason to make a repetition. Fixed.

Evildoor:
@mgolosova - I believe that the merge can be performed now. Please, approve once more.

PR #205, opened 2019-02-11
pyDKB stage fix (pt.4)

mgolosova:
Make pyDKB stage objects to log configuration.

---
+ (pt.3) #204


mgolosova:
@anastasiakaida, re-readed changes, and that\`s what hit me this time: there\`s no need in `ProcessorStage.log_configuration()` method: it just calls the ancestor\`s one.
I think I\`m going to remove it now...

PR #207, opened 2019-02-11
pyDKB stage fix (pt.6)

mgolosova:
pyDKB refactoring:
* unused method removed;
* list turned into iterator in InputStream class.

---
+ (pt.5) #206

### DIFF
--- a/Utils/Dataflow/pyDKB/dataflow/communication/stream/Stream.py
+++ b/Utils/Dataflow/pyDKB/dataflow/communication/stream/Stream.py
@@ -14,6 +14,7 @@ class Stream(object):
     """ Abstract class for input/output streams. """
 
     message_type = None
+    EOM = None
     _fd = None
 
     def __init__(self, fd=None, config={}):
--- a/Utils/Dataflow/pyDKB/dataflow/communication/stream/InputStream.py
+++ b/Utils/Dataflow/pyDKB/dataflow/communication/stream/InputStream.py
@@ -25,7 +25,7 @@ class InputStream(Stream):
         if self.EOM == '\n':
             self.__iterator = iter(fd.readline, "")
         elif self.EOM == '':
-            self.__iterator = iter([fd.read()])
+            self.__iterator = iter(fd.read, "")
         else:
             self.__iterator = custom_readline(fd, self.EOM)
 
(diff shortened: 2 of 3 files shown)


Utils/Dataflow/pyDKB/dataflow/communication/stream/Stream.py
```diff
@@ -14,6 +14,7 @@ class Stream(object):
     """ Abstract class for input/output streams. """
 
     message_type = None
+    EOM = None
```
anastasiakaida: @mgolosova 
why the default EOM is None, not '\n'?

mgolosova: `None` sounds like fine default value for variable declaration. Why `'\n'` would be better?

My point is that objects of this class should never be used before the `configure()` method is called (especially because it *is* called in `__init__()`), in which `EOM` is set as `config['eom']` [1]:
https://github.com/PanDAWMS/dkb/blob/8477646a4e88e8a15ca8c733bd4f35449e825b93/Utils/Dataflow/pyDKB/dataflow/communication/stream/Stream.py#L43-L48
Setting any default value but `None` could mislead to thoughts like "If I do not pass any EOM, so it should be `'\n'` (or whatever)" -- which is not the case. `None` kind of says: this property is either set during object initiation or with proper method(s).

Also, even thought `ARGS.eom` is *set* to `'\n'` during configuration process by default, in terms of `argsparse` its default value is `None` as well:
https://github.com/PanDAWMS/dkb/blob/8477646a4e88e8a15ca8c733bd4f35449e825b93/Utils/Dataflow/pyDKB/dataflow/stage/AbstractStage.py#L125-L135
...so, again, even if nothing happens with `EOM`, default value specified here will be rewritten by that default `None`.

If one day we need to use this class separately from the `stage` context and this `None` causes any problems, it will be changed, yet for now I don\`t see if `None` can lead to any issues... but maybe I\`m missing something, so any comments are welcome :)

[1] It may lead to `KeyError`, but not if we use this class through `stage`, which passes `ARGS` as config and thus `eom` key always exists; so for now it is fine; if unit tests later reveal any problems, they will be fixed in their time.

anastasiakaida: @mgolosova 
Ok. I see. So I agree with your point of view, it seems reasonable to leave `EOM` here, let's say, as a "very default" value. :)

PR #208, opened 2019-02-11
pyDKB stage fix (pt.7)

mgolosova:
EOM handling:
* test cases;
* library fixes.
---
+ (pt.6) #207


mgolosova:
@anastasiakaida,

> There should be "reads" instead of "read" in the 54dfc97 description.

Do you mean "is/be fully read" parts? It\`s "to be + Past Participle", wich is "read" for "read" ;)

anastasiakaida:
@mgolosova 
I apologize a lot! It's about 42cfb8f.

```
When stage read input with custom EOM, it read first message, and then
'resetted' input stream, loosing buffered in `custom_readline()` generator
data.
```
Maybe ```When stage reads <...>, it read first <...>```?

mgolosova:
@anastasiakaida, no, I meant to say it in the past (which is also "read" ;) ). Will change to "used to read" to avoid confuses...

anastasiakaida:
@mgolosova Ok, thank you!

mgolosova:
@anastasiakaida, thank you for the reminder, rewording done.

PR #209, opened 2019-02-12
Stage 009: Improve query saving procedure.

Evildoor:
Problems with error handling:

- A single missing parameter leads to KeyError. Add interception for it
to display a message that points to the configuration.
- The whole configuration missing leads to (even more puzzling)
ORA-00911. Add a warning when no parameters were configured that also
points towards configuration. Since it is possible to use a query
without parameters, no warning should be displayed in such case. Add
a check to only display the warning when the query contains single '%'
(double percent '%%' means that '%' was escaped, so it's allowed).

Also, a file is opened but not closed in the function - fixed that.

### DIFF
--- a/Utils/Dataflow/009_oracleConnector/dbConnection.py
+++ b/Utils/Dataflow/009_oracleConnector/dbConnection.py
@@ -117,8 +117,8 @@ class OracleConnection(dbConnection):
     def save_query_file(self, qname, src_filename, params=None):
         """ Read query from file and save it in query hash. """
         try:
-            f = open(src_filename)
-            q = f.read().rstrip().rstrip(';')
+            with open(src_filename) as f:
+                q = f.read().rstrip().rstrip(';')
             if isinstance(params, dict):
                 q = q % params
             elif '%' in q.replace('%%', ''):
--- a/Utils/Dataflow/009_oracleConnector/dbConnection.py
+++ b/Utils/Dataflow/009_oracleConnector/dbConnection.py
@@ -124,8 +124,9 @@ class OracleConnection(dbConnection):
             elif '%(' in q:
                 # Check for '%(' is done because a valid query without
                 # parameters (and their configuration) is possible.
-                sys.stderr.write("(WARN) No query parameters were configured "
-                                 "for '%s'. This can result in "
+                sys.stderr.write("(WARN) '%s': query seems to be parametric, "
+                                 "but no query parameters were configured. "
+                                 "This can result in "
                                  "'ORA-00911' error.\n" % qname)
             self.queries[qname]['query'] = q
         except IOError, err:
--- a/Utils/Dataflow/009_oracleConnector/dbConnection.py
+++ b/Utils/Dataflow/009_oracleConnector/dbConnection.py
@@ -121,8 +121,8 @@ class OracleConnection(dbConnection):
                 q = f.read().rstrip().rstrip(';')
             if isinstance(params, dict):
                 q = q % params
-            elif '%' in q.replace('%%', ''):
-                # Check for single '%' is done because a valid query without
+            elif '%(' in q:
+                # Check for '%(' is done because a valid query without
                 # parameters (and their configuration) is possible.
                 sys.stderr.write("(WARN) No query parameters were configured "
                                  "for '%s'. This can result in "
--- a/Utils/Dataflow/009_oracleConnector/dbConnection.py
+++ b/Utils/Dataflow/009_oracleConnector/dbConnection.py
@@ -121,9 +121,20 @@ class OracleConnection(dbConnection):
             q = f.read().rstrip().rstrip(';')
             if isinstance(params, dict):
                 q = q % params
+            elif '%' in q.replace('%%', ''):
+                # Check for single '%' is done because a valid query without
+                # parameters (and their configuration) is possible.
+                sys.stderr.write("(WARN) No parameters were configured "
+                                 "for '%s' query. This can result in "
+                                 "'ORA-00911' error.\n" % qname)
             self.queries[qname]['query'] = q
         except IOError, err:
             sys.stderr.write("(ERROR) Failed to read query file: %s\n" % err)
+            return False
+        except KeyError, err:
+            sys.stderr.write("(ERROR) Failed to update query: expected "
+                             "parameter %s not found in the "
+                             "configuration.\n" % err)
             return False
 
         return True
--- a/Utils/Dataflow/009_oracleConnector/dbConnection.py
+++ b/Utils/Dataflow/009_oracleConnector/dbConnection.py
@@ -124,17 +124,15 @@ class OracleConnection(dbConnection):
             elif '%' in q.replace('%%', ''):
                 # Check for single '%' is done because a valid query without
                 # parameters (and their configuration) is possible.
-                sys.stderr.write("(WARN) No parameters were configured "
-                                 "for '%s' query. This can result in "
+                sys.stderr.write("(WARN) No query parameters were configured "
+                                 "for '%s'. This can result in "
                                  "'ORA-00911' error.\n" % qname)
             self.queries[qname]['query'] = q
         except IOError, err:
             sys.stderr.write("(ERROR) Failed to read query file: %s\n" % err)
             return False
         except KeyError, err:
-            sys.stderr.write("(ERROR) Failed to update query: expected "
-                             "parameter %s not found in the "
-                             "configuration.\n" % err)
+            sys.stderr.write("(ERROR) Query parameter missing: %s\n" % err)
             return False
 
         return True
--- a/Utils/Dataflow/config/009.cfg.example
+++ b/Utils/Dataflow/config/009.cfg.example
@@ -8,7 +8,7 @@ tasks: %__009_query_dir__%/prodsys2ES.sql
 datasets: %__009_query_dir__%/datasets.sql
 
 [queries.params]
-# Parametric values for all queries:
+# Query parameters for all queries:
 # string '%(<param_name>)s' in sql file will be replaced
 # with the value of <param_name> (if value is quoted, quotes
 # will appear in the query as well).


Utils/Dataflow/009_oracleConnector/dbConnection.py
```diff
@@ -121,10 +121,21 @@ def save_query_file(self, qname, src_filename, params=None):
...
+                                 "for '%s' query. This can result in "
+                                 "'ORA-00911' error.\n" % qname)
             self.queries[qname]['query'] = q
         except IOError, err:
             sys.stderr.write("(ERROR) Failed to read query file: %s\n" % err)
             return False
+        except KeyError, err:
+            sys.stderr.write("(ERROR) Failed to update query: expected "
+                             "parameter %s not found in the "
+                             "configuration.\n" % err)
```
mgolosova: When I read the log:
```
(INFO) Establish connection to the DB...
(INFO) Connection established.
(ERROR) Failed to update query: expected parameter 'production_or_analysis_cond' not found in the configuration.
(WARN) Failed to save query 'tasks'
(ERROR) Failed to update query: expected parameter 'production_or_analysis_cond' not found in the configuration.
(WARN) Failed to save query 'datasets'
(ERROR) Queries seem to be misconfigured. Exiting.
```
my first question was "what does it mean, ``update query''?"

Maybe make it less literary, like... 
```
(INFO) Establish connection to the DB...
(INFO) Connection established.
(ERROR) Query parameter missed: 'production_or_analysis_cond' (config section 'queries.params').
(WARN) Failed to save query 'tasks'
(ERROR) Query parameter missed: 'production_or_analysis_cond' (config section 'queries.params').
(WARN) Failed to save query 'datasets'
(ERROR) Queries seem to be misconfigured. Exiting.
```

Not that I think this variant is perfect, I wish it was better... so any rewording will be appreciated ;)
But the idea is to make error message less wordy (like `Query parameter missed`), while add more specific details (in which section of the config one should look for parameters). After all, `(ERROR)` itself supposes that some failure is encountered, so we can skip saying it explicitly.

Utils/Dataflow/009_oracleConnector/dbConnection.py
```diff
@@ -121,10 +121,21 @@ def save_query_file(self, qname, src_filename, params=None):
             q = f.read().rstrip().rstrip(';')
             if isinstance(params, dict):
                 q = q % params
+            elif '%' in q.replace('%%', ''):
+                # Check for single '%' is done because a valid query without
+                # parameters (and their configuration) is possible.
+                sys.stderr.write("(WARN) No parameters were configured "
+                                 "for '%s' query. This can result in "
+                                 "'ORA-00911' error.\n" % qname)
```
mgolosova: Not sure that I understand, why '%%' looks fine for you?

And '%' is still a valid symbol for LIKE statements... but you\`re right, it is hardly possible to know if this or that `%` is valid until one tries to compile query, so it is better to show the warning.

Utils/Dataflow/009_oracleConnector/dbConnection.py
Evildoor: > my first question was "what does it mean, ``update query''?"

Oh, I see. Can indeed be confusing.

> add more specific details (in which section of the config one should look for parameters)

I thought about something like this, but then it occured to me that this is a method that applies a dictionary to a query, and knows nothing about config files or their sections. Would it be okay to mention them - if we consider the fact that it is possible in the future for this method to be called while taking parameters not from the config file?

How about this:

`(ERROR) Query parameter missing: 'production_or_analysis_cond'`

And we change the example config's description of 'queries.params' from

    # Parametric values for all queries:

to

    # Query parameters for all queries:

I think this should be enough to find the section in the config, and

`(ERROR) Queries seem to be misconfigured. Exiting.`

points to the config.

P.S. Other related messages should be changed to use the term `query parameter` too in such case.
P.P.S. Actually, the method knows nothing about Oracle errors as well, but it is preparing queries for Oracle and warns about a possible outcome - so this, I believe, is fine.

mgolosova: > this is a method that applies a dictionary to a query, and knows nothing about config files or their sections. Would it be okay to mention them...?

Right you are, config file and sections belong to 009, while dbConnection seem to be a candidate for a move to (some) common library.

Then OK, let\`s leave only the first part, as you suggest, and update config example. One will possibly use his own config file without comments, yet there still is a proper config example; yet somewhere in configdocs (one day) it should be described as well; example is good, documentation is better.

P.S. The better solution, as I see now, would be to turn all errors and `return False` into exceptions. Then we could raise proper exception, something like `InvalidQueryParameter`, catch it in the main code and output proper log message, pointing to the section in config file. But it would require quite a refactoring, so... not this time, I believe?
Funny thing, when I wrote the code I intentionally tried to avoid exceptions... yet it is Pythonic style, indeed ;)

mgolosova:
[Stage 009: query parametric values: errors handling](https://trello.com/c/OqUs89QI/71-stage-009-query-parametric-values-errors-handling)

Utils/Dataflow/009_oracleConnector/dbConnection.py
Evildoor: > Not sure that I understand, why '%%' looks fine for you?

Let's take a look at different possible queries regarding this issue:

    1) SELECT a AS b FROM c;
    2) SELECT %(a)s AS b FROM c;
    3) SELECT a AS b%% FROM c;

In this function we perform `q % params` operation - it does nothing with 1 and 3, and changes 2 if the parameter 'a' was configured properly. The warning is about this operation of formatting, not the correctness of query in terms of Oracle - it mentions Oracle's problems because they can appear as a consequence of wrong formatting. If the configuration is missing, then the warning must appear only in case of 2: 1 is correct and needs no configuration, and 3 is also correct in terms of formatting. It's incorrect in terms of Oracle and will throw the same 'ORA-00911', but if we throw the same warning when 3 is encountered it would be misleading because the warning hints 'something with configuration', while actually the problem would be 'something with query'.

However...

> And '%' is still a valid symbol for LIKE statements... but you`re right, it is hardly possible to know if this or that '%' is valid until one tries to compile query, so it is better to show the warning.

... this is reasonable as well, so we can play on the safe side and just warn when any '%' is present. What do you think?

Utils/Dataflow/009_oracleConnector/dbConnection.py
Evildoor: Done.

Utils/Dataflow/009_oracleConnector/dbConnection.py
```diff
@@ -119,6 +119,7 @@ def save_query_file(self, qname, src_filename, params=None):
         try:
             f = open(src_filename)
             q = f.read().rstrip().rstrip(';')
+            f.close()
```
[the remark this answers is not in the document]
Evildoor: Yes, that construction is usually better, but I wanted to make the change as small as possible because it's a bit outside of this PR's aim. Since you suggested it - done.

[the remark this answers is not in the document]
mgolosova: No matter how small the correction is -- if it is out of place, it is out of place ;) If a reviewer pay no attention to the change because it is "just one line added, nothing to think about", then shame on the reviewer.

You\`re right saying:
> it's a bit outside of this PR's aim.

To make this change less out-of-place, there are two possible solutions:
* change the PR description (not "query parameters error handling", but "stage (minor?) improvements");
* move it to another PR.

Utils/Dataflow/009_oracleConnector/dbConnection.py
mgolosova: > Let's take a look at different possible queries regarding this issue:
> 
> ```
> 1) SELECT a AS b FROM c;
> 2) SELECT %(a)s AS b FROM c;
> 3) SELECT a AS b%% FROM c;
> ```
> In this function we perform `q % params` operation - it does nothing with 1 and 3, and changes 2 if the parameter 'a' was configured properly.

Not exactly: after `q % params`  case 3 will become `SELECT a AS b% FROM c;`. But it is not to happen when no parameters specified (due to the check `if isinstance(params, dict)` before this statement).

Maybe add something to sound like "it looks like a parametric query, so here\`s the warning."?
```
(WARN) 'tasks': query seems to be parametric, but no query parameters were configured. May produce 'ORA-00911' error at compilation."
```

---

And, if (by chance) we want to be nice and give user a possibility to switch this warning off...
```
[queries]
 # absolute (or relative to config file) path to tasks query
 tasks: %__009_query_dir__%/prodsys2ES.sql
 # absolute (or relative to config file) path to datasets query
 datasets: %__009_query_dir__%/datasets.sql

+# Uncomment to switch off check for parameters configuration:
+# parametric_check: False
```
?

`True` by default, but if there are `LIKE` statements with proper `%` in the query and the warning is too annoying, it can be turned off.
Extra-complication, yes, but as it is obvious that the warning may appear when it is not welcome...

---

I also thought about more elaborate checks, like:
* step 1: is there '%'? If yes:
* step 2: try `query % {}`:
  * KeyError -- most likely it is a missed parameter;
  * ValueError, TypeError, passed -- most likely it has nothing to do with parameters.

How it looks and what produces this exceptions:
```
>>> # Looks like parametric query indeed
...
>>> query = 'SELECT %(a)s FROM b WHERE c;'
>>> query % {}
KeyError: 'a'
>>>
>>> # Not a parametric query
...
>>> query = 'SELECT %a FROM b WHERE c;'
>>> query % {}
ValueError: unsupported format character 'a' (0x61) at index 8
>>>
>>> # Not a parametric query
...
>>> query = 'SELECT % FROM b WHERE c;'
>>> query % {}
TypeError: float argument required, not dict
>>>
>>> # Doesn`t look much like a parametric query: parameter must be named
...
>>> query = 'SELECT %s FROM b WHERE c;'
>>> query % {}
'SELECT {} FROM b WHERE c;'
```

And, thinking about this "parameters must be named" -- it is also possible to look for `%(` instead of pure `%`.
In this case I think it is most likely that the warning won\`t appear when there were no sign of parametric query and thus this extra config parameter `parametric_check` is not needed (that much).

Utils/Dataflow/009_oracleConnector/dbConnection.py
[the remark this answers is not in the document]
Evildoor: Changed the PR's title and description.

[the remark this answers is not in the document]
mgolosova: Maybe "query saving procedure" or "function"? Took me a while to read the new title correctly; still it can be just me and Friday night...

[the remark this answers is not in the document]
Evildoor: I thought about giving the exact function name, but we are also changing the config example, so I decided against that. However, "query saving procedure" looks fine to me. Done.

Utils/Dataflow/009_oracleConnector/dbConnection.py
Evildoor: > Maybe add something to sound like "it looks like a parametric query, so here`s the warning."

Makes sense. Done.

> And, thinking about this "parameters must be named" -- it is also possible to look for `%(` instead of pure `%`.

Good idea. Done.

mgolosova:
47c9ce5 -- I think as `open()/close()` clause was turned into `with open()...:`, commit log can be changed (as it is not about `close()` itself, actually).

Evildoor:
Done. Actually, the commit log was changed before, from 'add missing close()' to 'fix missing close()'. However, this way it should be more clear.

mgolosova:
Merged, thank you!

PR #210, opened 2019-02-12
pyDKB stage fix (pt.8)

mgolosova:
Make `EOM`-related requirement more strict: every message should end with `EOM`, no matter if it is the last one in a file or stream, is followed by EOP, etc.

---
+ (pt.7) #208


mgolosova:
@anastasiakaida, sorry, but the review was dismissed when I changed the base branch. Could you re-approve, please?

anastasiakaida:
@mgolosova

Sure!

PR #211, opened 2019-02-12
pyDKB stage fix (pt.9)

mgolosova:
Stage test utility improvements:
* `--list` parameter for test cases overview;
* `--case CASE[,...]` parameter to run specific test case;
* colored states.

---
+ (pt.8) #210

PR #212, opened 2019-02-12
pyDKB stage fix (pt.10) 

mgolosova:
Add test cases 11-18:
* `-i input_dir file file`;
* empty files among input;
* stream mode:
  * empty input;
  * EOM: standard ans custom;
  * broken stream (no EOM in the end): standard and custom EOM.

---
+ (pt.9) #211


mgolosova:
@anastasiakaida, `err` files are binary as well. It is due to the default EOP marker (`\0`) -- it is a non-printable character, which is not supposed to be in the text files.
If you open the file in terminal (e.g. with `less` or `vi`), you will see something like `^@` in the text -- that\`s it.

anastasiakaida:
@mgolosova 
Thank you for the explanation!

PR #213, opened 2019-02-12
pyDKB stage fix (pt.11)

mgolosova:
Add test cases for file output.

---
+ (pt.10) #212


mgolosova:
There were an issue with line numbers in exception trace (due to the removed in pt.4 piece of code), but now it is fixed and to be propagated to the next PRs.

PR #220, opened 2019-02-12
pyDKB stage fix (pt.18)

mgolosova:
Add test case 28 (`--mode s --dest f`).

---
+ (pt.17) #219


mgolosova:
@anastasiakaida, thank you, fixed in #225.

PR #221, opened 2019-02-12
pyDKB stage fix (pt.19)

mgolosova:
1. Test cases: empty EOM for NDjson file and NDjson stream input.
2. Improve stage `help` message.
3. Fixed empty EOM handling in case of stream input.

---
+ (pt.18) #220

PR #222, opened 2019-02-12
pyDKB stage fix (pt.20)

mgolosova:
Added test cases 31-33:
* custom EOP in (f)ile and (s)tream mode;
* (s)tream mode with file input.

---
+ (pt.19) #221


Utils/Dataflow/test/pyDKB/case/32/err
```diff
@@ -0,0 +1,15 @@
...
+(INFO) (ProcessorStage)   input_dir   : 'None'
+(INFO) (ProcessorStage)   eom         : '\n'
+(INFO) (ProcessorStage)   output_dir  : 'out'
+(INFO) (ProcessorStage)   source      : 's'
+(INFO) (ProcessorStage)   eop         : 'EOP'
+(INFO) (ProcessorStage)   mode        : 's'
+(INFO) (ProcessorStage)   config      : 'None'
+(INFO) (ProcessorStage)   input_files : '[]'
+
+Type '{"stop": ""}' to interrupt.
```
anastasiakaida: @mgolosova 
Is it supposed to type ```{"stop": ""}``` exactly this way?

mgolosova: Short answer: yes.
Short question: have you tried? ;)

Long answer: if you check the code of the [json2TTL.py](https://github.com/PanDAWMS/dkb/blob/635aacd70642b78dc1bdb374643ce7eedb158d70/Utils/Dataflow/test/pyDKB/json2TTL.py), that produces this message, you will see rather strange things: as soon as `'stop'` key appears in the input message, it calls some `pre_stop_process()` function, that does nothing but for some reason iterates 3 times before calling the `stop_process()` (which, in its turn, raises `KeyError`).

What is it I am about... yes, I meant to say I\`m surprised you\`ve asked about the format, but not the meaning of all this jazz ;) Answering your question -- yes, exactly this way, as input message must be a JSON with 'stop' in it. It can actually be any other JSON with 'stop' in it (e.g. `["stop"]` or even `"stop"` will also work), but it was the first option that came into my mind when it was written... didn\`t think there can be a JSON object outside the hash ;) JSON format states that strings must go in double quotes, so double quotes are mandatory. Hash... well, as I see now -- not that much. So the message can be changed, or maybe even removed together with the stoping mechanism (together with the test cases expected STDERR update) -- the process can always be stopped (even more silently) by simple `Ctrl+C`. I needed this magic one day to have this long long long debug message from the exception handling mechanism. But maybe, for further tests, it still be needed... don\`t know for sure. So it is still there.

anastasiakaida: @mgolosova I see now -- that was a good convenient idea to implement that as a key-value pair. Tested successfully in a stream mode.

PR #241, opened 2019-04-03
DF/data4es: fixes and minor improvements.

mgolosova:
Minor fixes:
* removed obsolete piece of code;
* typo fixed (log message);
* ensure all error messages go to STDERR;
* fix typo in variable name (fortunately, it didn\`t affect anything).

Improvements:
* add `log()` function to unify logging;
* add process start/end log messages with (overall) timing.

### DIFF
--- a/Utils/Dataflow/run/data4es-start
+++ b/Utils/Dataflow/run/data4es-start
@@ -1,6 +1,5 @@
 #!/bin/bash -l
 
-START_OFFSET="09-05-2016 00:00:00"
 BATCH_SIZE=100
 DEBUG=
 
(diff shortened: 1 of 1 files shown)


Utils/Dataflow/run/data4es-start
```diff
@@ -1,11 +1,17 @@
 #!/bin/bash -l
 
-START_OFFSET="09-05-2016 00:00:00"
 BATCH_SIZE=100
-DEBUG=
+DEBUG=1
```
Evildoor: Shouldn't this change be mentioned in commit/PR?

Utils/Dataflow/run/data4es-start
```diff
@@ -1,11 +1,17 @@
...
 
 base_dir=$( cd "$(dirname "$(readlink -f "$0")")"; pwd)
 
+log() {
+  level=INFO
+  [ $# -eq 2 ] && level="$1" && shift
+  msg="$1"
+  echo "($level) `date +'%d-%m-%Y %T'` (data4es) $msg" >&2
```
Evildoor: Since log() is a candidate for moving into a library, it could be useful to replace this `data4es` with a variable - a value passed to log(), or script name, or something else. Probably, not now and not in this PR.

Utils/Dataflow/run/data4es-start
mgolosova: I´d rather said "it shouldn´t be in the commit at all". Will be fixed, thank you!

Utils/Dataflow/run/data4es-start
mgolosova: Good point; I decided not to hold it back for too long, so here it is (b5d6284).
Thank you!

PR #243, opened 2019-04-08
DF/09: add possibility to specify timestamps time zone (set to UTC).

mgolosova:
Previously 'naive' datetime objects were used, meaning in fact that we
treat `timestamp` field as timestamp in the local time zone. It has some
effect only when `now()` is used as a right limit of the integration
interval.

It is said (by M.Borodin) that `timestamp` is supposed to be in UTC, but
in fact it can be UTC or CET or CEST depending on when and who writes
the data.

Setting timestamp time zone to local time zone meant that records,
written with proper UTC, are likely to be missed (as we check every interval
one or two hours earlier then records are written).

Setting timestamp time zone to the UTC means that for records with
timestamp in CE(S)T we won't see changes as soon as they are written to
the Oracle DB; but they will be handled later. It is not best, but still
better than missing records. When things get standardized and all the
update operations start setting UTC timestamp, it will work just fine.

### DIFF
--- a/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
+++ b/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
@@ -6,6 +6,7 @@ import ConfigParser
 import sys
 from datetime import datetime, timedelta
 import time
+import pytz
 from collections import defaultdict
 
 from OffsetStorage import FileOffsetStorage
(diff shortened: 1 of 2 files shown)


mgolosova:
[009: UTC timestamps](https://trello.com/c/LpphhsYi/91-009-utc-timestamps)

mgolosova:
[009: UTC timestamps](https://trello.com/c/LpphhsYi/91-009-utc-timestamps)

PR #244, opened 2019-04-26
[WIP] pyDKB storages

mgolosova:
This PR introduces to `pyDKB` library new submodule `storages`.

The module is intended to accumulate all the functionality related to the interaction with the storages we use, including both external and internal ones. It basically provides unified way to interact with any of them: object of `Storage` class, that understands basic set of methods, like:
* apply configuration (create client object(s));
* send `get()` request (get item by ID);
* send custom (parametric) query.

It is needed as we have at least 3 pieces of (Python) code that requires interaction with ES@DKB:
* [API server](https://github.com/PanDAWMS/dkb/blob/api/Utils/API/server/lib/dkb/api/storages/es.py);
* consistency check (#240);
* "safe" dataset metadata update ([91-ds-safe](https://github.com/PanDAWMS/dkb/tree/91-ds-safe)).

Also, almost same way as in the last case, we need to update Stage 95 (AMI-related stage).
To avoid copy-and-paste code cloning, this common module is developed.

---

ToDo / to think about:

* It is supposed to be used in API server as well (only that it would be better to have possibility to import this submodule without the rest of the library).
* Maybe it would be better to place it inside the `pyDKB.common` module.
* Stage 009 dbConnection also calls for the integration to this submodule.


mgolosova:
Force-push was just to fix codestyle in the original commits, instead of a separate 'Codestyle fix' commit.

PR #247, opened 2019-04-28
[obsolete] DF/data4es: 'update' mode for safe dataset metadata update.

mgolosova:
Closed for #253 + #262 + #263 in sum do the trick: we can run `data4es` with `-i 91_in,91_out,95` and be sure that data that already present in ES won't be spoiled.

#320 is the next step in this direction that will allow getting data from Rucio but in case of error fall back to the "update" scenario.

However it does not provide functionality like "query Rucio only if we don't have these data in ES". It'd be nice, but in fact it should be done by introducing another stage that would extract all available information from ES in the beginning of the `data4es` process, instead of asking every stage to query ES for its own data.

Original description
---

Applies functionality added in #245 and #246 to the `data4es` process, allowing to start this process in a normal (basic integration) and 'safe' (archive update) mode, which can be turned on with `--update` option.

[WIP] is due to the pyDKB-related changes: they clearly do not belong here. By the way, even after this change I have seen that `ConnectionTimeout` exception; but as we are talking about 'archived metadata update', it seems to me quite OK to be interrupted in case of overloaded ES and restart after an hour or so.

---

Waits for #245, #246.

### DIFF
--- a/Utils/Dataflow/pyDKB/storages/es.py
+++ b/Utils/Dataflow/pyDKB/storages/es.py
@@ -22,7 +22,8 @@ except ImportError:
 
 DEFAULT_CFG = {
     'host': '127.0.0.1',
-    'port': '9200'
+    'port': '9200',
+    'timeout_retry': 3
 }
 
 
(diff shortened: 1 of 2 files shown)


mgolosova:
Closed for #253 + #262 + #263 in sum do the trick: we can run data4es with -i 91_in,91_out,95 and be sure that data that already present in ES won't be spoiled.

#320 is the next step in this direction that will allow getting data from Rucio but in case of error fall back to the "update" scenario.

However it does not provide functionality like "query Rucio only if we don't have these data in ES". It'd be nice, but in fact it should be done by introducing another stage that would extract all available information from ES in the beginning of the data4es process, instead of asking every stage to query ES for its own data.

PR #249, opened 2019-05-16
PyDKB hdfs fix

Evildoor:
Fix some HDFSProducer-related errors discovered during tests.

### DIFF
--- a/Utils/Dataflow/pyDKB/common/hdfs.py
+++ b/Utils/Dataflow/pyDKB/common/hdfs.py
@@ -193,24 +193,24 @@ def listdir(dirname, mode='a'):
     return result
 
 
-def basename(path):
+def basename(p):
     """ Return file name without path. """
-    if path is None:
-        path = ''
-    return path.basename(path).strip()
+    if p is None:
+        p = ''
+    return path.basename(p).strip()
 
 
-def dirname(path):
+def dirname(p):
     """ Return dirname without filename. """
-    if path is None:
-        path = ''
-    return path.dirname(path).strip()
+    if p is None:
+        p = ''
+    return path.dirname(p).strip()
 
 
-def join(path, filename):
+def join(p, filename):
     """ Join path and filename. """
-    if path is None:
-        path = ''
+    if p is None:
+        p = ''
     if filename is None:
         filename = ''
-    return path.join(path, filename).strip()
+    return path.join(p, filename).strip()
--- a/Utils/Dataflow/pyDKB/common/hdfs.py
+++ b/Utils/Dataflow/pyDKB/common/hdfs.py
@@ -207,10 +207,12 @@ def dirname(p):
     return path.dirname(p).strip()
 
 
-def join(p, filename):
-    """ Join path and filename. """
-    if p is None:
-        p = ''
-    if filename is None:
-        filename = ''
-    return path.join(p, filename).strip()
+def join(p, *args):
+    """ Join given paths. """
+    paths = []
+    for a in [p] + list(args):
+        if a is None:
+            paths.append('')
+        else:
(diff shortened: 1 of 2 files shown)


mgolosova:
631adcb says:
> - The function was working incorrectly due to same name - path - used for
>  module and variable. However, this is obsolete after this commit.

It doesn\`t look very nice for at least two reasons:
* making `hdfs.join()` work with any number of args (the commit summary) by itself doesn\`t give (at least, me; I can\`t say for anyone else) any good reason to rename the first parameter, as it is related only to the second one;
* this commit mentions (and solves) some local problem, which is a part of a more common issue (you wrote about this `path`/`path` collision in e-mail and mentioned that there are a number of functions with this problem; and one can even see it (in the diff context): `return path.dirname(path).strip()` -- last line of another function).
I know it is not the case, but (if I haven\`t had that email) it looks like you haven\`t noticed other occurencies of this issue or simply decided to ignore it

Maybe the `path` -> `first_path` renaming was really just a part of adaptation of the function for multiple arguments; still the commit mentions it separately, and it makes reader to pay special attention and ask questions. So to avoid misunderstanding it would be better to either fix the common issue before the 631adcb, or do it right after it. After is a bit worse (as misunderstanding is already here), but would work if you have simply renamed the parameter and didn\`t notice the issue at all at that moment. Yet you did ;)

Evildoor:
My apologies - while I've seen the problems with the other functions and mentioned them in the email, I forgot about them while fixing hdfs.path(). Moved the renaming into a separate commit and fixed the other functions in it.

Utils/Dataflow/pyDKB/common/hdfs.py
```diff
@@ -207,10 +207,14 @@ def dirname(p):
...
         p = ''
-    if filename is None:
-        filename = ''
-    return path.join(p, filename).strip()
+    for a in args:
+        if a is None:
+            paths.append('')
+        else:
+            paths.append(a)
+    return path.join(p, *paths).strip()
```
mgolosova: And now I noticed: is there any need to leave `p` as a separate parameter? Looks like
```
join(*args):
  ...
  return path.join(*paths).strip()
```
will also work.

Evildoor: True. I've left the separate parameter to make the function similar to [os.path.join()](https://docs.python.org/2/library/os.path.html#os.path.join)'s definition, but it seems that it isn't required and we can simplify our code.

Implemented the change.

mgolosova: Er, I didn\`t notice this... In fact, it changes things: in `os.path.join()` (and in `posixpath.join()`, as well) it is not only about definition, it is about a required argument. So I wasn\`t exactly right: the version I have suggested is not really an equivalent to what it was (and it would be OK to point out this mistake ;) ).
So now we have a couple of options:
* rely on `posixpath.join()` exception;
* return empty string if no arguments passed;
* make the first argument required as it was (just handle it together with the rest of the arguments, like `for p in ([path] + args): ...`).

First variant is not very good, as on paper our function doesn\`t require arguments to be specified yet fails if there\`s none in deed. Which one of the rest of list is better I don\`t know: to follow the convenient of such functions or to return what user asks for (want to join nothing? Alright, get nothing).

Evildoor: Well, I picked this option:

> rely on posixpath.join() exception;

but you are right - if the function fails without arguments, this fact should be reflected somewhere, for example, in its definition or description. Done this instead:

> make the first argument required as it was (just handle it together with the rest of the arguments, like for p in ([path] + args): ...).

PR #251, opened 2019-05-30
DF/019: EOM handling improvement

mgolosova:
* fix input stream handling (no EOM was taken into account for the input stream);
* more strict output message definition (it must include last `\n` symbol as it is required by ES bulk API and must get there -- and if it is added as EOM, it will be removed on the next stage);
* new default EOM value (as `\n` appears in the message body, it can not be used as EOM);
* README updated accordingly.

### DIFF
--- a/Utils/Dataflow/019_esFormat/README
+++ b/Utils/Dataflow/019_esFormat/README
@@ -38,8 +38,8 @@ To produce regular (seamless NDJSON) samples, run:
 
 Comment:
  * input samples can not be used (and output one -- saved) as-is
-   due to the specific EOM marker (most of the stages use NEWLINE, which
-   is not allowed here because of the output message format);
+   due to the specific EOM marker (most of the stages use NEWLINE, which,
+   being part of the output messages body, is not allowed here);
  * `tr` is a Unix utility to translate or delete characters (for details please
    refer to man pages);
  * output message body contains trailing NEWLINE, so RS is simply removed from
--- a/Utils/Dataflow/019_esFormat/README
+++ b/Utils/Dataflow/019_esFormat/README
@@ -31,7 +31,21 @@ JSON documents, one document per line:
 
 Samples
 -------
-The 'output' directory contains two pairs of samples:
+To produce regular (seamless NDJSON) samples, run:
+
+  cat input/sample_name.ndjson | tr $'\n' $'\x1e' | \
+    ./run.sh -E '' | tr -d $'\x1e' > output/sample_name.ndjson
+
+Comment:
+ * input samples can not be used (and output one -- saved) as-is
+   due to the specific EOM marker (most of the stages use NEWLINE, which
+   is not allowed here because of the output message format);
+ * `tr` is a Unix utility to translate or delete characters (for details please
+   refer to man pages);
+ * output message body contains trailing NEWLINE, so RS is simply removed from
+   the output stream (not replaced with another NEWLINE).
+
+The 'output' directory contains two pairs of regular samples:
 
 2016 data:
 tasks2016.ndjson (from 016's 2016 sample)
--- a/Utils/Dataflow/019_esFormat/README
+++ b/Utils/Dataflow/019_esFormat/README
@@ -32,15 +32,6 @@ JSON documents, one document per line:
  ...
 }}}
 
-TODO
-----
-Currently, it operates as if one line was one message (piece of information).
-For input it is OK, but for output it would be better to take <index info> and
-  <data> JSON documents together, as one message. I don`t know where it will
-  be critical for now, but something tells me that it will.
-This task requires accurate EOMessage and EOProcess markers handling (see the
-  `kafka-dataflow-markers` branch).
-
 Samples
 -------
 The 'output' directory contains two pairs of samples:
--- a/Utils/Dataflow/019_esFormat/README
+++ b/Utils/Dataflow/019_esFormat/README
@@ -27,9 +27,6 @@ JSON documents, one document per line:
 {{{
  <index info>
  <data>
- <index info>
- <data>
- ...
 }}}
 
 Samples
--- a/Utils/Dataflow/019_esFormat/esFormat.php
+++ b/Utils/Dataflow/019_esFormat/esFormat.php
@@ -8,7 +8,7 @@ set_error_handler("exception_error_handler");
 $DEFAULT_INDEX = 'tasks_production';
 $ES_INDEX = NULL;
 $EOP_DEFAULTS = Array("stream" => chr(0), "file" => "");
-$EOM_DEFAULTS = Array("stream" => "\n", "file" => "\n");
+$EOM_DEFAULTS = Array("stream" => chr(30), "file" => chr(30));
 
 function check_input($row) {
   $required_fields = array('_id', '_type');
--- a/Utils/Dataflow/019_esFormat/esFormat.php
+++ b/Utils/Dataflow/019_esFormat/esFormat.php
@@ -133,7 +133,7 @@ if ($h) {
     $index = constructIndexJson($row);
 
     echo json_encode($index)."\n";
-    echo json_encode($row);
+    echo json_encode($row)."\n";
     echo $EOM_MARKER;
     echo $EOP_MARKER;
   }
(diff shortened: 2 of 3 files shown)


mgolosova:
[Stage 019: работа с EOM](https://trello.com/c/rvPcARIw/107-stage-019-%D1%80%D0%B0%D0%B1%D0%BE%D1%82%D0%B0-%D1%81-eom)

mgolosova:
Force-update is to fix mistake in the commit that changes default EOM: I forget to change it for file mode, though mentioned it in the commit log message (2bd522c -> 316a001).

mgolosova:
I don\`t know if I should update samples in `output/`: there must appear non-printable symbol RS (`\x1e`), which is very likely to make samples unreadable through GitHub interface (and not straightly usable as input for the next stage).

I think I can add "real" samples, marking already existed ones as "adapted" ones. Or just add a note in the README, saying that output samples can be obtained with provided command:
```
./run.sh -d s input/sample_name.ndjson | tr -d $'\x1e' > output/sample_name.ndjson
```

@Evildoor, recently you worked with samples a lot; what do you think?

Evildoor:
@mgolosova, I've looked at this, what can I say...
So far the situation in the repository is simple and straightforward: stage (any stage, not just this one), when given input sample, produces result identical to the corresponding output sample (with a caveat that some output may depend on configs, i.e. this stage takes index name from the config). This PR changes the situation - unless I'm missing something,

    ./run.sh input/sample_name.ndjson | tr -d $'\x1e' > output/sample_name.ndjson

will not work because the input samples (output of 016 and 095) are still using `\n` as EOM. The same thing will, as you correctly pointed out, happen with the output - 069 was not changed in this PR, so it will be unable to work with the new samples. [1]
I see several ways to proceed:

1. To uphold the "input produces output by default" rule it would be logical to change 016, 095 and 069's behavior to produce/accept messages with RS by default.  However, the stages use the same EOM for input and output messages, so it would require either changing ALL the stages to use RS (you said it would be possible and logical later, so doing it now may not be a good idea) or "breaking them in half" to be able to use `\n` for input (or output) and RS for the other channel (doesn't look good either, and may also create impression that, for example, stage 016's function is to do X and also change EOM, while in reality it just to do X, and EOM change is a crutch).
2. Refrain from using RS. However, you already explained why using `\n` would be incorrect, and any other value is no better than RS regarding other stages being unaware of it.
3. Somehow wrap or separate usage of the stage for real dataflow and for "input -> output" producing of samples. Writing an actual wrapper/different mode invoked by an argument/etc to run the stage with the different default value of a parameter looks like overkill, so your idea of adding a note about special way of getting the samples with `-e` looks appealing (with the correction that the input should be altered as well as the output, and a short explanation of the issue).
4. Do nothing and leave the current samples. Can be confusing.
5. Keep two versions of samples, as you've suggested. Can be confusing as well.

Neither of the solutions look very appealing to me because they either upset the existing "simple and straightforward" order of things or are unacceptable due to other reasons. Since the correct (in my opinion) solution - to change the whole dataflow to use RS as EOM - would be possible in the future, I think we should wait for it. In the meantime... Option 3, in the suggested form of addition to README, seems to be the best, but before that - is option 2 really that hopeless? I mean, if the case was simply about "stage that works fine in dataflow, but requires a bit of special treatment in testing environment" then it would be fine, but we are discussing possible changes to samples or their handling, and at the same time you had to change `data4es-start` to essentially make the same alterations to input and output of the stage in dataflow. Looks counter-intuitive. Maybe we can do something about addition of '\n' into the message?

[1] Actually, 069 in its current state won't give a damn about 019's output sample because, unlike other stages, it has a separate input sample instead of a link to 019's output. In this regard I'm talking about how things should be, and not how they are.

mgolosova:
@Evildoor,
> ```
> ./run.sh input/sample_name.ndjson | tr -d $'\x1e' > output/sample_name.ndjson
> ```
> will not work because the input samples (output of 016 and 095) are still using \n as EOM.

Fair point, then it must be something like this:
```
cat input/sample_name.ndjson | tr $'\n' $'\x1e' | ./run.sh | tr -d $'\x1e' > output/sample_name.ndjson
```

Not exactly readable, though...(

> The same thing will, as you correctly pointed out, happen with the output - 069 was not changed in this PR, so it will be unable to work with the new samples.

It is if we ***do*** place this uncommon EOM in the output sample; which is not necessary, if we decide to confine ourselves to the note in the README, that explains how to get this "purified" sample.

> ...changing ALL the stages to use RS (you said it would be possible and logical later)

I think I meant either that it can be done within `data4es-start` script, or that in case of Kafka supervisor it won\` be necessary. Yet I might have said something different indeed; if I did, please consider what is said above more relevant version.

> "breaking them in half" to be able to use `\n` for input (or output) and RS for the other channel

If doing something like this, then only on the global level; I mean, it will require dividing `EOM` in two (`EOM_input` and `EOM_output`) anywhere it appears in our code. Even if we may leave it be in `pyDKB` for now (saying that "for the time being it is OK to use same `EOM` for both streams"), the supervising program must be aware of it.
I don\`t say it is impossible; I just clarify that it must be implemented not as a "crutch", but as a regular solution. There are many versions of crutch solution, but this one seems to be abla to cause more problems then any other.

> Somehow wrap or separate usage of the stage for real dataflow and for "input -> output" producing of samples.

In context of #180 it even looks as a "regular" solution, and then README note may be considered as a "crutch solution", that One Day™ will be more formally implemented.
**So, this way it shall go**...

---

> Maybe we can do something about addition of '\n' into the message?

**Shortly**: yes, there is a way. But about a year ago it was decided not to go there.

**The long version.**
The only way to do it correctly is to state that `EOM` symbol, whatever it was, is allowed within the message body if properly escaped. It makes "read_message" operation a bit more complicated (not every `EOM` symbol is `EOM` indeed), and requires additional operation on every message: "code" it before sending ( == find and escape everything that may be mixed up with `EOM`) and "decode" on receiving. Every- and anywhere we work with messages (`pyDKB`, shell- and php- scripts, supervisors).

Escaping things is annoying: you can\`t simply say "we will replace NEWLINE symbols with literal `'\n'`"; you must ensure that all appearings of literal `'\n'` within the message body are escaped too; etc.
Remember regular expressions in Python, where you must type numerous '\' when need to compare with literal backslash symbol (unless r-strings are used):
```
>>> print "\\\\n"
\\n
>>> print r"\\n"
\\n
```

If we decide to do something like this, we have 4 extra operations at every stage; and it is exactly what we tried to avoid introducing custom `EOM`. 

And -- it is very likely that someone who tries to implement worker script at betch scale will forget about this (de)coding operations. While the whole idea was to make things simple: "output your message and add `EOM`; just don\`t forget to configure the supervisor to use the same `EOM`."
Everything else -- the command-line parameters, `pyDKB` wrappers around this and that -- is to avoid re-implementation of commonly used operations, manual adjustments of what we already have. And, of course, to bring some order: input/output related things have changed a couple of times, and we wanted to be sure that every stage is equally aware of this changes... and that no one steps into a trap that someone else has already detected and even found a way around.

---

*There\`s no such obstacle we couldn\`t possibly create.* ;)

mgolosova:
@Evildoor, please take a look: 8019cb2

Evildoor:
@mgolosova, thanks for the explanation. Addition to README looks good, I'm just going to suggest clarifying this point:

     * input samples can not be used (and output one -- saved) as-is
       due to the specific EOM marker (most of the stages use NEWLINE, which
       is not allowed here because of the output message format);

The last part - "because ... of the format" - sounds unclear to me, so I believe that something like

    ... use NEWLINE, which cannot be used in this stage because the output 
    messages it produces include the NEWLINE symbol);

would be easier to understand.

mgolosova:
@Evildoor, reworded now. Please take a look.

Evildoor:
@mgolosova, this looks better, but the order should probably be changed:

    most of the stages use NEWLINE, which, being part of the output messages' body, is not allowed here

But I'm not sure, maybe the current version is alright.

mgolosova:
@Evildoor,
both looks good enough for me, so changed to your version.

anastasiakaida:
@mgolosova 
I think it would be awesome to see a new output sample in this PR (even with artificial markers because of the difference between "Kafka-way" of communication and the current situation. :)

PR #254, opened 2019-06-03
DF/019: introduce 'update' action (ES updates, pt.1.1)

mgolosova:
Stage 019 now is able to produce records prepared to update data in ES via ES bulk API:
* from input records having special field `"_update": true`;
* from all input records, if run with `--update` command-line option.


Utils/Dataflow/019_esFormat/README
```diff
@@ -20,14 +23,16 @@ Required:
 
 Optional:
 * '_parent' (ES identifier for parent type record).
+* '_update' (corresponding ES record shouldbe "updated", not
```
Evildoor: Missing space: "should be".

Utils/Dataflow/019_esFormat/README
```diff
@@ -20,14 +23,16 @@ Required:
...
+             "inserted"/"indexed")
```
Evildoor: Missing "." at the end, to be in line with the previous element of the list.

Utils/Dataflow/019_esFormat/README
mgolosova: Fixed, thank you.

Utils/Dataflow/019_esFormat/README
mgolosova: Fixed, thank you.

mgolosova:
[Stage 019: разрешить операцию "update"](https://trello.com/c/coddXfAV/116-stage-019-%D1%80%D0%B0%D0%B7%D1%80%D0%B5%D1%88%D0%B8%D1%82%D1%8C-%D0%BE%D0%BF%D0%B5%D1%80%D0%B0%D1%86%D0%B8%D1%8E-update)

Evildoor:
Shouldn't this PR add new output samples (especially considering the fact that format changes a bit when "update" is used)? One the other hand, maybe the samples should be added in the future, when the whole "update" chain would be implemented (together with samples for other stages).

mgolosova:
@Evildoor, thank you for your comment, new sample was added.

PR #256, opened 2019-06-06
Stage 009: delay

Evildoor:
Add an option to get data in stage 009 with the final date set to relative 'N seconds/minutes/hours/... ago' - it is already possible, but only with N = 0. One of the goals of this change is to be able to have a delay between data loading and data consistency checking, to avoid discrepancies caused by constant updates of data in Oracle which are mirrored to ES only when another data loading is performed.

### DIFF
--- a/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
+++ b/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
@@ -325,7 +325,7 @@ def offset_now(tz=None, delay=None):
                time zone is used.
     :type tz: str, NoneType
     :param delay: delay in seconds. If not specified, default (globally set)
-               value is used.
+                  value is used.
     :type delay: int, NoneType
 
     :return: offset value (naive datetime, adjusted to OFFSET_TZ)
--- a/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
+++ b/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
@@ -143,8 +143,9 @@ def read_config(config_file):
     delay = config_get(config, 'timestamps', 'delay', '0')
     result['delay'] = interval_seconds(delay)
     if result['delay'] < 0:
-        # Delay is only used when 'final_date' is 'now'. It does not make much
-        # sense to ask for data from the future.
+        # Delay is used when right border of the data taking interval is
+        # 'now'.
+        # It does not make much sense to ask for data from the future.
         sys.stderr.write('(WARN) Delay is less than 0, setting it to 0.\n')
         result['delay'] = 0
     if result['step_seconds'] > 0:
--- a/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
+++ b/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
@@ -328,7 +328,8 @@ def offset_now(tz=None, delay=None):
                   value is used.
     :type delay: int, NoneType
 
-    :return: offset value (naive datetime, adjusted to OFFSET_TZ)
+    :return: offset value (naive datetime, adjusted to OFFSET_TZ with
+             OFFSET_DELAY)
     :rtype: datetime.datetime
     """
     TZ = OFFSET_TZ
--- a/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
+++ b/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
@@ -25,7 +25,7 @@ OUT = os.fdopen(sys.stdout.fileno(), 'w', 0)
 # (used to adjust 'now' to the proper value)
 OFFSET_TZ = None
 # Delay
-# (used to adjust 'now' when performing consistency check)
+# (used to adjust 'now' to avoid pulling of "unstable" data)
 OFFSET_DELAY = None
 # ---
 
--- a/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
+++ b/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
@@ -24,6 +24,9 @@ OUT = os.fdopen(sys.stdout.fileno(), 'w', 0)
 # Timezone of the data timestamps in the source DB
 # (used to adjust 'now' to the proper value)
 OFFSET_TZ = None
+# Delay
+# (used to adjust 'now' when performing consistency check)
+OFFSET_DELAY = None
 # ---
 
 
@@ -53,6 +56,8 @@ def main():
 
     # Offset timezone initialization
     init_offset_tz(config)
+    # Offset delay initialization
+    init_offset_delay(config)
 
     conn = OracleConnection(config['__dsn'])
     if not conn.establish():
@@ -135,11 +140,18 @@ def read_config(config_file):
     # Optional parameters
     result['timestamp_tz'] = config_get(config, 'timestamps', 'tz',
                                         time.tzname[0])
+    delay = config_get(config, 'timestamps', 'delay', '0')
+    result['delay'] = interval_seconds(delay)
+    if result['delay'] < 0:
+        # Delay is only used when 'final_date' is 'now'. It does not make much
+        # sense to ask for data from the future.
+        sys.stderr.write('(WARN) Delay is less than 0, setting it to 0.\n')
+        result['delay'] = 0
     if result['step_seconds'] > 0:
         default_initial = datetime.utcfromtimestamp(0)
         default_final = None
     else:
-        default_initial = offset_now(result['timestamp_tz'])
+        default_initial = offset_now(result['timestamp_tz'], result['delay'])
         default_final = datetime.utcfromtimestamp(0)
 
     result['final_date'] = str2date(config_get(config, 'timestamps', 'final',
@@ -226,6 +238,12 @@ def init_offset_tz(config):
     OFFSET_TZ = pytz.timezone(config['timestamp_tz'])
 
 
+def init_offset_delay(config):
+    """ Initialize OFFSET_DELAY. """
+    global OFFSET_DELAY
+    OFFSET_DELAY = config['delay']
+
+
 def init_offset_storage(config):
     """ Get configured (or default) offset file.
 
@@ -299,12 +317,15 @@ def get_offset(offset_storage):
     return result
 
 
-def offset_now(tz=None):
+def offset_now(tz=None, delay=None):
     """ Get offset value corresonding to the current moment of time.
 
     :param tz: time zone name. If not specified, default (globally set)
                time zone is used.
     :type tz: str, NoneType
+    :param delay: delay in seconds. If not specified, default (globally set)
+               value is used.
+    :type delay: int, NoneType
 
     :return: offset value (naive datetime, adjusted to OFFSET_TZ)
     :rtype: datetime.datetime
@@ -312,7 +333,10 @@ def offset_now(tz=None):
     TZ = OFFSET_TZ
     if tz:
         TZ = pytz.timezone(tz)
-    return datetime.now(TZ).replace(tzinfo=None)
+    dl = OFFSET_DELAY
+    if delay:
+        dl = delay
+    return datetime.now(TZ).replace(tzinfo=None) - timedelta(seconds=dl)
 
 
 def plain(conn, queries, start_date, end_date):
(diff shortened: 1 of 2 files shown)


Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -476,9 +483,10 @@ def process(conn, offset_storage, config):
     offset_date = get_offset(offset_storage)
     if not reverse and not final_date:
         # In case of normal data load 'final_date' may be None:
-        # it means we need to adjust it to current timestamp
-        # before every check
-        final_date = offset_now()
+        # it means we need to set it to current timestamp
+        # before every check. 'delay' can be used to adjust the resulting
+        # final_date in this case.
+        final_date = offset_now() - timedelta(seconds=config['delay'])
```
mgolosova: Seems like same operation should be done everywhere `offset_now()` is used.
E.g. here, as "right" border of time interval is adjusted within the loop:
https://github.com/PanDAWMS/dkb/blob/715d86ff923058ea5ae9605f823b806988d709ac/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py#L518-L519

Maybe it should be done in within the function, just like time zone is taken into account only there? It would require global variable, of course (and there\`s nothing good in it), yet will allow to "automatically" take the delay into account whenever we need "current timestamp".

Utils/Dataflow/config/009.cfg.example
```diff
@@ -32,9 +32,17 @@ mode = SQUASH
 # initial timestamp (e.g. 01-05-2016 00:00:00)
 initial = %__009_init_offset__%
 # final timestamp (e.g. 10-05-2016 00:00:00)
+# If this parameter is not set, 'now' (current moment of time) will be used.
```
mgolosova: This comment is out of the commit scope, and not exactly correct: if we go "backwards" in time, it will be set to `1970-01-01 00:00:00 (UTC)`:
https://github.com/PanDAWMS/dkb/blob/715d86ff923058ea5ae9605f823b806988d709ac/Utils/Dataflow/009_oracleConnector/Oracle2JSON.py#L138-L143

Utils/Dataflow/config/009.cfg.example
```diff
@@ -32,9 +32,17 @@ mode = SQUASH
...
 final = %__009_final_offset__%
 # offset from current timestamp (e.g. 86400 <=> 86400s <=> 1d; 8h; 15m)
 step = %__009_step__%
+# delay (e.g. 86400 <=> 86400s <=> 1d; 8h; 15m)
+# If 'final' is not set, this parameter can be used to adjust 'now'.
+# For example, running the script at 10-05-2016 12:00:00 with
+# delay = 2h will set 'final' to 10-05-2016 10:00:00.
+# Has no effect if 'final' is set explicitly.
+# Cannot be negative.
```
mgolosova: I was tending to type `delay: -10m`, indeed :)

Maybe it would be OK to allow negatives, but use `abs(delay)`? Or it will be confusing?

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
Evildoor: I agree with

> Seems like same operation should be done everywhere offset_now() is used.
> ...
> Maybe it should be done in within the function, just like time zone is taken into account only there?

, but do we really need global variable? Delay is used for a single purpose, adjusting the "right" boundary of the time interval. At the moment it is done 3 times (via `offset_now()`), once in `read_config()` and twice in `process()`, both functions have access to config and delay. Maybe making `offset_now()` accept delay as a parameter will be enough? Or it is highly possible in the future to encounter a situation where `offset_now()` must be called, but config is unavailable?

Utils/Dataflow/config/009.cfg.example
Evildoor: Removed the line.
While I believe that it's not out of the scope by itself (because we are adding `delay`, which has to be described, which includes the fact that it works only if `final` is not set, which invites the mentioning of stage's behavior when `final` is not set in general), you correctly point out that it works differently in a particular case. This means that this case should be described as well, and returns us to an issue already raised at some point - all the nuances of time-boundaries-related work of this stage should be documented somewhere, and doing so is definitely out of scope of the commit, and the PR.

Utils/Dataflow/config/009.cfg.example
Evildoor: Current version is simple and straightforward, while allowed negatives and `abs()` may complicate the situation a bit. I doubt that they'll be too confusing, but I'm not sure if we need it.
Another reason not to use the negative values is speed - as someone (here, in KI, or maybe even DKB) educated me at some point, strings such as `name-of-something` are considered to be better than `name_of_something` (which, for some reason, I preferred to use) because `-` is slightly faster to type than `_`. Same, I believe, goes for `10m` vs `-10m`.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
mgolosova: > Maybe making `offset_now()` accept delay as a parameter will be enough? Or it is highly possible in the future to encounter a situation where `offset_now()` must be called, but config is unavailable?

I\`d say that it is much more possible that even when `config` is available, one can simply forget to pass `delay` (if, of course, `offset_now()` does not require both parameters to be excplicitly passed.

My point is that `delay` must affect the behaviour of the whole stage: whenever it uses current timestamp, it must be adjusted with `delay`.

Actually, the better way would be to wrap all the stage this to some kind of object, where `config` would be nothing but an instance variable. It is way too much, of course ;) But still made me think that maybe instead of global `OFFSET_TZ` and possible `DELAY` we might want to switch to global `CONFIG` (it might, by the way, had already been global once, but turned into local as superfluous).

Utils/Dataflow/config/009.cfg.example
mgolosova: Added card for this issue to not forget after the conversation is resolved: https://trello.com/c/YhPuVdIk

Utils/Dataflow/config/009.cfg.example
mgolosova: > strings such as `name-of-something` are considered to be better than `name_of_something` (which, for some reason, I preferred to use) because `-` is slightly faster/easier to type than `_`.
Same, I believe, goes for `10m` vs `-10m`.

I think that statement about `name-of-something` was related to the something that was supposed to be typed every now and then (filename, branch name, e-mail address, ...). Parameters in config files are not to write, but to read them fast ;)

But strictly positive delay is also OK, so let it be so.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
Evildoor: Fair enough. Made adjustments to the first commit and force-pushed the branch. Please, take a look.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -24,6 +24,9 @@
 # Timezone of the data timestamps in the source DB
 # (used to adjust 'now' to the proper value)
 OFFSET_TZ = None
+# Delay
+# (used to adjust 'now' when performing consistency check)
```
mgolosova:
```suggestion
# (used to adjust 'now' to avoid pulling of "unstable" data)
```

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -53,6 +56,8 @@ def main():
 
     # Offset timezone initialization
     init_offset_tz(config)
+    # Offset delay initialization
+    init_offset_delay(config)
```
mgolosova: May it be `init_offset_params(config)` instead of these two calls?

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -135,11 +140,18 @@ def read_config(config_file):
     # Optional parameters
     result['timestamp_tz'] = config_get(config, 'timestamps', 'tz',
                                         time.tzname[0])
+    delay = config_get(config, 'timestamps', 'delay', '0')
+    result['delay'] = interval_seconds(delay)
+    if result['delay'] < 0:
+        # Delay is only used when 'final_date' is 'now'. It does not make much
```
mgolosova:
```suggestion
        # Delay is used when right border of the data taking interval is 'now'.
```

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -135,11 +140,18 @@ def read_config(config_file):
...
+        # sense to ask for data from the future.
```
mgolosova:
```suggestion
        # It does not make much sense to ask for data from the future.
```

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -299,20 +317,26 @@ def get_offset(offset_storage):
...
 
-def offset_now(tz=None):
+def offset_now(tz=None, delay=None):
     """ Get offset value corresonding to the current moment of time.
 
     :param tz: time zone name. If not specified, default (globally set)
                time zone is used.
     :type tz: str, NoneType
+    :param delay: delay in seconds. If not specified, default (globally set)
+               value is used.
```
mgolosova:
```suggestion
                  value is used.
```

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
```diff
@@ -299,20 +317,26 @@ def get_offset(offset_storage):
...
+    :type delay: int, NoneType
 
     :return: offset value (naive datetime, adjusted to OFFSET_TZ)
```
mgolosova: `...with OFFSET_DELAY)` ?

Utils/Dataflow/config/009.cfg.example
```diff
@@ -35,6 +35,13 @@ initial = %__009_init_offset__%
 final = %__009_final_offset__%
 # offset from current timestamp (e.g. 86400 <=> 86400s <=> 1d; 8h; 15m)
 step = %__009_step__%
+# delay (e.g. 86400 <=> 86400s <=> 1d; 8h; 15m)
+# This parameter can be used to adjust 'now'.
+# For example, running the script at 10-05-2016 12:00:00 with
+# delay = 2h will set 'final' to 10-05-2016 10:00:00.
+# Has no effect if 'final' is set explicitly.
```
mgolosova:
```suggestion
# Has no effect if the right border of the data taking interval
# is set explicitly.
```

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
Evildoor: Done.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
Evildoor: Done.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
Evildoor: Done.

Utils/Dataflow/config/009.cfg.example
Evildoor: Done.

Utils/Dataflow/009_oracleConnector/Oracle2JSON.py
Evildoor: I thought about this before making the changes, but decided against it. Maybe in the future we will actually implement the suggestion described above, about moving the whole config into global variable, but I think we should start doing it when we make the decision, not now.

mgolosova:
[Stage 9: задержка относительно «сейчас»](https://trello.com/c/sPwSvMvs/58-stage-9-%D0%B7%D0%B0%D0%B4%D0%B5%D1%80%D0%B6%D0%BA%D0%B0-%D0%BE%D1%82%D0%BD%D0%BE%D1%81%D0%B8%D1%82%D0%B5%D0%BB%D1%8C%D0%BD%D0%BE-%D1%81%D0%B5%D0%B9%D1%87%D0%B0%D1%81)

PR #260, opened 2019-06-13
pyDKB: processor `--skip` option (ES updates, pt.2.1).

mgolosova:
New command line optional parameter added:
```
  --skip                Skip process and push input message forward
                        as-is (marking it as "incomplete").
```
Every `ProcessorStage` has a default implementation of the "skip"
procedure (`skip_process()` method):
* get content of input message;
* create output message of a proper type;
* mark it as 'incomplete';
* write it to the output buffer.

This method can be re-implemented for some specific use-case, if needed.

The procedure for marking message as "incomplete" depends on the message
type:
* JSON: set special field "_incomplete" to True;
* TTL: do nothing.

To test the added functionality, new test stage was added:
  /Utils/Dataflow/test/pyDKB/json2JSON.py

---

ToDo:
 - [x] merge `pyDKB-processor-skip` to `pyDKB` after approvement.

PR #261, opened 2019-06-14
DF/025: skip init (ES updates, pt.2.2a)

mgolosova:
If we are going to skip processing, we don't need to initialize the
connection to external source at all. So the connection will be
established not earlier than we start processing of the first input
message in a normal mode.

And if the external source is unavailable for some reason, processing still will be stopped after the unsuccessful try to initialise the client.

For now, even if the stage is failed and stopped, the whole processing chain stops only when the previous stage tries to write something to STDOUT and realise that no one is listening. Comparing to this scenario, now the chain will be intrrupted just slightly later.

---

~Waits for #260.~ (merged)


mgolosova:
[Stage 025,091,095: адаптировать к возможности использования опции `--skip`](https://trello.com/c/n7kFCxam/118-stage-025091095-%D0%B0%D0%B4%D0%B0%D0%BF%D1%82%D0%B8%D1%80%D0%BE%D0%B2%D0%B0%D1%82%D1%8C-%D0%BA-%D0%B2%D0%BE%D0%B7%D0%BC%D0%BE%D0%B6%D0%BD%D0%BE%D1%81%D1%82%D0%B8-%D0%B8%D1%81%D0%BF%D0%BE%D0%BB%D1%8C%D0%B7%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D1%8F-%D0%BE%D0%BF%D1%86%D0%B8%D0%B8-skip)

anastasiakaida:
@mgolosova 
Please, change word 'silightly' to 'slightly' in 8bdfd81, there is a typo. :)

Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
```diff
@@ -280,6 +291,24 @@ def process(stage, input_message):
         """
         raise NotImplementedError("Stage method process() is not implemented")
 
+    @staticmethod
+    def skip_process(stage, input_message):
+        """ Mark input message as incomplete and output.
+
+        May be re-implemented just like `process()`,
+        yet has default implememtation.
```
Evildoor: ImplemeNtation.

mgolosova:
@anastasiakaida,
> Please, change the word 'silightly' to 'slightly' in 8bdfd81, there is a typo. :)

Sorry, I saw your comment too late. This commit belongs to #260, which was already approved and just merged...

Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
mgolosova: This change belongs to #260, which was already approved and just merged. :(
Will fix in a separate PR.

mgolosova: Update: done (#266)

PR #262, opened 2019-06-14
DF/091: skip mode (ES updates, pt.2.2b)

mgolosova:
If we are going to skip processing, we don't need to initialize the
connection to external source at all. So the connection will be
established not earlier than we start processing of the first input
message in a normal mode.

And if the external source is unavailable for some reason, processing still will be stopped after the unsuccessful try to initialise the client.

For now, even if the stage is failed and stopped, the whole processing chain stops only when the previous stage tries to write something to STDOUT and realise that no one is listening. Comparing to this scenario, now the chain will be intrrupted just slightly later.

---

Waits for: #260


mgolosova:
[Stage 025,091,095: адаптировать к возможности использования опции `--skip`](https://trello.com/c/n7kFCxam/118-stage-025091095-%D0%B0%D0%B4%D0%B0%D0%BF%D1%82%D0%B8%D1%80%D0%BE%D0%B2%D0%B0%D1%82%D1%8C-%D0%BA-%D0%B2%D0%BE%D0%B7%D0%BC%D0%BE%D0%B6%D0%BD%D0%BE%D1%81%D1%82%D0%B8-%D0%B8%D1%81%D0%BF%D0%BE%D0%BB%D1%8C%D0%B7%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D1%8F-%D0%BE%D0%BF%D1%86%D0%B8%D0%B8-skip)

Utils/Dataflow/091_datasetsRucio/datasets_processing.py
```diff
@@ -133,6 +153,41 @@ def process_output_ds(stage, message):
...
+def skip_process_output_ds(stage, message):
+    """ Process output datasets from input message.
+
+    Generate output JSON document of the following structure:
+        { "datasetname": <DSNAME>,
+          "_type": "output_dataset",
+          "_parent": <TASKID>,
+          "_id": <DSNAME>
+        }
+    """
```
Evildoor: I think this description, and maybe even function name, should be changed.

At the moment the description's text part is identical to `process_output_ds`, and only the JSON examples differ. Since I know a few things about this PR and the stage's logic, I can look at the code and the examples, and understand that `skip_process_output_ds()`, unlike its "real" counterpart, skips the data retrieval from Rucio and instead simply defines `datasetname` and other crucial things that are required and that can be defined without Rucio. However, it takes some effort, and it would be even harder for someone less familiar with the situation - I believe that changing the text part to something like "process A but skip complex B and do simple C instead (optional: and all this is done because of D)" will make the the code easier to read.

The function name looks a bit wrong because, even if the description is changed to something like suggested above, the name says "skip process", but the description and logic are "process, but with simplifications".

mgolosova: `skip_process_*()`, just like `process_*`, comes from `ProcessorStage.skip_process()` method name.
It can be changed, of course, but I can\`t invent more "self-explaining" name for it, just as I couldn\`t earlier.
I mean, for me the name means not "to skip processing of output dataset", but "the skip processing for output dataset".

Will it look better if I change the description to something like this (see below)?
```
def skip_process_output_ds(stage, message):
    """ Implementation of ProcessorStage.skip_process() method.

    Convert input message, representing task, into set of messages,
    representing task output datasets.
    Each output message contains dataset UID (name) and service fields:
        { "datasetname": <DSNAME>,
          "_type": "output_dataset",
          "_parent": <TASKID>,
          "_id": <DSNAME>
        }
    """
```

Evildoor: > It can be changed, of course, but I can`t invent more "self-explaining" name for it, just as I couldn`t earlier.

Yes, I thought of this as well... The problem is, the idea was to "skip processing" meaning "take the input message, mark it as incomplete, and pass further without changing", but it turns out that sometimes "without changing" is not exactly true...

> Will it look better if I change the description to something like this (see below)?

Looks better to me, I think we should implement this, and then leave things as-is, for now. Maybe the problems with "skip but actually not exactly skip" should be addressed somewhere, like pyDKB readme or ProcessorStage.skip_process()'s description, but this PR is no place for that.

mgolosova: > The problem is, the idea was to "skip processing" meaning "take the input message, mark it as incomplete, and pass further without changing", but it turns out that sometimes "without changing" is not exactly true...
> Maybe the problems with "skip but actually not exactly skip" should be addressed somewhere...

https://trello.com/c/A1piDcdN, yeah ;) This "sometimes" should never happen, but it does.

> ...like pyDKB readme or ProcessorStage.skip_process()'s description, but this PR is no place for that.

It is, in fact: as one is allowed (and welcomed) to re-implement `skip_process()`, it must invite the assumption that meaning of "skip message" may differ from stage to stage. But it can be done more clearly, you\`re right: the summary description of the `ProcessorStage.skip_process()` must more generally state that it is the method that will be used if `--skip` option is specified.

Will take a look.

mgolosova: > Will take a look

Done: #269

mgolosova: >> Will it look better if I change the description to something like this (see below)?
>
> Looks better to me, I think we should implement this

Done.

PR #264, opened 2019-06-18
DF/019: "incomplete" messages handling (ES updates, pt.1.2).

mgolosova:
Information about message (in)completeness is, basically, some service information for this stage.
Currently it was decided to simply keep in the ES some service field that will allow to mark "incomplete" documents; later we will use it to estimate, how many records are affected on a regular basis, and decide what to do: allow "update" operation when we know we don't have the full document, but still want new information in the storage, or something else.

So Stage 19 now turns `_incomplete` field of the input message to the `update_required` ES field, and tries not to do something stupid about it (e.g. not to mark all record as "incomplete" when all we wanted was to update information from one specific external source).

ES mapping is updated accordingly.

**Question**: shouldn\`t the ES service field also start with underscore, like `"_update_required"`? This field should not be used by any of the existing stages, so it won\`t be erroneously ignored as a *message* service field; and it is obviously not a "data field", so if we take some data from ES, we most likely will ignore it (unless we want to get all the records that require update). Not sure about this point.

Points to consider:
 - [x] the **question** above *(discussed at Thu meetings; ES seems to allow such naming for fields)*;
 - [ ] before we can use `_incomplete` as a signal to `update` action, all the stages must leave undefined fields empty (for the fields that must  survive in ES after update).

---

~(Waits for: #254.)~ *(merged)*

### DIFF
--- a/Utils/Dataflow/019_esFormat/esFormat.php
+++ b/Utils/Dataflow/019_esFormat/esFormat.php
@@ -95,8 +95,8 @@ function constructDataJson($row) {
   if (isset($incompl)) {
 
     # "_update_required" field must be specified explicitly in two cases:
-    #  * message is incomplete (so that record in ES get properly marked);
-    #  * we perfom "update" operation (since we want to use 'doc_as_upsert' ES
+    #  * message is incomplete (so that record in ES gets properly marked);
+    #  * we perform "update" operation (since we want to use 'doc_as_upsert' ES
     #    option, if possible, using "insert_data" for both cases --
     #    insert-if-missed and existing document update -- and still get
     #    resulting document properly marked)
--- a/Utils/Dataflow/019_esFormat/esFormat.php
+++ b/Utils/Dataflow/019_esFormat/esFormat.php
@@ -80,18 +80,46 @@ function constructActionJson($row) {
 function constructDataJson($row) {
   $data = $row;
 
+  if (isset($data['_incomplete'])) {
+    $incompl = $data['_incomplete'];
+  }
+
   foreach ($data as $key => $val) {
     if (strncmp($key, '_', 1) === 0)
       unset($data[$key]);
   }
 
   $act = getAction($row);
+  $insert_data = $data;
+  $update_data = $data;
+  if (isset($incompl)) {
+    # We *may* not specify "update_required" value on insert, if it is False,
+    # (for it is a default value), but to allow usage of 'doc_as_upsert'
+    # ES option, it can be specified explicitly. So in case of 'update' action
+    # it should always be so.
+    if ($incompl or $act == 'update') {
+      $insert_data['update_required'] = $incompl;
+    }
+    if (!$incompl) {
+      # We must specify explicitly that update is not required after this operation
+      # (just in case it was set to "true" previously).
+      $update_data['update_required'] = false;
+    }
+  }
 
   if ($act == 'update') {
+    # We can do "clean upsert" (when insert and update documents are the same)
+    # only when $incompl is NOT set to True:
+    $clean_upsert = !(isset($incompl) and $incompl);
     $data = Array(
-      'doc' => $data,
-      'doc_as_upsert' => true
+      'doc' => $update_data,
+      'doc_as_upsert' => $clean_upsert
     );
+    if (!$clean_upsert) {
+      $data['upsert'] = $insert_data;
+    }
+  } else {
+    $data = $insert_data;
   }
 
   return $data;
--- a/Utils/Dataflow/019_esFormat/esFormat.php
+++ b/Utils/Dataflow/019_esFormat/esFormat.php
@@ -93,13 +93,17 @@ function constructDataJson($row) {
   $insert_data = $data;
   $update_data = $data;
   if (isset($incompl)) {
-    # We *may* not specify "_update_required" value on insert, if it is False,
-    # (for it is a default value), but to allow usage of 'doc_as_upsert'
-    # ES option, it can be specified explicitly. So in case of 'update' action
-    # it should always be so.
+
+    # "_update_required" field must be specified explicitly in two cases:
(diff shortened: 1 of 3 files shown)


mgolosova:
[Stage 019: работа с "неполными" сообщениями](https://trello.com/c/DEABmAcg/119-stage-019-%D1%80%D0%B0%D0%B1%D0%BE%D1%82%D0%B0-%D1%81-%D0%BD%D0%B5%D0%BF%D0%BE%D0%BB%D0%BD%D1%8B%D0%BC%D0%B8-%D1%81%D0%BE%D0%BE%D0%B1%D1%89%D0%B5%D0%BD%D0%B8%D1%8F%D0%BC%D0%B8)

Utils/Elasticsearch/mapping/tasks.mapping
```diff
@@ -236,6 +236,9 @@
         },
         "n_files_to_be_used": {
           "type": "long"
+        },
+        "update_required": {  /* service field to mark  documents that contain incomplete */
```
Evildoor: Extra space: `mark__documents`.

mgolosova: [Fixed](https://github.com/PanDAWMS/dkb/compare/4afb4dd9b8c311b30a59fab1b0dfe7e221498f37..a2e2cc48dd6f04744d458af5dafede69b67ce57d), thank you!

Evildoor:
a2e2cc4: maybe place the comments into README instead of mapping, for the same reason it's done in 9a8ffd3?

mgolosova:
@Evildoor,
> a2e2cc4: maybe place the comments into README instead of mapping, for the same reason it's done in 9a8ffd3?

Good point. On this occasion, rebased the branch on the current version of `master`.

Utils/Elasticsearch/mapping/README
```diff
@@ -17,3 +17,6 @@ toths06_failed: 'wasted' CPU resources (see description for `toths06`).
 input_events: derived value based on primary_input_events, requested_events,
               n_files_per_job, n_events_per_job, n_files_to_be_used.
 
+_update_required: service field to mark documents that contain incomplete
+                  information about object and thus must be updated sooner
+                  or later
```
Evildoor:
```suggestion
                  or later.
```

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -80,18 +80,46 @@ function constructActionJson($row) {
...
   }
 
   $act = getAction($row);
+  $insert_data = $data;
+  $update_data = $data;
+  if (isset($incompl)) {
+    # We *may* not specify "_update_required" value on insert, if it is False,
+    # (for it is a default value), but to allow usage of 'doc_as_upsert'
+    # ES option, it can be specified explicitly. So in case of 'update' action
+    # it should always be so.
```
Evildoor: This comment is confusing. I've read d44ec42's description and the code block, but still have hard time understanding the comment. How about:

    There are four possible situations depending on the combination
    of two values: action (insert or update) and data incompleteness = "_update_required" (True or False).
    While there is no need to explicitly set "_update_required" value to False
    (for it is a default value), doing so in case of update will allow usage of 'doc_as_upsert'
    (update that adds new, complete document that was missing).
    This means that in three out of four cases - insert of incomplete data
    and update of data of any kind - the value must be set explicitly.

Correct me if I'm wrong.
P.S. Suggested comment, if accepted, should be checked for line length.

mgolosova: @Evildoor 
> ```
> ...usage of 'doc_as_upsert'
> (update that adds new, complete document that was missing).
> ```

1. In fact, `doc_as_upsert` allows us to send one document (for update of an existing one and to be inserted if document does not exist) instead of two (one for update, another -- for insert), so I would say that the suggested version is incorrect or at least inaccurate.

2. I don\`t think we need to talk here about all four scenarios we have: this comment is related only to the statement right after it, which says "set the `_update_required` field explicitly, if action is 'update' or message is incomplete".

3. I still will rewrite this part, since it is obviously not very clear :)

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -93,13 +93,17 @@ function constructDataJson($row) {
   $insert_data = $data;
   $update_data = $data;
   if (isset($incompl)) {
-    # We *may* not specify "_update_required" value on insert, if it is False,
-    # (for it is a default value), but to allow usage of 'doc_as_upsert'
-    # ES option, it can be specified explicitly. So in case of 'update' action
-    # it should always be so.
+
+    # "_update_required" field must be specified explicitly in two cases:
+    #  * message is incomplete (so that record in ES get properly marked);
```
Evildoor: "record ... getS ... marked"?

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -93,13 +93,17 @@ function constructDataJson($row) {
...
+    #  * we perfom "update" operation (since we want to use 'doc_as_upsert' ES
```
Evildoor: "perfoRm".

Utils/Dataflow/019_esFormat/esFormat.php
Evildoor: Aha, I see. New variant looks fine to me, apart from some language-related nuances (commented below).

Utils/Dataflow/019_esFormat/esFormat.php
mgolosova: Fixed: cdfd67a

Utils/Dataflow/019_esFormat/esFormat.php
mgolosova: Fixed: cdfd67a

PR #267, opened 2019-06-19
Travis checks for README files

IATkachenko:
To check changes in README files by travis-ci added set of scripts to Tests/travis and updated travis configuration.

All checks goes in own check phase.

Fixes: #268

### DIFF
--- a/Tests/travis/README_tabs.sh
+++ b/Tests/travis/README_tabs.sh
@@ -1,0 +1,21 @@
+#!/bin/bash
+functions="$(dirname $0)/functions.sh"
+if [ -f "$functions" ]; then
+  . "$functions"
+else
+  echo "functions file '$functions' not found!"
+  exit 1
+fi
+
+e=0
+while read f; do
+  [ -z "$f" ] && continue
+  r=`getAddedLines "$f" | grep -P '^.*?\t+.*'`;
+  if [ $? -eq 0 ];then
+    echo "Tabs found in $f:"
+    getLinesFromFile "$f" "$r"
+    e=1
+  fi
+done <<< $(getChangedFiles | grep 'README$')
+exit $e
+
--- a/Tests/travis/README_tabs.sh
+++ b/Tests/travis/README_tabs.sh
@@ -10,7 +10,7 @@ fi
 e=0
 while read f; do
   [ -z "$f" ] && continue
-  r=`getAddedLines "$f" | grep -P '^.*?\t+.*'`;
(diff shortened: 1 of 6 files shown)


Tests/travis/README_tabs.sh
```diff
@@ -0,0 +1,21 @@
...
+  . "$functions"
+else
+  echo "functions file '$functions' not found!"
+  exit 1
+fi
+
+e=0
+while read f; do
+  [ -z "$f" ] && continue
+  r=`getAddedLines "$f" | grep -P '^.*?\t+.*'`;
```
mgolosova: Won\`t `grep -P '\t'` work just as well?

Tests/travis/README_trailing_spaces.sh
```diff
@@ -0,0 +1,21 @@
...
+  . "$functions"
+else
+  echo "functions file '$functions' not found!"
+  exit 1
+fi
+
+e=0
+while read f; do
+  [ -z "$f" ] && continue
+  r=`getAddedLines "$f" | grep -e '^.*\s\+$'`;
```
mgolosova: Won\`t `grep -e '\s$'` work?
I mean, single trailing space must already be enough to say that the line is corrupted.

Tests/travis/README_tabs.sh
IATkachenko: Thank you, regexp was simplified.

Tests/travis/README_trailing_spaces.sh
IATkachenko: Thank you, regexp was simplified.

Tests/travis/functions.sh
```diff
@@ -0,0 +1,23 @@
...
+  git diff HEAD^ -- "$file" | grep -v "^+++ b/$file" | grep -e '^+' | sed -e 's/^+//'
+}
+
+getLinesFromFile() {
+  file="$1"
+  lines="$2"
+
+  echo "$lines" | sort -u | while IFS= read -r l; do
+      grep -n "^$l\$" -- "$file"
+  done
```
mgolosova: `sort -u` is to be sure that every line will be found only once, right?
Won\`t it make `grep` output out of order? Maybe sort and make unique the result (line numbers), not lines themselves?

*(Didn\`t try it yet, so just a thought)*

IATkachenko: If we don't make line unique, then grep will do work twice.
I'll add ordering by line number into output.

IATkachenko:
Commit logs will be reworded.

PR #272, opened 2019-06-25
DF/009: fix issue with zero value of the 'delay' parameter.

mgolosova:


Evildoor:
While I agree with the change - there is a difference between None and 0 - while looking at the code I've found out that an exception is thrown when delay is None:

    >>> offset_now()
      File "Oracle2JSON.py", line 341, in offset_now
        return datetime.now(TZ).replace(tzinfo=None) - timedelta(seconds=dl)
    TypeError: unsupported type for timedelta seconds component: NoneType

Shouldn't the default value OFFSET_DELAY be changed to 0?

mgolosova:
@Evildoor, fair point, thank you. Fixed now.

PR #274, opened 2019-07-03
pyDKB: improve logging system.

mgolosova:
While #135 is an attempt to do something more elegant, this PR simpy makes the code more DRY and improves log messages style.

---

~(Waiting for: #273)~ *(merged)*


Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -0,0 +1,48 @@
...
+    :param message: message to output
+    :type message: str
+    :param level: log level of the message
+    :type level: ``pyDKB.common.types.logLevel`` member
+    :param *args: additional prefixes (will be output between log
+                  level prefix and message body)
+    :type *args: str
+    """
+    if not logLevel.hasMember(level):
+        self.log("Unknown log level: %s" % level, logLevel.WARN)
```
Evildoor: It's a function, not class method, so `self` makes no sense (and produces an error).

Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -0,0 +1,48 @@
...
+    if type(message) == list:
+        lines = message
+    else:
+        lines = message.splitlines()
+    if args:
+        prefix = ' ' + ' '.join(['(%s)' % p for p in args])
+    else:
+        frm = inspect.stack()[1]
+        mod = inspect.getmodule(frm[0])
+        prefix = ' (%s)' % mod.__name__
```
Evildoor: This can cause errors if `__name__` is missing, for example when calling the function from interactive python interpreter:

    [vaulov@aiatlas171 Dataflow]$ python
    Python 2.7.5 (default, Jun 20 2019, 20:27:34)
    [GCC 4.8.5 20150623 (Red Hat 4.8.5-36)] on linux2
    Type "help", "copyright", "credits" or "license" for more information.
    >>> from pyDKB.common.misc import log
    2019-08-01 16:21:06 (WARN) (pyDKB.dataflow.cds) Submodule failed (No module named invenio_client.contrib)
    >>> log('123')
      File "pyDKB/common/misc.py", line 40, in log
        prefix = ' (%s)' % mod.__name__
    AttributeError: 'NoneType' object has no attribute '__name__'

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: Fixed, thank you.

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: Fixed, thank you.

mgolosova:
Bugs fixed, please re-review the PR.

Utils/Dataflow/pyDKB/common/types.py
```diff
@@ -0,0 +1,9 @@
+"""
+pyDKB.common.types
+
+Definitions of types used across all the library moduless.
```
Evildoor: Double 's' in 'modules'.

Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -0,0 +1,49 @@
...
+    if args:
+        prefix = ' ' + ' '.join(['(%s)' % p for p in args])
+    else:
+        frm = inspect.stack()[1]
+        mod = inspect.getmodule(frm[0])
+        modname = getattr(mod, '__name__', 'main')
+        prefix = ' (%s)' % modname
+    if lines:
+        dtime = datetime.now().strftime(DTFORMAT)
+        out_message = "%s (%s)%s %s" % (dtime, logLevel.memberName(level),
```
Evildoor: Wouldn't it be better to keep the pre-message part of the strings the same length by adjusting log level format to their longest member? To replace current

    2019-08-06 15:48:54 (WARN) (__main__) 123
    2019-08-06 15:48:54 (FATAL) (__main__) 456

with something 'nicer', such as

    2019-08-06 15:48:54 (WARN)  (__main__) 123
    2019-08-06 15:48:54 (FATAL) (__main__) 456

?

Evildoor:
I have some other concerns regarding this PR and log system changes, but I'll voice them either on team meeting or through mail.

Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -0,0 +1,49 @@
...
+def log(message, level=logLevel.INFO, *args):
+    """ Output log message with given log level.
+
+    :param message: message to output
+    :type message: str
+    :param level: log level of the message
+    :type level: ``pyDKB.common.types.logLevel`` member
+    :param *args: additional prefixes (will be output between log
+                  level prefix and message body)
+    :type *args: str
```
Evildoor: *args can be anything because they are converted to str. Question is, how to indicate this in `:type:`?
https://github.com/PanDAWMS/dkb/pull/188#discussion_r241459053 - some discussion on this point.

Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -0,0 +1,49 @@
...
+    """
+    if not logLevel.hasMember(level):
```
Evildoor: Does not work with incorrect members because `level` is supposed to be member of `logLevel`, which prompts its call in this check:

    >>> log('abc', logLevel.WARN)
    2019-08-26 12:26:47 (WARN) (main) abc
    >>> log('abc', logLevel.WARNN)
    AttributeError: 'Type' object has no attribute 'WARNN'

Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -0,0 +1,49 @@
...
+        log("Unknown log level: %s" % level, logLevel.WARN)
+        level = logLevel.INFO
+    if type(message) == list:
+        lines = message
+    else:
+        lines = message.splitlines()
```
Evildoor: Maybe `str(message).splitlines()`? This is already done for prefixes `*args` and each member of `lines` if `message` is a `list`:

```
>>> log('abc')
2019-08-26 14:55:21 (INFO) (main) abc
>>> log('abc', logLevel.INFO, str)
2019-08-26 14:55:40 (INFO) (<type 'str'>) abc
>>> log(['abc', str])
2019-08-26 14:55:55 (INFO) (main) abc
(==) <type 'str'>
>>> log(str)
  File "pyDKB/common/misc.py", line 34, in log
    lines = message.splitlines()
TypeError: descriptor 'splitlines' of 'str' object needs an argument
>>> log(int)
  File "pyDKB/common/misc.py", line 34, in log
    lines = message.splitlines()
AttributeError: type object 'int' has no attribute 'splitlines'

```

This will also simplify some cases of usage because one wouldn't need to convert to `str` each time. Here:
https://github.com/PanDAWMS/dkb/pull/274/files#diff-9be9001acb47b76c6c21745268c1c088R37
https://github.com/PanDAWMS/dkb/pull/274/files#diff-50722d7919f4fc2cf73626855f67350aR109

Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -0,0 +1,49 @@
...
+
+# Datetime format for log messages
+DTFORMAT = '%Y-%m-%d %H:%M:%S'
+
+
+def log(message, level=logLevel.INFO, *args):
+    """ Output log message with given log level.
+
+    :param message: message to output
+    :type message: str
```
Evildoor: By current function logic it's not `str`, it's `str` or `list` of anything. After resolution of https://github.com/PanDAWMS/dkb/pull/274/files#r317591401 this can probably change to `anything`.

Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -0,0 +1,49 @@
...
+        frm = inspect.stack()[1]
+        mod = inspect.getmodule(frm[0])
+        modname = getattr(mod, '__name__', 'main')
+        prefix = ' (%s)' % modname
+    if lines:
+        dtime = datetime.now().strftime(DTFORMAT)
+        out_message = "%s (%s)%s %s" % (dtime, logLevel.memberName(level),
+                                        prefix, lines[0])
+        for l in lines[1:]:
+            out_message += "\n(==) %s" % l
```
Evildoor: Similar to comment above - maybe adjust the lines' alignment?

```
2019-08-26 15:07:38 (INFO) (main) 1
(==) 2
(==) 3
(==) 4
(==) 5
```
to
```
2019-08-26 15:07:38 (INFO) (main) 1
                             (==) 2
                             (==) 3
                             (==) 4
                             (==) 5
```
?

Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -0,0 +1,49 @@
...
+from types import logLevel
+
+# Datetime format for log messages
+DTFORMAT = '%Y-%m-%d %H:%M:%S'
+
+
+def log(message, level=logLevel.INFO, *args):
+    """ Output log message with given log level.
+
+    :param message: message to output
```
Evildoor: Shouldn't the special treatment of lists and strings with newlines be mentioned somewhere here?

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: There may be somethig in this idea, but:
* broken alignment may make it easier for an eye to catch a changed pattern;
* the second prefix (here: `(__main__)`) is also not supposed to be aligned with others (which may be `(ProcessorStage)` or `(pyDKB)` or whatever, depending on the caller);
* when one parses the log automatically, it is easier if he/she can count on single space after the log level marker. Something like `\s+` will also work, of course, but why make things more difficult?

So I\`d rather leave it as is for now; if it will be annoying, it can always be changed later.

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: Concerning the comment you mention -- why not use `object`? Right now I can not invent an examle of non-object entity.. doesn\`t mean it is not possible, though.

Here, it was specified as 'string' to state that it will be *treated* as string. So if one wants to be sure about what shall appear in the output -- it is better to use string values.

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: Toooooo wide empty space: with `(main)` you have 30 spaces; with `(ProcessorStage)` it\`s gonna be even more. Since logs are more likely to be checked in terminal, 80-characters width is still worth remembering ;) Imagine longer lines:
```
2019-08-26 15:07:38 (INFO) (main) Some error occured: 
                             (==) here are some details about the error: and HERE IS THE MOST IMPORTANT INFORMATION
                             (==) another piece of details: and YET ANOTHER PIECE OF IMPORTANT INFORMATION
```

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: It depends.. if you need to parse some log file, generated with this mechanism -- will you look into the example file to decide what you need to do, or go to the logging implementation?

Utils/Dataflow/pyDKB/common/misc.py
Evildoor: Fair enough.

Utils/Dataflow/pyDKB/common/misc.py
Evildoor: Fair enough.

Utils/Dataflow/pyDKB/common/misc.py
Evildoor: I agree that when tasked with parsing the log the person in question would be more likely to check the examples, but wouldn't it be easier to use the function for its intended purpose if this treatment is mentioned? Someone may assume that function is meant for single lines only and proceed to explicitly separate strings and put them into separate `log()` calls.

Utils/Dataflow/pyDKB/common/types.py
mgolosova: Fixed: 678604ae903c5e1dd91409629ea8e08edbbc22ec.

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: Done: 5ada0e6f623391d551790ef59a94f89a62d9eb7c

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: Changed in 5ada0e6f623391d551790ef59a94f89a62d9eb7c

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: Added more detailed description: 273a39f.

mgolosova:
@Evildoor, I have just updated the `pyDKB` version info, since there were some changes added after the previous value (`0.3.20190710`) had been set.
Please, re-approve.

PR #277, opened 2019-07-10
[WIP] pyDKB: storages 

mgolosova:
Rewrites #244
---

This PR introduces to pyDKB library new submodule storages.

The module is intended to accumulate all the functionality related to the interaction with the storages we use, including both external and internal ones. It basically provides unified way to interact with any of them: object of `Client` class, that understands basic set of methods, like:

* apply configuration (create client object(s));
* send `get()` request (get item by ID);
* send `custom (parametric)` query.

Current modules structure is as follows:
```
pyDKB
  |---> storages
          |---> exceptions // Exceptions definitions
          |
          |---> client
          |        |---> Client  // Interface class
          |
          |---> atlas
                   |---> rucio  // Interface implementation for ATLAS-specific (external) storage
```

Suggested way to use the new functionality is like this:
```
import pyDKB.storages as s

s.setScope('atlas')
rucio = s.getClient('rucio')

mdata = rucio.get('data18_13TeV.00351223.physics_Main.merge.AOD.f937_m197',
                  fields=['events', 'bytes', 'deleted'])

print mdata2
```

For storages like ES will be added following submodules:
* `pyDKB.storages.client.es` -- wrapper for ES client with `pyDKB.storages.client.Client` interface;
* `pyDKB.storages.atlas.chicagoES` -- Chicago ES client;
* `pyDKB.storages.atlas.dkb.es` -- DKB ES client.


Utils/Dataflow/pyDKB/storages/exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+"""
+
+
+class StorageException(Exception):
+    """ Base exception for all storage-related exceptions. """
+    pass
+
+
+class NotFound(StorageException):
+    """ Exeption indicating that record with given ID  not found. """
```
Evildoor: "ExCeption", double space after "ID".

Evildoor:
> Suggested way to use the new functionality is like this: ...

I presume that the last line of the example should be `print mdata` instead of `print mdata2` - no special functionality for changing names like that was implemented here, right?

Side note - I've missed the [WIP] status of the PR, so my apologies if the problems here were already noted and planned to be dealt with later, and I started pointing at them too early.

Utils/Dataflow/pyDKB/storages/client/Client.py
```diff
@@ -0,0 +1,34 @@
...
+
+        :param cfg: configuration parameters
+        :type cfg: dict
+        """
+        raise NotImplementedError
+
+    def get(self, oid, **kwargs):
+        """ Get object / record from storage by ID.
+
+        Raise ``NotFound`` exception if object / record not found.
```
Evildoor: "... record WAS not found."?

mgolosova:
> > Suggested way to use the new functionality is like this: ...
> 
> I presume that the last line of the example should be `print mdata` instead of `print mdata2` - no special functionality for changing names like that was implemented here, right?

Right, it was just a copy-paste mistake, thank you for bringing it to my attention.
But I wonder now if there can *possibly* be such a functionality, that would change name of a variable outside the function... :)

> I've missed the [WIP] status of the PR, so my apologies if the problems here were already noted and planned to be dealt with later, and I started pointing at them too early.

It\`s Ok, the main idea of creating this PR was exactly to let others see what\`s going on and where the development goes. Any comments are welcome, including general ones like "what it is supposed to look like for $this or $that kind of storage?", "suggested way of usage does not look very natural, I\`d expect something like $this or $that" etc.
I am looking at this from "within" so can make very questionable decisions, and it will be very useful to see these questions now, before the development gets to its end and it becomes difficult to change some basic things.

Utils/Dataflow/pyDKB/storages/client/Client.py
mgolosova: Docstrings are more about providing necessary information than to be a literary writing; so "terse language" looks appropriate for me, even though it is not a error message or newspaper headline (check related discussion [here](https://english.stackexchange.com/a/73369)). It may be a matter of habit (and my bad taste, possibly ;) ), but I don\`t think it should be changed.

However if you think that we should set some common rule for such cases, I am open to the further discussion :) (But better not in this PR, as it is not about this one only, but more about common style.)

Utils/Dataflow/pyDKB/storages/atlas/rucio.py
```diff
@@ -0,0 +1,110 @@
...
+from pyDKB.common.misc import (log, logLevel)
+from pyDKB.common.misc import try_to_import
+
+
+if not os.environ.get("VIRTUAL_ENV", None):
+    user_rucio_dir = os.path.expanduser("~/.rucio")
+    if os.path.exists(user_rucio_dir):
+        os.environ["VIRTUAL_ENV"] = os.path.join(user_rucio_dir)
+    else:
+        os.environ["VIRTUAL_ENV"] = os.path.join(base_dir, ".rucio")
```
Evildoor: `base_dir` is not defined. Forgot to copy
https://github.com/PanDAWMS/dkb/blob/master/Utils/Dataflow/091_datasetsRucio/datasets_processing.py#L15
?

mgolosova: Good point, fixed now.

Utils/Dataflow/pyDKB/storages/exceptions.py
mgolosova: Fixed, thank you.

Utils/Dataflow/pyDKB/storages/client/Client.py
```diff
@@ -0,0 +1,34 @@
...
+
+        :param oid: object / record identfier
```
Evildoor: "identifier".

Utils/Dataflow/pyDKB/storages/exceptions.py
```diff
@@ -0,0 +1,25 @@
...
+"""
+
+
+class StorageException(Exception):
+    """ Base exception for all storage-related exceptions. """
+    pass
+
+
+class NotFound(StorageException):
+    """ Exception indicating that record with given ID not found. """
```
Evildoor: Same as `Client.get()` above, "ID WAS not found". When consensus will be reached, this should be changed (or left as-is) accordingly.

mgolosova: @Evildoor, consensus takes at least a discussion, but I still don\`t know your opinion on [this comment](https://github.com/PanDAWMS/dkb/pull/277#discussion_r302506396) ;)

Utils/Dataflow/pyDKB/storages/client/Client.py
mgolosova: Fixed, thank you (2d88d52).

Utils/Dataflow/pyDKB/storages/atlas/rucio.py
```diff
@@ -0,0 +1,111 @@
...
+    """ Implement common interface for Rucio client. """
+
+    def __init__(self, *args, **kwargs):
+        """ Initialize instance as parent client class object. """
+        ParentClientClass.__init__(self, *args, **kwargs)
+
+    def get(self, oid, **kwargs):
+        """ Get dataset metadata.
+
+        Implementation of interface method `Clent.get()`.
```
Evildoor:
```suggestion
        Implementation of interface method `Client.get()`.
```

Utils/Dataflow/pyDKB/storages/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+
+    :return: storage client
+    :rtype: client.Client
+    """
+    if scope:
+        scope = getScope(scope)
+    else:
+        scope = _scope
+    if scope is None:
+        raise exceptions.StorageException("Storages scope not specified")
```
Evildoor:
```suggestion
        raise exceptions.StorageException("Storage's scope not specified")
```

mgolosova: It is not a mistake (or, better to say, it is a *different* mistake): the message says that scope of storages, from which should be picked one (or maybe "storage lookup scope"?), is not specified. "Storage scope" is more like a scope of a single storage -- which wouldn\`t have much sense in this situation.

Will be changed, thank you!

Utils/Dataflow/pyDKB/common/misc.py
```diff
@@ -46,3 +47,25 @@ def log(message, level=logLevel.INFO, *args):
...
+    try:
+        result = importlib.import_module(modname)
+        if attrname:
+            result = getattr(result, attrname)
+    except ImportError, err:
+        log("Failed to import '%s'.\nDetails: %s" % (modname, err),
+            logLevel.ERROR)
+    except AttributeError:
+        if attrname:
+            log("Failed to import '%s' from '%s'" % (attrname, modname))
```
Evildoor: 1. In this case module will be returned as `result`. What's the point of returning the whole module when an attribute is expected? Maybe result should stay `False`?
2. Shouldn't this message be logged as ERROR, and not INFO?
3. Shouldn't the error message be displayed?

mgolosova:
> mgolosova [force-pushed](https://github.com/PanDAWMS/dkb/compare/7b74454a9ce5c40599798e7c883a3030043a3dec..a349cb6bbedd37986d7bed2a1e560b80aa9944c1) the pyDKB-storages-2  branch from 7b74454 to a349cb6

...to rebase PR on the current version of the `pyDKB` (where `LoggableObject` is already merged).

Utils/Dataflow/pyDKB/common/misc.py
mgolosova: @Evildoor,

1-2: fair point, fixed in 6360f5d.

3: a bit unclear, but still a fair point too -- if you meant that in case of module failed to be imported due to the `AttributeError` we kind of ignore this error, silently returning `False`.
Fixed this in 5891f41 (and, coincidentally, made code much simplier). Also made return value less confusing: in case of error function will return its own specific value `NOT_IMPORTED`, defined in same module, instead of `False` (see 10ac41a).

---

*Please take a look and mark this conversation as solved if everything is OK.*

PR #278, opened 2019-07-12
DF: update README

anastasiakaida:
This update is about the description of stage types.
The main reason to add this block is a declaration what stage in a dataflow
pipeline is responsible for data preparation before uploading (sink or
transform stage).

We consider to leave sink stages pretty simple. All data preparation
work should be done by pre-sink stage -- the last one transform stage in
any data pipeline.

### DIFF
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -141,15 +141,16 @@ There are three types of stages:
 1. Source stage - an initial stage in a dataflow pipeline that gets data
 from an external source (CDS, Rucio, AMI, etc.).
 
-2. Transform stage - these stages perform a set of operations for different
-cases on metadata and form output messages according to a particular pattern
-(e.g. select relevant metadata from a raw dataset, append metadata
-to a message, prepare messages for upload).
+2. Transform stage - a transitional stage that performs single logical
+operation on the metadata. Set (chain) of transform stages forms the whole
+transformation process for messages emitted by Source. Transformation
+operations may include: format change, derived values calculation,
+message extension with metadata from an additional source, etc.
 
 2.1. Pre-sink stage - the last transform stage in a dataflow pipeline that
 prepares metadata for upload via sink stage; it awares about messages
-formatting rules. All the preparations should be done by the pre-sink
-stage.
+formatting rules. All the preparations for upload should be done by
+the pre-sink stage.
 
 3. Sink stage - a stage that uploads data into the DKB storage
 (Virtuoso, ES, ...). Every single sink stage should be preceded by
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -156,7 +156,7 @@ the pre-sink stage.
 (Virtuoso, ES, ...). Every single sink stage should be preceded by
 a corresponding pre-sink stage that transforms data and forms ready-to-send
 messages with a specific structure. Sink stage only gets prepared messages
-and uploads it without any re-formatting. It does not care about messages
+and uploads them without any re-formatting. It does not care about messages
 content. If message does not follow the template, most likely it
 will be rejected by the storage load API.
 
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -106,14 +106,16 @@ messages to upload).
 
 2.1 Pre-sink stage - the last transform stage in a dataflow pipeline that
 prepares metadata to upload via sink stage; it awares about messages
-formatting. All the messages preparations should be done by a pre-sink stage.
+formatting rules. All the messages preparations should be done by a pre-sink
+stage.
 
-3. Sink stage - a stage that uploads data into the DKB storage (Virtuoso
-or ES). We consider that every single sink stage requires a particular
-pre-sink stage that transform data and form ready-to-send messages with a
-particular structure. Sink stage only gets prepared messages and upload it
-without any re-formatting. It does not care about messages content.
-If message does not follow the template, it will not be uploaded.
+3. Sink stage - a stage that uploads data into the DKB storage
+(Virtuoso, ES, ...). Every single sink stage should be preceded by
+a corresponding pre-sink stage that transforms data and forms ready-to-send
+messages with a specific structure. Sink stage only gets prepared messages
+and uploads it without any re-formatting. It does not care about messages
+content. If message does not follow the template, most likely it
+will be rejected by the storage load API.
 
 ==============
 * REFERENCES *
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
 ES. It gets a very basic set of metadata from ProdSys2, adds ES-related
 fields, and checks that it is present in ES rather than uploading it.
 
+================
+* Stages types *
+================
+There are three types of stages:
+
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.)
+
+2. Transform stage - a stage that makes a set of different operations
+with metadata forming messages according to a particular pattern (e.g. select
+relevant metadata from a raw dataset, append metadata to a message, prepare
+messages to upload).
+
+2.1 Pre-sink stage - the last transform stage in a dataflow pipeline that
+prepares metadata to upload via sink stage; it awares about messages
+formatting. All the messages preparations should be done by a pre-sink stage.
+
+3. Sink stage - a stage that uploads data into the DKB storage (Virtuoso
+or ES). We consider that every single sink stage requires a particular
+pre-sink stage that transform data and form ready-to-send messages with a
+particular structure. Sink stage only gets prepared messages and upload it
+without any re-formatting. It does not care about messages content.
+If message does not follow the template, it will not be uploaded.
+
 ==============
 * REFERENCES *
 ==============
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -148,7 +148,7 @@ operations may include: format change, derived values calculation,
 message extension with metadata from an additional source, etc.
 
 2.1. Pre-sink stage - the last transform stage in a dataflow pipeline that
-prepares metadata for upload via sink stage; it awares about messages
+prepares metadata for upload via sink stage; it is aware of messages'
 formatting rules. All the preparations for upload should be done by
 the pre-sink stage.
 
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -143,7 +143,7 @@ from an external source (CDS, Rucio, AMI, etc.).
 
 2. Transform stage - a transitional stage that performs single logical
 operation on the metadata. Set (chain) of transform stages forms the whole
-transformation process for messages produced by Source. Transformation
+transformation process for messages produced by a source stage. Transformation
 operations may include: format change, derived values calculation,
 message extension with metadata from an additional source, etc.
 
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -146,7 +146,7 @@ cases on metadata and form output messages according to a particular pattern
 (e.g. select relevant metadata from a raw dataset, append metadata
 to a message, prepare messages for upload).
 
-2.1 Pre-sink stage - the last transform stage in a dataflow pipeline that
+2.1. Pre-sink stage - the last transform stage in a dataflow pipeline that
 prepares metadata for upload via sink stage; it awares about messages
 formatting rules. All the preparations should be done by the pre-sink
 stage.
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -99,7 +99,7 @@ There are three types of stages:
 1. Source stage - an initial stage in a dataflow pipeline that gets data
 from an external source (CDS, Rucio, AMI, etc.).
 
-2. Transform stage - a stage that makes a set of different operations
+2. Transform stage - these stages make a set of operations for different cases
 with metadata forming messages according to a particular pattern (e.g. select
 relevant metadata from a raw dataset, append metadata to a message, prepare
 messages to upload).
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -97,7 +97,7 @@ fields, and checks that it is present in ES rather than uploading it.
 There are three types of stages:
 
 1. Source stage - an initial stage in a dataflow pipeline that gets data
-from an external source (CDS, Rucio, AMI, etc.)
+from an external source (CDS, Rucio, AMI, etc.).
 
 2. Transform stage - a stage that makes a set of different operations
 with metadata forming messages according to a particular pattern (e.g. select
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -105,8 +105,8 @@ relevant metadata from a raw dataset, append metadata to a message, prepare
 messages to upload).
 
 2.1 Pre-sink stage - the last transform stage in a dataflow pipeline that
-prepares metadata to upload via sink stage; it awares about messages
-formatting rules. All the messages preparations should be done by a pre-sink
+prepares metadata for upload via sink stage; it awares about messages
+formatting rules. All the preparations should be done by the pre-sink
 stage.
 
 3. Sink stage - a stage that uploads data into the DKB storage
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -92,7 +92,7 @@ ES. It gets a very basic set of metadata from ProdSys2, adds ES-related
 fields, and checks that it is present in ES rather than uploading it.
 
 ================
-* Stages types *
+* Stage types *
 ================
 There are three types of stages:
 
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -91,32 +91,6 @@ intended for making sure that information is consistent between ProdSys2 and
 ES. It gets a very basic set of metadata from ProdSys2, adds ES-related
 fields, and checks that it is present in ES rather than uploading it.
 
-================
-* Stage types *
-================
-There are three types of stages:
-
-1. Source stage - an initial stage in a dataflow pipeline that gets data
-from an external source (CDS, Rucio, AMI, etc.).
-
-2. Transform stage - these stages perform a set of operations for different
-cases on metadata and form output messages according to a particular pattern
-(e.g. select relevant metadata from a raw dataset, append metadata
-to a message, prepare messages for upload).
-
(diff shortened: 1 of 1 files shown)


anastasiakaida:
@Evildoor during the previous meeting you told that you have something to discuss about this PR. Please, review. It would be nice to get your feedback. :)

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
 ES. It gets a very basic set of metadata from ProdSys2, adds ES-related
 fields, and checks that it is present in ES rather than uploading it.
 
+================
+* Stages types *
```
Evildoor: I think that "stages" should be in singular form: "Stage types".

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+================
+There are three types of stages:
+
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.)
```
Evildoor: Missing period at the end of sentence: "... AMI, etc.).". I had my doubts about this, but it seems that this particular case have already been discussed: https://english.stackexchange.com/questions/23022/how-to-deal-with-abbreviations-like-etc-at-the-end-of-parentheses-which-are-c.

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+
+2. Transform stage - a stage that makes a set of different operations
```
Evildoor: 1) "... stage that PERFORMS a set of ..." or something else. "Stage makes operations" sounds unnatural.
2) I'm not sure that operations are always different. What if stage performs a single operation? What if it performs several, but all of them are very similar?

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+with metadata forming messages according to a particular pattern (e.g. select
+relevant metadata from a raw dataset, append metadata to a message, prepare
+messages to upload).
+
+2.1 Pre-sink stage - the last transform stage in a dataflow pipeline that
+prepares metadata to upload via sink stage; it awares about messages
```
Evildoor: "... via sink stage; it is aware of messages' formatting."

Utils/Dataflow/README
mgolosova: "... -- these stages perform different operations..."?
The idea, I believe, was to emphasize that "transformation" actions can be really different for different cases. Or maybe not )

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+================
+* Stages types *
+================
+There are three types of stages:
+
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.)
+
+2. Transform stage - a stage that makes a set of different operations
+with metadata forming messages according to a particular pattern (e.g. select
```
mgolosova: ...on metadata...

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+relevant metadata from a raw dataset, append metadata to a message, prepare
+messages to upload).
+
+2.1 Pre-sink stage - the last transform stage in a dataflow pipeline that
+prepares metadata to upload via sink stage; it awares about messages
+formatting. All the messages preparations should be done by a pre-sink stage.
```
mgolosova: ...all message preparations...
(or "all the messages should be prepared by...", or "all the preparations")

But maybe "by the pre-sink stage" (or "by the pre-sink stages" -- to emphasize that there may be many pre-sinks), for the whole paragraph is a description of this specific type of stages and the last sentence refers to the same thing the reader is already aware about.

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+
+3. Sink stage - a stage that uploads data into the DKB storage (Virtuoso
+or ES). We consider that every single sink stage requires a particular
```
mgolosova: Better to keep the list open: "(Virtuoso, ES, ...)" -- for technically we are not limited to these two storages, and "sink stage" will remain "sink" even if it works with Neo4j or Cassandra ;)

"We consider" -> "we expect" / "it is supposed" / "every sink stage requires" ?

"particular" -> "specific"? Maybe it is just me, but "particular" sounds more like "something in question" for me, like "this particular case".

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+pre-sink stage that transform data and form ready-to-send messages with a
+particular structure. Sink stage only gets prepared messages and upload it
+without any re-formatting. It does not care about messages content.
+If message does not follow the template, it will not be uploaded.
```
mgolosova: "...most likely it will be rejected by the storage load API"?
If simply say "not uploaded", it is not clear who is responsible for "not uploading" the message.

Utils/Dataflow/README
Evildoor: I suggest making this impersonal and changing some other things: "... or ES). Every single sink stage should be preceded by a corresponding pre-sink stage that ..."

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+relevant metadata from a raw dataset, append metadata to a message, prepare
+messages to upload).
+
+2.1 Pre-sink stage - the last transform stage in a dataflow pipeline that
+prepares metadata to upload via sink stage; it awares about messages
+formatting. All the messages preparations should be done by a pre-sink stage.
+
+3. Sink stage - a stage that uploads data into the DKB storage (Virtuoso
+or ES). We consider that every single sink stage requires a particular
+pre-sink stage that transform data and form ready-to-send messages with a
```
Evildoor: "transformS", "formS".

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+particular structure. Sink stage only gets prepared messages and upload it
```
Evildoor: "uploadS".

Utils/Dataflow/README
Evildoor: This, in my opinion, is a part of the larger problem. Usually it is possible to say "perform operations with metadata", but in this case we have a construction "perform operations with metadata forming messages according to pattern", which can be interpreted in many ways (especially if a possibility of missing comma or two is considered), such as:
1. Perform operations, and in these operations metadata forms messages according to pattern.
2. Perform operations according to pattern, and in these operations metadata forms messages.
3. Perform operations with metadata and, by doing so, form messages according to pattern.
4. Perform operations with some of the metadata - the ones that form messages according to pattern.
I suggest rewording the sentence to avoid such ambiguity, and changing "with" to "on" looks like a good start, as it "cuts off" some "branches".

Utils/Dataflow/README
Evildoor: I still doubt the word "different" here. Also, come to think of it - possibility (and existence of 016) of stage performing a single operation is an argument against "set of" as well, not only "different".
Emphasizing of possible difference - maybe it can be done elsewhere, or in another way?

anastasiakaida: @Evildoor As Marina said, transformation operations can be different, but I can emphasize it like this "...these stages perform a set of operations for different cases". Hope, it sounds correct.

Utils/Dataflow/README
```diff
@@ -91,6 +91,30 @@ intended for making sure that information is consistent between ProdSys2 and
...
+* Stages types *
+================
+There are three types of stages:
+
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.)
+
+2. Transform stage - a stage that makes a set of different operations
+with metadata forming messages according to a particular pattern (e.g. select
+relevant metadata from a raw dataset, append metadata to a message, prepare
```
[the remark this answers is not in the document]
anastasiakaida: @mgolosova It means not already transformed, but uploaded with any source stage data. :)

anastasiakaida:
@mgolosova, @Evildoor 
Now README is fixed according to your comments. Thank you!
Please, review. :)

Utils/Dataflow/README
```diff
@@ -133,8 +133,34 @@ following requirements:
...
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.).
+
+2. Transform stage - these stages perform a set of operations for different
+cases on metadata and form output messages according to a particular pattern
+(e.g. select relevant metadata from a raw dataset, append metadata
+to a message, prepare messages for upload).
+
+2.1 Pre-sink stage - the last transform stage in a dataflow pipeline that
+prepares metadata for upload via sink stage; it awares about messages
```
Evildoor: "... ; is is aware of messages' formatting rules. ..."

Utils/Dataflow/README
```diff
@@ -133,8 +133,34 @@ following requirements:
...
+==================
+There are three types of stages:
+
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.).
+
+2. Transform stage - these stages perform a set of operations for different
+cases on metadata and form output messages according to a particular pattern
+(e.g. select relevant metadata from a raw dataset, append metadata
+to a message, prepare messages for upload).
```
Evildoor: Maybe "... prepare a message for upload"? To follow the style of two previous examples.

Utils/Dataflow/README
```diff
@@ -133,8 +133,34 @@ following requirements:
...
+==================
+* 3. Stage types *
+==================
+There are three types of stages:
+
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.).
+
+2. Transform stage - these stages perform a set of operations for different
+cases on metadata and form output messages according to a particular pattern
```
mgolosova: Maybe it worth to be completely reworded?
Something like
```
2. Transform stage -- transitional stage that performs single logical
operation on the metadata. Set (chain) of transform stages forms the whole
tarnsformation process for messages emitted by Source.
Transformation opeations may include: format change, derived values
calculation, message extention with metadata from additional source, etc.
```

Utils/Dataflow/README
mgolosova: Maybe "preparation for upload" no to be mentioned here at all? I mean, we have "2.1. Pre-sink stage" section below -- it looks like a rightful place to talk about this for me...

Utils/Dataflow/README
```diff
@@ -133,8 +133,34 @@ following requirements:
...
+(e.g. select relevant metadata from a raw dataset, append metadata
+to a message, prepare messages for upload).
+
+2.1 Pre-sink stage - the last transform stage in a dataflow pipeline that
```
mgolosova:
```suggestion
2.1. Pre-sink stage - the last transform stage in a dataflow pipeline that
```

Utils/Dataflow/README
mgolosova: Resolving conversation in favor of the next iteration here: https://github.com/PanDAWMS/dkb/pull/278#discussion_r313291818

Utils/Dataflow/README
mgolosova: New variant uses the `operations on metadata` version so I guess this discussion may be resolverd now.

Utils/Dataflow/README
```diff
@@ -133,8 +133,34 @@ following requirements:
...
+* 3. Stage types *
+==================
+There are three types of stages:
+
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.).
+
+2. Transform stage - these stages perform a set of operations for different
+cases on metadata and form output messages according to a particular pattern
+(e.g. select relevant metadata from a raw dataset, append metadata
```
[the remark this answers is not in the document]
mgolosova: Sorry I missed related conversation above(https://github.com/PanDAWMS/dkb/pull/278#discussion_r304831028), so I resolve this conversation and will add a coment there.

Utils/Dataflow/README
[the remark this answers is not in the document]
mgolosova: In terms of ATLAS-related dataflow the "raw dataset" (as well as any other "dataset") is more likely to be interpreted as the dataset itself (the RAW one, containing raw data from the detector, or any other). I guess these datasets do contain some metadata, butwe do not work with any datasets directly, so... I think it would be better to say not "dataset", but "dataset metadata" -- should sound less confusing.

Utils/Dataflow/README
mgolosova: @Evildoor, seems like you missed my reply to the similar comment of yours above:
 (https://github.com/PanDAWMS/dkb/pull/278#discussion_r304834350). Could you take a look and let me (and @anastasiakaida) know what do you think?

Evildoor: I'd rather say that what's wrong here is the fact that I've made the comment and then (regardless of your reply's existence) forgot about it and made the same suggestion again. My bad.

Utils/Dataflow/README
Evildoor: I think that "message formatting rules" is the best out of the three suggested options.

Utils/Dataflow/README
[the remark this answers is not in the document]
anastasiakaida: @mgolosova And you asked (and I even answered ;) ), but there was a difference between definitions.
From my point of view, it was something like "not already transformed, but uploaded with any source stage data", I mean "data from an external source as it was", but I see -- in terms of ATLAS-related dataflow, it is a completely different thing. I suppose I can reword it to prevent misunderstanding.

anastasiakaida:
@mgolosova, @Evildoor 
Fixed! Thank you! :)

Utils/Dataflow/README
mgolosova: The conversation is not solved (I mean, the original comment of @Evildoor is still not addressed in the current version of the README).

Utils/Dataflow/README
```diff
@@ -133,8 +133,35 @@ following requirements:
...
+* 3. Stage types *
+==================
+There are three types of stages:
+
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.).
+
+2. Transform stage - a transitional stage that performs single logical
+operation on the metadata. Set (chain) of transform stages forms the whole
+transformation process for messages emitted by Source. Transformation
```
Evildoor: "Emitted" is usually applied to energy/light/etc. How about "produced", "constructed", or something else?

anastasiakaida:
> Few line comments to some particular moments, and one more general: maybe move the whole section to "REFERENCES"? It is said to contain "Conventions", and convention it is: what specific stages should look like to fit the architecture/concept. Somewhere after "streaming mode", I think?

Fixed. Now it is under "references".

Utils/Dataflow/README
anastasiakaida: @Evildoor fixed

Utils/Dataflow/README
anastasiakaida: According to the discussion:
>The dot in etc. If it's inside the parentheses, like in this case, then you should use two dots, since one is for the abbreviation, the other is for the sentence.

Fixed. Now it looks like you've described.

Utils/Dataflow/README
anastasiakaida: @mgolosova My bad. Fixed now.

@Evildoor Thank you!

Utils/Dataflow/README
anastasiakaida: @mgolosova 
Fixed! Thank you!

Utils/Dataflow/README
anastasiakaida: @Evildoor 
Fixed!

Utils/Dataflow/README
anastasiakaida: @Evildoor 
I think, "produced" may be correct enough for the context

anastasiakaida:
@Evildoor @mgolosova 
I think, everything is fixed now. Please, check this PR (hope, according to @Evildoor comment, this will be the last one).
:)

Utils/Dataflow/README
```diff
@@ -133,8 +133,35 @@ following requirements:
...
+* 3. Stage types *
+==================
+There are three types of stages:
+
+1. Source stage - an initial stage in a dataflow pipeline that gets data
+from an external source (CDS, Rucio, AMI, etc.).
+
+2. Transform stage - a transitional stage that performs single logical
+operation on the metadata. Set (chain) of transform stages forms the whole
+transformation process for messages produced by Source. Transformation
```
Evildoor: The word "Source" should not be capitalized: "source".

Utils/Dataflow/README
```diff
@@ -133,8 +133,35 @@ following requirements:
...
+2.1. Pre-sink stage - the last transform stage in a dataflow pipeline that
+prepares metadata for upload via sink stage; it is aware of messages'
+formatting rules. All the preparations for upload should be done by
+the pre-sink stage.
+
+3. Sink stage - a stage that uploads data into the DKB storage
+(Virtuoso, ES, ...). Every single sink stage should be preceded by
+a corresponding pre-sink stage that transforms data and forms ready-to-send
+messages with a specific structure. Sink stage only gets prepared messages
+and uploads it without any re-formatting. It does not care about messages
```
mgolosova:
```suggestion
and uploads them without any re-formatting. It does not care about messages
```

Utils/Dataflow/README
mgolosova: If it should go with lowercase initial letter, then I suggest to extend it to "source stage". Else one may take it as "metadata source" (just as in all the other cases where lowercase "source" appears), while what produces *messages* should clearly be the stage.

mgolosova:
@anastasiakaida, last two comments left, I believe :) We almost there, thank you for bringing things to order!

Utils/Dataflow/README
anastasiakaida: @mgolosova, @Evildoor
I suppose to change it into a "source stage". :)

anastasiakaida:
@mgolosova Fixed! :)

mgolosova:
@anastasiakaida, great! I have approved the PR, so now we wait for the final review from @Evildoor and if everything\`s fine it will be merged soon.

PR #281, opened 2019-07-17
017 get_category() improvement

Evildoor:
- Remove dictionary redifinition.
- Replace a succession of IFs with a cycle.

### DIFF
--- a/Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
+++ b/Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
@@ -28,6 +28,33 @@ except Exception, err:
     sys.exit(1)
 
 
+PHYS_CATEGORIES_MAP = {
+    'BPhysics': ['charmonium', 'jpsi', 'bs', 'bd', 'bminus', 'bplus',
+                 'charm', 'bottom', 'bottomonium', 'b0'],
+    'BTag': ['btagging'],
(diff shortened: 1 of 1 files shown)


Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
```diff
@@ -28,6 +28,45 @@
...
+    'SUSY': ['bino', 'susy', 'pmssm', 'leptosusy', 'rpv', 'mssm'],
+    'Triboson': ['triplegaugecoupling', 'triboson', 'zzw', 'www'],
+    'TTbar': ['ttbar'],
+    'TTbarX': ['ttw', 'ttz', 'ttv', 'ttvv', '4top', 'ttww'],
+    'Upgrade': ['upgrad'],
+    'Wjets': ['w'],
+    'Zjets': ['z']}
+
+
+PHYS_CATEGORIES_TASKNAME_MAP = {
```
mgolosova: It is more like "PHYS_SHORT_MAP" (or "PH_SORT_MAP"), then "TASKNAME". Just to avoid misunderstanding, or else someone might believe we really check the whole taskname, while in fact it is only "physics short" field.

Evildoor: Done.

PR #282, opened 2019-07-31
[override] 095 update: part 1

Evildoor:
Stage needs to be improved per AMI team's request (https://trello.com/c/x0hvd4aQ). However, attempts in doing so are somewhat hindered by other existing problems with the stage, such as error processing (https://trello.com/c/WDccdBxS). Therefore, this PR's aim is to fix or improve the "local" issues not related (or related unintentionally) to the question of lessening the pressure on AMI - attempts to solve the question of pressure will follow in a separate PR.

### DIFF
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -95,7 +95,16 @@ def get_ami_client():
 
 
 def process(stage, message):
-    """ Single message processing. """
+    """ Process a message.
+
+    Implementation of :py:meth:`.ProcessorStage.process` for hooking
+    the stage into DKB workflow.
+
+    :param stage: stage instance
+    :type stage: pyDKB.dataflow.stage.ProcessorStage
+    :param message: input message with data
+    :type message: pyDKB.dataflow.communication.messages.JSONMessage
+    """
     data = message.content()
     if not isinstance(data, dict):
         sys.stderr.write("(WARN) Cannot update non-dict data: %r\n" % data)
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -167,7 +167,7 @@ def amiPhysValues(data):
 
 
 def change_key_names(data):
-    """ Change parameter names according to PHYS_VALUES dictionary.
+    """ Change parameter names from ones used by AMI to corresponding ES ones.
 
     :param data: JSON string
     :return: JSON string
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -122,8 +122,7 @@ def amiPhysValues(data):
     - k_factor
     - cross_section_ref
     """
-    dataset = data['datasetname']
-    container = remove_tid(dataset)
+    container = container_name(data)
     ami_client = get_ami_client()
     try:
         res = ami_client.execute(['GetPhysicsParamsForDataset',
@@ -170,17 +169,17 @@ def change_key_names(data):
     return data
 
 
-def remove_tid(dataset):
-    """ Remove TaskID (_tidXX) part from dataset name.
+def container_name(data):
+    """ Retrieve container name from information about dataset.
 
     As AMI GetPhysicsParamsForDataset works with containers,
     we construct the container name from each dataset name,
     removing the _tid{...} part
 
-    :param dataset: dataset name
+    :param data: dataset information
     :return: dataset name without _tid => container name
     """
-    return re.sub('_tid(.)+', '', dataset)
+    return re.sub('_tid(.)+', '', data['datasetname'])
 
 
 if __name__ == '__main__':
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -130,7 +130,12 @@ def amiPhysValues(data):
                                   '--logicalDatasetName=%s' % container],
                                  format='json')
         json_str = json.loads(res)
-        for row in json_str['AMIMessage'][0]['Result'][0]['rowset'][0]['row']:
+        rowset = json_str['AMIMessage'][0]['Result'][0]['rowset']
+        if not rowset:
+            sys.stderr.write("(WARN) No values found in AMI for dataset '%s'\n"
+                             % data['datasetname'])
+            return False
+        for row in rowset[0]['row']:
             p_name, p_val = None, None
             for field in row['field']:
                 if field['@name'] == 'paramName':
@@ -147,9 +152,9 @@ def amiPhysValues(data):
         sys.stderr.write("(ERROR) Failed to process dataset '%s': "
                          "%r\n" % (data['datasetname'], e))
         return False
-    except Exception:
-        sys.stderr.write("(WARN) No values found in AMI for dataset '%s'\n"
-                         % data['datasetname'])
+    except Exception as e:
+        sys.stderr.write("(ERROR) Failed to process dataset '%s': "
+                         "%r\n" % (data['datasetname'], e))
         return False
 
 
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -137,10 +137,12 @@ def amiPhysValues(data):
                     data[p_name] = p_val
                     p_name, p_val = None, None
                     continue
-        return change_key_names(data)
+        change_key_names(data)
+        return True
     except Exception:
         sys.stderr.write("(WARN) No values found in AMI for dataset '%s'\n"
                          % data['datasetname'])
+        return False
 
 
 def change_key_names(data):
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -123,7 +123,7 @@ def amiPhysValues(data):
     ami_client = get_ami_client()
     try:
         res = ami_client.execute(['GetPhysicsParamsForDataset',
-                                  "--logicalDatasetName=%s" % container],
+                                  '--logicalDatasetName=%s' % container],
                                  format='json')
         json_str = json.loads(res)
         for row in json_str['AMIMessage'][0]['Result'][0]['rowset'][0]['row']:
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -36,7 +36,11 @@ FILTER = ['AOD', 'EVNT', 'HITS']
 
 
 def main(argv):
-    """ Main program body. """
+    """ Main program body.
+
+    :param argv: arguments
+    :type argv: list
+    """
     stage = pyDKB.dataflow.stage.ProcessorStage()
     stage.set_input_message_type(messageType.JSON)
     stage.set_output_message_type(messageType.JSON)
@@ -63,8 +67,9 @@ def init_ami_client(userkey='', usercert=''):
     client object.
 
     :param userkey: user key pem file
+    :type userkey: str
     :param usercert: user certificate pem file
-    :return:
+    :type usercert: str
     """
     global ami_client
     try:
@@ -89,7 +94,11 @@ def init_ami_client(userkey='', usercert=''):
 
 
 def get_ami_client():
-    """ Get configured AMI client. """
+    """ Get configured AMI client.
+
+    :return: AMI client instance (global variable)
+    :rtype: pyAMI.client.Client
+    """
     if not ami_client:
         init_ami_client()
     return ami_client
@@ -105,6 +114,9 @@ def process(stage, message):
     :type stage: pyDKB.dataflow.stage.ProcessorStage
     :param message: input message with data
     :type message: pyDKB.dataflow.communication.messages.JSONMessage
+
+    :return: False (failed to process message) or True (otherwise)
+    :rtype: bool
     """
     data = message.content()
     if not isinstance(data, dict):
@@ -129,7 +141,14 @@ def process(stage, message):
 
 
 def amiPhysValues(data):
-    """ Update data with information from AMI. """
+    """ Update data with information from AMI.
+
+    :param data: data to update
+    :type data: dict
+
+    :return: True (update was successful) or False (otherwise)
+    :rtype: bool
+    """
     container = container_name(data)
     if not container:
         return False
@@ -169,8 +188,11 @@ def amiPhysValues(data):
 def change_key_names(data):
     """ Change parameter names from ones used by AMI to corresponding ES ones.
 
-    :param data: JSON string
-    :return: JSON string
+    :param data: data to update
+    :type data: dict
+
+    :return: updated data
+    :rtype: dict
     """
     for item in PHYS_VALUES:
         if item["ami"] in data:
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -172,12 +172,14 @@ def change_key_names(data):
 def container_name(data):
     """ Retrieve container name from information about dataset.
 
-    As AMI GetPhysicsParamsForDataset works with containers,
-    we construct the container name from each dataset name,
-    removing the _tid{...} part
+    The container name is extracted from dataset name by removing
+    the '_tid...' part.
 
     :param data: dataset information
-    :return: dataset name without _tid => container name
+    :type data: dict
+
+    :return: container name if it was determined successfully, False otherwise
+    :rtype: str or bool
     """
     return re.sub('_tid(.)+', '', data['datasetname'])
 
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -123,6 +123,8 @@ def amiPhysValues(data):
     - cross_section_ref
     """
     container = container_name(data)
+    if not container:
+        return False
     ami_client = get_ami_client()
     try:
         res = ami_client.execute(['GetPhysicsParamsForDataset',
@@ -181,7 +183,19 @@ def container_name(data):
     :return: container name if it was determined successfully, False otherwise
     :rtype: str or bool
     """
-    return re.sub('_tid(.)+', '', data['datasetname'])
+    if 'datasetname' in data:
+        dataset = data['datasetname']
+    else:
+        sys.stderr.write("(WARN) Required field 'datasetname' not found "
+                         "in data: %r\n" % data)
+        return False
+    if not isinstance(dataset, (str, unicode)):
+        sys.stderr.write("(WARN) Invalid type of 'datasetname' field: "
+                         "expected string, got %s.\n"
+                         "(==) Data: "
+                         "%r\n" % (dataset.__class__.__name__, data))
+        return False
+    return re.sub('_tid(.)+', '', dataset)
 
 
 if __name__ == '__main__':
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -97,9 +97,11 @@ def init_ami_client(userkey='', usercert=''):
     try:
         ami_client.execute('ListCommands')
     except Exception as e:
-        sys.stderr.write("(WARN) Failed to perform test command "
+        sys.stderr.write("(ERROR) Failed to perform test command "
                          "ListCommands. Are you sure you have"
                          " a valid certificate?\nException: %s\n" % str(e))
+        raise DataflowException("Failed to perform test command"
+                                " ListCommands.")
 
 
 def get_ami_client():
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -50,12 +50,15 @@ def main(argv):
 
     stage.configure(argv)
     stage.process = process
-    if stage.ARGS.userkey and stage.ARGS.usercert:
+    try:
         init_ami_client(stage.ARGS.userkey, stage.ARGS.usercert)
-    exit_code = stage.run()
+    except Exception:
+        exit_code = 1
+    else:
+        exit_code = stage.run()
 
-    if exit_code == 0:
-        stage.stop()
+        if exit_code == 0:
+            stage.stop()
 
     sys.exit(exit_code)
 
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -93,6 +93,9 @@ def get_ami_client():
 def process(stage, message):
     """ Single message processing. """
     data = message.content()
+    if not isinstance(data, dict):
+        sys.stderr.write("(WARN) Cannot update non-dict data: %r\n" % data)
+        return False
     # 'data_format' field contains a list of strings,
     # e.g. ['DAOD_SUSY5', 'DAOD']
     formats = data.get('data_format', [])
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -80,8 +80,8 @@ def init_ami_client(userkey='', usercert=''):
                                          cert_file=usercert)
         AtlasAPI.init()
     except NameError:
-        sys.stderr.write("(FATAL) Failed to initialise AMI client: "
-                         "pyAMI module is not loaded.\n")
+        sys.stderr.write("(FATAL) Failed to initialise AMI client:"
+                         " pyAMI module is not loaded.\n")
         raise DataflowException("Module not found: 'pyAMI'")
     except Exception, err:
         sys.stderr.write(
@@ -89,16 +89,16 @@ def init_ami_client(userkey='', usercert=''):
             " Are you sure you have a valid certificate?\n")
         raise DataflowException(str(err))
     if ami_client.config.conn_mode == ami_client.config.CONN_MODE_LOGIN:
-        sys.stderr.write("(ERROR) Login authentication mode is not "
-                         "supported. Please provide user certificate or create"
-                         " proxy.\n")
-        raise DataflowException("Failed to initialise AMI client: certificate "
-                                "not provided or not found.")
+        sys.stderr.write("(ERROR) Login authentication mode is not"
+                         " supported. Please provide user certificate or"
+                         " create proxy.\n")
+        raise DataflowException("Failed to initialise AMI client: certificate"
+                                " not provided or not found.")
     try:
         ami_client.execute('ListCommands')
     except Exception as e:
-        sys.stderr.write("(ERROR) Failed to perform test command "
-                         "ListCommands. Are you sure you have"
+        sys.stderr.write("(ERROR) Failed to perform test command"
+                         " ListCommands. Are you sure you have"
                          " a valid certificate?\nException: %s\n" % str(e))
         raise DataflowException("Failed to perform test command"
                                 " ListCommands.")
@@ -224,18 +224,18 @@ def container_name(data):
     if 'datasetname' in data:
         dataset = data['datasetname']
     else:
-        sys.stderr.write("(WARN) Required field 'datasetname' not found "
-                         "in data: %r\n" % data)
+        sys.stderr.write("(WARN) Required field 'datasetname' not found"
+                         " in data: %r\n" % data)
         return False
     if not isinstance(dataset, (str, unicode)):
-        sys.stderr.write("(WARN) Invalid type of 'datasetname' field: "
-                         "expected string, got %s.\n"
-                         "(==) Data: "
-                         "%r\n" % (dataset.__class__.__name__, data))
+        sys.stderr.write("(WARN) Invalid type of 'datasetname' field:"
+                         " expected string, got %s.\n"
+                         "(==) Data:"
+                         " %r\n" % (dataset.__class__.__name__, data))
         return False
     if len(dataset) == 0:
-        sys.stderr.write("(WARN) Required field 'datasetname' is empty "
-                         "in data: %r\n" % data)
+        sys.stderr.write("(WARN) Required field 'datasetname' is empty"
+                         " in data: %r\n" % data)
     return re.sub('_tid(.)+', '', dataset)
 
 
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -25,12 +25,12 @@ except Exception, err:
 
 ami_client = None
 # Field names in terms of AMI and ES schemes.
-PHYS_VALUES = [{"ami": "genFiltEff", "es": "gen_filt_eff"},
-               {"ami": "crossSection", "es": "cross_section"},
-               {"ami": "crossSectionRef", "es": "cross_section_ref"},
-               {"ami": "kFactor", "es": "k_factor"},
-               {"ami": "processGroup", "es": "process_group"},
-               {"ami": "mePDF", "es": "me_pdf"},
+PHYS_VALUES = [{'ami': 'genFiltEff', 'es': 'gen_filt_eff'},
+               {'ami': 'crossSection', 'es': 'cross_section'},
+               {'ami': 'crossSectionRef', 'es': 'cross_section_ref'},
+               {'ami': 'kFactor', 'es': 'k_factor'},
+               {'ami': 'processGroup', 'es': 'process_group'},
+               {'ami': 'mePDF', 'es': 'me_pdf'},
                ]
 FILTER = ['AOD', 'EVNT', 'HITS']
 
@@ -199,8 +199,8 @@ def change_key_names(data):
     :rtype: dict
     """
     for item in PHYS_VALUES:
-        if item["ami"] in data:
-            data[item["es"]] = data.pop(item["ami"])
+        if item['ami'] in data:
+            data[item['es']] = data.pop(item['ami'])
     return data
 
 
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -24,6 +24,7 @@ except Exception, err:
     sys.exit(1)
 
 ami_client = None
+# Field names in terms of AMI and ES schemes.
 PHYS_VALUES = [{"ami": "genFiltEff", "es": "gen_filt_eff"},
                {"ami": "crossSection", "es": "cross_section"},
                {"ami": "crossSectionRef", "es": "cross_section_ref"},
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -98,7 +98,8 @@ def init_ami_client(userkey='', usercert=''):
         ami_client.execute('ListCommands')
     except Exception as e:
         sys.stderr.write("(WARN) Failed to perform test command "
-                         "ListCommands. Exception: %s\n" % str(e))
+                         "ListCommands. Are you sure you have"
+                         " a valid certificate?\nException: %s\n" % str(e))
 
 
 def get_ami_client():
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -230,6 +230,9 @@ def container_name(data):
                          "(==) Data: "
                          "%r\n" % (dataset.__class__.__name__, data))
         return False
+    if len(dataset) == 0:
+        sys.stderr.write("(WARN) Required field 'datasetname' is empty "
+                         "in data: %r\n" % data)
     return re.sub('_tid(.)+', '', dataset)
 
 
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -121,19 +121,14 @@ def process(stage, message):
     # or not set at all.
     if update or not formats:
         amiPhysValues(data)
+        change_key_names(data)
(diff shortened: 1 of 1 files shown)


Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -118,7 +118,16 @@ def amiPhysValues(data):
...
     """
-    dataset = data['datasetname']
+    if not isinstance(data, dict):
+        sys.stderr.write("(WARN) Cannot update non-dict data:%r\n" % data)
+        return False
+    dataset = data.get('datasetname')
+    if dataset is None:
+        sys.stderr.write("(WARN) No datasetname found in data:%r\n" % data)
+        return False
+    elif not isinstance(dataset, str) and not isinstance(dataset, unicode):
```
mgolosova: *Just a comment:* `isinstance()` can work with a number of types/classes, and returns `True` if at least one class matches:
```
elif not isinstance(dataset, (str, unicode)):
    ...
```

Evildoor: Hm. If my memory serves, I looked for something like this, and found nothing, which means I've searched poorly. Thanks for enlightening me. Updated.

Evildoor:
Force-pushed to relocate data type check into `process()`, where `data.get(...)` is performed before `amiPhysValues(data)` (in which the check was initially implemented).

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -153,9 +158,9 @@ def amiPhysValues(data):
         sys.stderr.write("(WARN) Could not process dataset '%s' due to "
                          "SSLError:%r\n" % (data['datasetname'], e))
         return False
-    except Exception:
-        sys.stderr.write("(WARN) No values found in AMI for dataset '%s'\n"
-                         % data['datasetname'])
+    except Exception as e:
+        sys.stderr.write("(WARN) Exception while processing dataset '%s':"
+                         "%r\n" % (data['datasetname'], e))
```
mgolosova: Is it still "WARN", maybe it is more like an "ERROR"? I don\`t know for myself, in which situations we reach this point, so it is just a question.

"Exception while processing" sounds a bit... not friendly? I mean, error messages are supposed to be user-oriented and explain what happened, not simply state it. Maybe "Failed to process dataset", or "Dataset processing failure"?

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -180,8 +183,8 @@ def remove_tid(dataset):
     """ Remove TaskID (_tidXX) part from dataset name.
 
     As AMI GetPhysicsParamsForDataset works with containers,
-    we construct the container name from each dataset name,
-    removing the _tid{...} part
+    the container name is extracted from each dataset name by
+    removing the _tid{...} part.
```
mgolosova: Well, talking about this function... Information about GetPhysicsParamsForDataset does not seem to be a relevant description at all, however reworded. Maybe it deserves renaming?
Something like this:
```
def container_name(dataset):
    """ Construct contained name from dataset name.

    The container name is extracted from each dataset name by
    removing the _tid{...} part.

    :param ...: ...
    ...
    """
```

mgolosova: And, I am not sure about curly braces here: `...removing the _tid{...} part.`

I mean, why curly? In regular expressions they are used to set number of repetitions; it may also refer to bash [brace expansion](https://www.linuxjournal.com/content/bash-brace-expansion)... and only after realizing it is none of the above I would start thinking about some custom meaning.
Maybe replace them with asterisk? It looks like more or less standard wildcard symbol: `...removing '_tid*' part`. Or, simply leave omission points: `...removing '_tid...' part`.

Not a big deal, of course, but since you raise docs improvement question for this stage...;)

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -95,7 +95,16 @@ def get_ami_client():
...
-    """ Single message processing. """
+    """ Process a message.
+
+    Implementation of :py:meth:`.ProcessorStage.process` for hooking
+    the stage into DKB workflow.
+
+    :param stage: stage instance
+    :type stage: pyDKB.dataflow.stage.ProcessorStage
+    :param msg: input message with data
+    :type msg: pyDKB.dataflow.communication.messages.JSONMessage
```
mgolosova: Should be `message`, not `msg`.

Evildoor:
> [e060f14](https://github.com/PanDAWMS/dkb/commit/e060f143f865632d69b74a523c3c9d6a0b1ae75b) (Add SSLError detection)
> 
> I`d like to check how it looks now; I can of course guess what to do to get there, but if you could provide some examples in which situations this exception is raised it would be really helpful ;)

Sure thing.

```
[vaulov@aiatlas171 095_datasetInfoAMI]$ ./amiDatasets.py --usercert inp --userkey inp -s f -d s inp
(WARN) pyDKB.dataflow.cds failed (No module named invenio_client.contrib)
(INFO) (ProcessorStage) Configuration parameters:
(INFO) (ProcessorStage)   hdfs         : 'False'
(INFO) (ProcessorStage)   dest         : 's'
(INFO) (ProcessorStage)   input_dir    : '.'
(INFO) (ProcessorStage)   skip_process : 'False'
(INFO) (ProcessorStage)   eom          : '\n'
(INFO) (ProcessorStage)   output_dir   : 'out'
(INFO) (ProcessorStage)   source       : 'f'
(INFO) (ProcessorStage)   eop          : ''
(INFO) (ProcessorStage)   mode         : 'f'
(INFO) (ProcessorStage)   userkey      : 'inp'
(INFO) (ProcessorStage)   usercert     : 'inp'
(INFO) (ProcessorStage)   config       : 'None'
(INFO) (ProcessorStage)   input_files  : '['inp']'
(INFO) (ProcessorStage) Starting stage execution.
(WARN) Could not process dataset 'mc15_13TeV.342970.aMcAtNloPythia8EvtGen_A14NNPDF23LO_HZZvvqq900w15.recon.AOD.e4501_s2726_r7772_tid08345155_00' due to SSLError:SSLError(336445449, u'[SSL] PEM lib (_ssl.c:2574)')
{"datasetname": "mc15_13TeV.342970.aMcAtNloPythia8EvtGen_A14NNPDF23LO_HZZvvqq900w15.recon.AOD.e4501_s2726_r7772_tid08345155_00", "deleted": true, "bytes": -1, "_type": "output_dataset", "data_format": ["AOD"], "_parent": 8345155, "_id": "mc15_13TeV.342970.aMcAtNloPythia8EvtGen_A14NNPDF23LO_HZZvvqq900w15.recon.AOD.e4501_s2726_r7772_tid08345155_00"}
(WARN) Could not process dataset 'mc16_13TeV.395366.MGPy8EG_A14N23LO_CC_directZlHlWv_550.recon.AOD.e6715_e5984_a875_r10201_tid14196092_00' due to SSLError:SSLError(336445449, u'[SSL] PEM lib (_ssl.c:2574)')
{"datasetname": "mc16_13TeV.395366.MGPy8EG_A14N23LO_CC_directZlHlWv_550.recon.AOD.e6715_e5984_a875_r10201_tid14196092_00", "deleted": true, "bytes": -1, "_type": "output_dataset", "data_format": ["AOD"], "_parent": 14196092, "_id": "mc16_13TeV.395366.MGPy8EG_A14N23LO_CC_directZlHlWv_550.recon.AOD.e6715_e5984_a875_r10201_tid14196092_00"}
(WARN) No datasetname found in data:{}
{}
(WARN) Could not process dataset 'crap' due to SSLError:SSLError(336445449, u'[SSL] PEM lib (_ssl.c:2574)')
{"datasetname": "crap"}
(INFO) (ProcessorStage) Stopping stage.
```

Cause of problems: `inp` is not a correct key/certificate.

A few comments:

1. This commit should probably be removed as the following 4be275f catches the error just fine:

```
(WARN) Exception while processing dataset 'mc15_13TeV.342970.aMcAtNloPythia8EvtGen_A14NNPDF23LO_HZZvvqq900w15.recon.AOD.e4501_s2726_r7772_tid08345155_00':SSLError(336445449, u'[SSL] PEM lib (_ssl.c:2574)')
```

2. The error message is not very informative and is displayed for each dataset, while in this case it should be (by logic, since we know that the problem arises due to incorrect credentials) detected once. This is not ideal and does not fully address the trello card, but this is better than previous situation where it was reported as "no data in AMI". Further attempts to improve this issue can be performed, in this PR as well as outside of it.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: Fixed, thank you.

mgolosova:
>> e060f14 (Add SSLError detection)
>> 
>> I`d like to check how it looks now; I can of course guess what to do to get there, but if you could provide some examples in which situations this exception is raised it would be really helpful ;)
>
> Sure thing.
> <...>

Thank you, although I only asked about situations in which SSLError is raised, not about log examples themselves ;)

> This commit should probably be removed as the following 4be275f catches the error just fine

If SSLError appears *only* in situations where there\`s something wrong with user certificate, then it may worth falling back to "skip" policy or even interrupt execution (as if we failed to initialize client).

> The error message is not very informative and is displayed for each dataset, while in this case it should be (by logic, since we know that the problem arises due to incorrect credentials, and it's a problem of authentication, not of individual dataset) detected once.

Right, the error is not related to a specific dataset, but some datasets may be processed without valid cert, right? So if we do not interrupt processing, I think it still makes sense to log the error every time it occurs. This error is the reason why this exact message can not be processed properly, and when we find out that AMI metadata for dataset XXX are missed I will start looking into this probleb from everything related to the given dataset from log files. And the single message saying "SSL error occured, please check your certificate" most likely will appear somewhere high above and I won\`t see it immediately.

---

I would also suggest to replace `SSLError(336445449, u'[SSL] PEM lib (_ssl.c:2574)')` with something more friendly, like "SSL library error (errno: ERROR_CODE)". And since sometimes there *may* be something more sensible in the original message (e.g. in case of the SSL protocol error), maybe replace only messages containing "_ssl.c:...", I don\`t know.

Make the message really helpful would be quite tricky: somehow the error that you got leads here:
https://github.com/python/cpython/blob/3.4/Modules/_ssl.c#L2574

*(Don\`t ask me why v.3.4: I don\`t know, but in the 2.7 branch there\`s nothing close to what we\`re looking for at line [2574](https://github.com/python/cpython/blob/master/Modules/_ssl.c#L2574))*

My guess is that the error code produced in this situation (336445449), as well as the error text ("PEM lib"), comes directly from the SSL library, while more self-explaining message from `_ssl.c` gets lost on its way to Python. If so, an attempt of reverse engineering of all possible error situations will simply require too much time... almost for nothing.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -95,6 +95,9 @@ def process(stage, message):
     data = message.content()
     # 'data_format' field contains a list of strings,
     # e.g. ['DAOD_SUSY5', 'DAOD']
+    if not isinstance(data, dict):
```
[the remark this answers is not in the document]
Evildoor: Done, thank you.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: Agreed on both points. Done.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -118,7 +121,13 @@ def amiPhysValues(data):
...
     - cross_section_ref
     """
-    dataset = data['datasetname']
+    dataset = data.get('datasetname')
+    if dataset is None:
+        sys.stderr.write("(WARN) No datasetname found in data:%r\n" % data)
+        return False
+    elif not isinstance(dataset, (str, unicode)):
+        sys.stderr.write("(WARN) Datasetname is not a string:%r\n" % dataset)
+        return False
```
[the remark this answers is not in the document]
Evildoor: Done: fec92d1.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -118,7 +121,13 @@ def amiPhysValues(data):
     - k_factor
     - cross_section_ref
     """
-    dataset = data['datasetname']
+    dataset = data.get('datasetname')
+    if dataset is None:
+        sys.stderr.write("(WARN) No datasetname found in data:%r\n" % data)
+        return False
+    elif not isinstance(dataset, (str, unicode)):
```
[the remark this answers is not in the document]
Evildoor: fec92d1

1. Similar to previous comment, moved this check into `container_name()` (former `remove_tid()`).
2. Changed `data.get()` to check that data contains 'datasetname' with the following assignment. This way missing 'datasetname' and explicit `None` will not be treated in a similar way.
3. Done.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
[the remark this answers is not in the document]
mgolosova: There\`s this difficulty with existing code refactoring/improvements: it is very tempting to just rewrite it all and say "code rewritten". But all the changes have to be done and logged step by step, so that it was easy to understand where things were just refactored to improve code readability/extendability/usability/..., and where something was changed in the actual code behaviour (reliability improved, bugs fixed, etc). For one simple reason: if something went wrong after this reworking, it may be anything: typo, copy-and-paste mistake, logical error, forgotten renaming... and it is better to have set of smaller commits where you know what deserves special attention -- then a single commit with the whole variety of (possible) mistakes.

The mentioned commit (fec92d1), as I see it, does the following things:
* changes one function behaviour (which indeed invites renaming and description update);
* introduces additional checks.

However the summary (and even the whole message) does not reallly explain what and why happened. "Rework" means "something changed", fine. But then the description simply contains a list of changes that gives no clue why it happened and why all these changes are gathered in one commit. I mean, what was the general goal of all these changes?

If split things into two commits, first one may still be named as "rework", but should be accompanied with some explaination -- why new version is better? Something like "I didn\`t like the function name" would also work (except, it doesn\`t explain why the func behaviour has to be changed); but it must explain why author decided change something and what he/she tried to achieve by this.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -84,15 +94,34 @@ def init_ami_client(userkey='', usercert=''):
...
+    :type stage: pyDKB.dataflow.stage.ProcessorStage
+    :param message: input message with data
+    :type message: pyDKB.dataflow.communication.messages.JSONMessage
+
+    :return: False (message contains incorrect data) or True (otherwise)
+    :rtype: bool
+    """
     data = message.content()
+    if not isinstance(data, dict):
+        sys.stderr.write("(WARN) Cannot update non-dict data:%r\n" % data)
```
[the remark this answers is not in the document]
Evildoor: Fixed 4 instances, including this one.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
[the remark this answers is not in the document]
Evildoor: Hm. I had spent some time pondering how to correctly handle this situation before making fec92d1, but I agree with your comment, which means that not all problems were solved with it. Reworked the commit into three ones (not two, to further separate change of behavior and stylistic fixes of the description):
12d5ee9
4f88409
bf15935

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -172,12 +172,14 @@ def change_key_names(data):
...
 
-    As AMI GetPhysicsParamsForDataset works with containers,
-    we construct the container name from each dataset name,
-    removing the _tid{...} part
+    The container name is extracted from dataset name by removing
+    the '_tid...' part.
 
-    :param dataset: dataset information
-    :return: dataset name without _tid => container name
+    :param data: dataset information
```
mgolosova: Parameter name has actually changed an another commit (12d5ee9).
It is fixed here (4f88409), among other improvements of the docstrings, but you still might want to amend previous commit or reword this one, mentioning the parameter name fixing. Not a big deal, maybe, but it catches the eye since you have a pretty detailed list of changes and this one is not mentioned.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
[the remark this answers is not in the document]
mgolosova: Looks good now, thank you!

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -105,6 +114,9 @@ def process(stage, message):
     :type stage: pyDKB.dataflow.stage.ProcessorStage
     :param message: input message with data
     :type message: pyDKB.dataflow.communication.messages.JSONMessage
+
+    :return: False (message contains incorrect data) or True (otherwise)
```
mgolosova: This line makes me think that `True` means "input data are correct", not "processing finished in a routine mode, without any (unexpected) problems", while `True`/`False` are meant to mean "processed successfully"/"failed to process message" (for whatever reason). Maybe for now it looks like the only reason to say "we failed, skip the message" is to have invalid input data, but I guess there is always a room for different unpleasant situations ;)
Can it be reworded somehow?

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: I missed this, thanks for pointing out. Moved the change to eddad82 (former 12d5ee9).

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -170,16 +171,28 @@ def change_key_names(data):
...
     """
+    if 'datasetname' in data:
+        dataset = data['datasetname']
+    else:
+        sys.stderr.write("(WARN) Required field 'datasetname' not found "
+                         "in data:%r\n" % data)
+        return False
+    if not isinstance(dataset, (str, unicode)):
+        sys.stderr.write("(WARN) Dataset name in data %r is not a string:"
+                         "%r\n" % (data, dataset))
```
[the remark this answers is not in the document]
Evildoor: Agree, changed the format.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: Done. Also changed the similar case catching SSL errors, for the same reason. Due to the changes the two cases (SSL errors and all other errors) perform the same actions, but this will change if I manage to transform the SSL errors into more meaningful form.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py

> Maybe for now it looks like the only reason to say "we failed, skip the message" is to have invalid input data, but I guess there is always a room for different unpleasant situations ;)

Actually, we already have different situations - `amiPhysValues()` returns `True` or `False` depending on its results (however, at the moment it returns `False` when catching exception and when simply no data was found in AMI), so it may be logical to output this return instead of `True` at the end of `process()`. Will look into it.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -123,7 +123,7 @@ def amiPhysValues(data):
     ami_client = get_ami_client()
     try:
         res = ami_client.execute(['GetPhysicsParamsForDataset',
-                                  "--logicalDatasetName=%s" % container],
+                                  '--logicalDatasetName=%s' % container],
```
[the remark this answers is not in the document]

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: After some consideration I decided not to change anything further because the return value of `amiPhysValues()` is not used anywhere. When (and if) it will be used, we will decide on this point. Who knows, maybe we'll need to know whether the data was updated - in this case situations "something went wrong and nothing was done" and "no data was found in AMI" will indeed be the same for us.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: Added traceback. Now this is more like it.

```
(ERROR) Failed to process dataset 'mc15_13TeV.342970.aMcAtNloPythia8EvtGen_A14NNPDF23LO_HZZvvqq900w15.recon.AOD.e4501_s2726_r7772_tid08345155_00': SSLError(336445449, u'[SSL] PEM lib (_ssl.c:2574)')
(DEBUG) Traceback start.
  File "./amiDatasets.py", line 165, in amiPhysValues
    format='json')
  File "/usr/lib/python2.7/site-packages/pyAMI/client.py", line 215, in execute
    self.httpClient.connect(endpoint)
  File "/usr/lib/python2.7/site-packages/pyAMI/httpclient.py", line 105, in connect
    context = context
  File "/usr/lib64/python2.7/httplib.py", line 1244, in __init__
    context.load_cert_chain(cert_file, key_file)
(DEBUG) Traceback end.
```

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
mgolosova: @Evildoor, 
Thank you.

---

I mark the conversation as "resolved" (since it *is* resolved), but to make things clear: my point was about `process()` function, not about *any* function with boolean return values. Any arbitrary taken function is allowed to have its own interpretation of boolean return values -- but `process()`, being an *implementation* of an abstract method, must follow the definition given in the library:
https://github.com/PanDAWMS/dkb/blob/66f68d10f80cefb39e197d3b405cba5233d15007/Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py#L278-L288

If it does not follow the definition, but is used as the method implementation, we will eventually face an unpredictable behaviour of a stage -- which we try to avoid using the common library.
Now it is only a matter of docstring, but following the inaccurate documentation the function might later be changed to something unhealthy.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
mgolosova: Ok, thank you.

By the way, there\`s an implementation of traceback logging in `pyDKB` in [AbstractStage](https://github.com/PanDAWMS/dkb/blob/master/Utils/Dataflow/pyDKB/dataflow/stage/AbstractStage.py#L254) -- can\`t it be used here?

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -135,36 +178,61 @@ def amiPhysValues(data):
...
+    if 'datasetname' in data:
+        dataset = data['datasetname']
+    else:
+        sys.stderr.write("(WARN) Required field 'datasetname' not found "
+                         "in data: %r\n" % data)
+        return False
+    if not isinstance(dataset, (str, unicode)):
+        sys.stderr.write("(WARN) Invalid type of 'datasetname' field: "
+                         "expected string, got %s.\n"
+                         "(==) Data containing the field: "
```
mgolosova: This information ("containing the field") looks a bit excessive: if we have a value of invalid type in some field, and then output some data as the additional information for this error, it is not difficult to guess that displayed data is where we tried to find our field. Maybe reword it somehow, depending on what it should provide? My variants:
* `"(==) Data: %r\n" % data` *(to output all data)*
* `"(==) Invalid value: %r\n" % dataset` *(to output invalid field value)*

Or maybe some other variant. What do you think?

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: Hm. It can be used, but I'm not sure that it should be.

`output_error()` is a method of `AbstractStage`, which in our case is represented by `stage` in `process()`. `amyPhysValues()` is called in `process()` with a single parameter `data`. This means that `amiPhysValues()` has no option to call `output_error()` in its current state, and `sys.exc_info()` call from `process()` does not catch the info from `amiPhysValues()`. Possibilities exist to make it work, so far I've thought of the following:

1. Pass `stage` as an argument to `amiPhysValues()`.
2. Return something (result of `sys.exc_info()`, for example) into `process` to work with.
3. Do 1 or 2, but with global variable or something else instead of argument/return.
4. Make (and use in this stage) a special class (`AMIStage` or something) based on `ProcessorStage` and turn `amiPhysValues()` into its method.

Neither of these look too appealing to me, more like unnecessary complications for the sake of using specific method to report an error. Perhaps 2 is the least "illogical", but we may need return status for something else.

Long story short, I think we should either process errors on the spot (and leave everything as-is), or, if we really need it, consider the implementation of some sort of general mechanism to return all errors (and not just SSLError, and not just in this stage) into process/stage and work with them there.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: Sure, just "Data:" should do fine. Amended and force-pushed.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
mgolosova: > Possibilities exist to make it work, so far I've thought of the following:
> 1. Pass stage as an argument to amiPhysValues().
> 1. Return something (result of sys.exc_info(), for example) into process to work with.
> 1. Do 1 or 2, but with global variable or something else instead of argument/return.
> 1. Make (and use in this stage) a special class (AMIStage or something) based on ProcessorStage and turn amiPhysValues() into its method.

All the suggestions are equally bad, for my liking. Except for the last -- it could be a way to organize any of the stages, in fact. I mean -- to wrap the whole stage into an object with functions turned into methods, instead of `main()` plus the bunch of standalone functions. Doesn\`t look bad, it\`s just what we\`ve never done before.

However, there are many other possibilities -- and the most straightforward would be to simply allow the exception to pop from the `amiPhysValues()` and process it on the `process()` level. Why not? Besides, it would also allow more or less accurate handling of any other exception that might appear in this section:
https://github.com/PanDAWMS/dkb/blob/574ca5026f70c96783798712d787551745768c33/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py#L141-L143
It is, by the way, indeed what you\`re talking about:
> some sort of general mechanism to return all errors (and not just SSLError, and not just in this stage) into process/stage and work with them there.

-- and it is already invented for us.

Other option would be to look at the situation from the different point: not "how to apply what we have", but "how to make it comfortable". My thoughts are: if there *is* an implementation of what you need (and it includes some pretty stuff like proper prefixes for all the lines etc.), but hidden so deep that makes it difficult to use without acrobatic exercises -- it is a direct indicator that something\`s not right and worth improvement. Maybe not right now, but still deserves attention. So, from this point of view, the question is: how would you like it to be? what way you\`d feel natural to use the library implementation?

---

I\`m not saying anything of this should be done to make me accept this PR -- it can be done later by any of us, just with the same result; it\`s just "since we are here, let\`s not pass by this possibility of improvement and make life easier for those in the same situation -- should it be a month, or two (or more) from now".

I mean, it hit me because I personally spent some time figuring how to output the traceback in the common context of logging; here you did the same; next time it will be some one else, or one of us copy-and-pasting the solution invented before. Which is simply not right, and looks like an opportunity to learn how to do things better and avoid such situations in the future, when one can not use results of someone else's work and has to do it by him/herself.

Evildoor: > However, there are many other possibilities -- and the most straightforward would be to simply allow the exception to pop from the amiPhysValues() and process it on the process() level.

Ah, you meant to use `raise`, right? Have not thought about that, and it seems to solve the problem in a proper way. Done: 8d1a11d.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -24,6 +24,7 @@
     sys.exit(1)
 
 ami_client = None
+# List for adapting names of AMI values to ES.
```
mgolosova: Does adding this comment have anything to do with "Separate request to AMI and adaptation to ES."? 
And maybe it would be better to reword it a little, for example: "Field names in terms of AMI and ES schemes"? At this point it is not very clear what is "AMI values" and what is "adapting"; after reading the code (at least the part where this list is used) one will understand, but comments are usually help to understand things without reading code (or to understand the code itself).

Same question about changing the `change_key_names()` description (and I am not sure if these two changes belong to the same commit, by the way).

Evildoor: True, split the commit into several ones, with some rewordings.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -137,10 +137,12 @@ def amiPhysValues(data):
                     data[p_name] = p_val
                     p_name, p_val = None, None
                     continue
-        return change_key_names(data)
+        change_key_names(data)
+        return True
```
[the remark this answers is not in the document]
Evildoor: Done (in form of removal of try/except per suggestion below).

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -179,14 +187,8 @@ def amiPhysValues(data):
...
-    except SSLError as e:
-        sys.stderr.write("(ERROR) Failed to process dataset '%s': "
-                         "%r\n" % (data['datasetname'], e))
-        return False
-    except Exception as e:
-        sys.stderr.write("(ERROR) Failed to process dataset '%s': "
-                         "%r\n" % (data['datasetname'], e))
-        return False
+    except Exception:
+        raise
```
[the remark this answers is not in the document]

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -81,18 +91,42 @@ def init_ami_client(userkey='', usercert=''):
                          "proxy.\n")
         raise DataflowException("Failed to initialise AMI client: certificate "
                                 "not provided or not found.")
+    try:
+        ami_client.execute('ListCommands')
+    except Exception as e:
+        sys.stderr.write("(WARN) Failed to perform test command "
+                         "ListCommands. Exception: %s\n" % str(e))
```
mgolosova: I have a feeling that we have discussed this question before, which is captured in the commit log of 05283d63979d6635e16ebf75cabd58d90931ecb7:
```
Information about some datasets is available from AMI without authentication -
display warning instead of stopping the stage if the test command fails, to
allow such datasets to be processed.
```
, so I'm sorry for not being self-consistent... but now here's what I think.

This function (`init_ami_client()`) raises `DataflowException` whenever there's some sign of a failure to establish the connection (https://github.com/PanDAWMS/dkb/pull/282/files#diff-02fcd172861b2b59cff44046a135be30R79-R93):
* no `pyAMI` module (pretty understood);
* any `Exception` thrown by `pyAMI` (although we wil get here only if... I don't know, pass non-string values as `{key|cert}_file`);
* no certificate provided (although `pyAMI` would fall back to password authentication, still providing info for those datasets that do not require special authorization).

First two cases are understood: we can not init the client at all. But the third is much the same as the case of invalid certificate; so why would we fail in one case but try to proceed in another?

Evildoor: If I understand this correctly, no we are changing our mind and aborting the procession if something is wrong with credentials. This is fine by me, made the changes.

Evildoor:
Addressed the comments.

> When empty dataset name passed, stage provides no feedback to `STDERR` like it would do in case of missed or "invalid" name (of wrong type, or just unknown to AMI). Before this PR there was just same error message: `(WARN) No values found in AMI for dataset ''`
> Guess here's the reason:
> 
> https://github.com/PanDAWMS/dkb/blob/0801c506bb36cc8cd02cbe3208b0d82543ab2ceb/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py#L165-L167

While this is true, that output is missing due to this check treating False and empty string in the same way, I introduced a new check into `container_name()` rather than changing this to something like `if container == False` - empty string is treated differently somewhere down the line and produces unclear error message instead of expected "No values found in AMI for dataset ''".

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -50,12 +50,15 @@ def main(argv):
...
     stage.configure(argv)
     stage.process = process
-    if stage.ARGS.userkey and stage.ARGS.usercert:
+    try:
         init_ami_client(stage.ARGS.userkey, stage.ARGS.usercert)
-    exit_code = stage.run()
+    except Exception:
+        exit_code = 1
+    else:
+        exit_code = stage.run()
```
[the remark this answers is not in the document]
Evildoor: While the change, indeed, breaks `--skip` usage and should be fixed - thanks for pointing this out - I'm not sure that `if` should be left in place as-is.

Commits related to the issue:

1. dc13dc7 - here the `if` is introduced, precisely for working with `--skip`.
2. ce05ca8 - this commit unintentionally introduces the problem: message 'Result' for each dataset starting with the second when no credentials were supplied. This happens because calling `get_ami_client()` in `amiPhysValues()` leads to exception: before the commit the call is performed normally and the exception stops the program execution, while after the commit it is caught by `try-except` and the execution continues.
3. ce3841d - here the problem is solved by always trying to perform the initialization, regardless of (un)supplied credentials, and stopping the execution if something is wrong.

Now, while the initial code (one in `master`) works correctly - it tries to process one dataset and stops due to missing credentials - it does so due to the fact that this first processing throws an unhandled exception. While such logic can be left as-is in some cases, I believe that this one is different - after all, one of the "symptoms" that led to this whole PR was the code talking about "no data in AMI" for each dataset when credentials were missing.

Therefore, I believe that the logic "if certificate and key are supplied, initialize the client" should be replaced with something like "if `--skip` is absent, initialize the client". This way, missing credentials will be properly caught and reported once, while `--skip` will function properly. Will look into it further to determine how the history should be rewritten in #312 to be easier to understand.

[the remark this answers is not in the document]
mgolosova: @Evildoor,
> Now, while the initial code (one in master) works correctly - it tries to process one dataset and stops due to missing credentials - it does so due to the fact that this first processing throws an unhandled exception. 

1. I wouldn't call it an ***unhandled*** exception. Technically, "unhandled" is the one that reaches the outermost context of the executed program without being caught. It looks like this:
    ```
    (INFO) Start program execution.
    Exception: This is an unhandled exception!
    ```
    But there are situations when exceptions indicate that program should be stopped, right? Then it may look like this:
    ```
    (INFO) Start program execution.
    (ERROR) Some error occurred!
    (DEBUG) Traceback (most recent call last):
    (DEBUG)   File "<stdin>", line 1, in <module>
    (DEBUG) Exception: Some error occured!
    (INFO) Stop program execution.
    ```
    -- the exception was caught, analyzed, and since it is not possible to go on with execution -- user was provided with debug information about the situation, and the program was gracefully stopped.

    Here's what we have in `master`:
    ```
    2020-01-17 11:23:22 (INFO) (ProcessorStage) Starting stage execution.
    (ERROR) Login authentication mode is not supported. Please provide user certificate or create proxy.
    2020-01-17 11:23:22 (ERROR) (ProcessorStage) Failed to initialise AMI client: certificate not provided or not found.
    2020-01-17 11:23:22 (DEBUG) (ProcessorStage) Traceback (most recent call last):
    (==)   File "/home/dkb/dkb-dev.git/Utils/Dataflow/095_datasetInfoAMI/../pyDKB/dataflow/stage/ProcessorStage.py", line 231, in run
    (==)     if msg and process(self, msg):
    (==)   File "./amiDatasets.py", line 107, in process
    (==)     amiPhysValues(data)
    (==)   File "./amiDatasets.py", line 123, in amiPhysValues
    (==)     ami_client = get_ami_client()
    (==)   File "./amiDatasets.py", line 89, in get_ami_client
    (==)     init_ami_client()
    (==)   File "./amiDatasets.py", line 82, in init_ami_client
    (==)     raise DataflowException("Failed to initialise AMI client: certificate "
    (==) DataflowException: Failed to initialise AMI client: certificate not provided or not found.
    2020-01-17 11:23:22 (INFO) (ProcessorStage) Stopping stage.
    ```
    Looks pretty handled to me, m?

2. It is "unhandled" from the `process()` point of view -- but it only means one thing: it should be properly handled. And, no matter what -- I insist that "proper handling" in this situation would be to gracefully stop the program execution. "Default" handling policy was "not to handle" == "propagate all exceptions" and it worked properly. When `try/except clause` was added, it changed. And it is exactly what I wanted you to analyze: what has changed, why it lead to the new bug, why previous version worked properly, and -- the most important part, -- are there other but "`--skip` case" situations, when it may possibly lead to an unexpected behaviour?

3. > ce3841d - here the problem is solved by always trying to perform the initialization, regardless of (un)supplied credentials, and stopping the execution if something is wrong.

    ...and introduces another problem: `--skip` option does not work properly.

    What I am saying, is that in situations like this: `code works` -> `change X introduces bug A` -> `fixing bug A introduces bug B` -- the "bug B" point is a good moment to stop making changes and take a closer look at the code. Because this situation means that author of the changes does not fully understand the logic of the code, and next fix may as well break something else... and who can guarantee that new issue will be detected immediately? I mean, if we had unit tests -- that's easier, just go-go-go, fix-fix-fix till all the tests say OK (although it would also be a good idea to stop for a moment and think about what's going on). But since we don't have them, there's a good chance that new issue stays undetected till someone steps on it -- and hopefully it will happen a controlled test environment, not in "production" version where it may do some damage.

4. > While such logic (<stop execution due to an unhandled exception>) can be left as-is in some cases, I believe that this one is different - after all, one of the "symptoms" that led to this whole PR was the code talking about "no data in AMI" for each dataset when credentials were missing.

    I am sorry, but (leaving aside that the logic is a bit different, see p.1), I don't see how from "program incorrectly handles situation with missed credentials" follows "program shouldn't stop execution due to an exception thrown during client initialization".

5. > Therefore, I believe that the logic "if certificate and key are supplied, initialize the client" should be replaced with something like "if --skip is absent, initialize the client". This way, missing credentials will be properly caught and reported once, while `--skip` will function properly.

    Although the "therefore"-preceding part sounds very doubtful to me, what you suggest worth trying: for now we seem to have only one use-case when we don't need to initialize the client, so it's better to fail early if the client can not be initialized. However, I want you to understand: ***it does not fix the bug -- it eliminates a detected symptom***.

---

Summary:
* the more code is changed, the more the probability of inviting new bugs;
* => to fix some bug, it's better to change less than more -- and therefore it's better to change what has introduced the bug, not the rest of the code;
* to change the way of allowing `--skip` will not fix the issue, only make it less visible (*maybe* removing the only possible effect ever, but I doubt it);
* but the way `--skip`-allowing logic was implemented is not good indeed and worth improving *(which is one of the reasons why do a code review: to improve things that worth improving when it's "work in progress", not "discussed, merged and forgotten for good" ;) )*.

Evildoor:
Closing this as the changes were remade in #309 and #312.

PR #284, opened 2019-08-21
095 update: part 2

Evildoor:
Update stage 095 to improve it per request from AMI team (https://trello.com/c/x0hvd4aQ).

Notes:
- Unlike #282, samples must be reproduced after/at the end of this PR, since it affects them.
- Depending on the work on this PR and #282, this PR would probably be merged into `master` instead of `095-update`.


Evildoor:
Part 1 was successfully merged. Brough this PR up to date.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -182,9 +182,18 @@ def amiPhysValues(data):
...
     ami_client = get_ami_client()
-    res = ami_client.execute(['GetPhysicsParamsForDataset',
-                              '--logicalDatasetName=%s' % container],
-                             format='json')
+    exec_args = ['GetPhysicsParamsForDataset',
+                 '--logicalDatasetName=%s' % container]
+    if container[:4] in ['mc15', 'mc16']:
+        # Scope must be set correctly, it is 'mc15' by default.
+        if container[3] == '6':
+            exec_args.append('-scope=mc16')
```
mgolosova, line 190: This looks very much like a dirty, hard-coded crutch.

* Too much magic: `[:4]`, `['mc15', 'mc16']`, `[3]`, `'6'`:
  * not clear, what does this list represent;
  * not clear, when to extend the list and when add `elif` clause;
  * not clear, why exactly 4 characters *(yes, the list contains only two 4-chars items, but no one knows it in advance)*;
  * not clear, why it's important that "the 3rd character is '6'" *(yes, it is in fact a check for "container name starting with `mc16`" -- but it is to be deducted)*;
  * very easy to break everything trying to change something.
* When AMI all of a sudden change the default scope *(by the way, haven't they yet?)*, current version will be incorrect.

Also, this code says that we only know one thing: "if the container name starts with `mc16` -- its scope is `mc16`". Maybe there is some more general rule about "what is a *scope* and how to get it from the container name", like the one we have [for the Rucio scope](https://github.com/PanDAWMS/dkb/blob/master/Utils/Dataflow/091_datasetsRucio/datasets_processing.py#L289-L307)? Will look a bit more complicated, but if we start talking about "scope" in terms of AMI -- would be great to know for sure what we're talking about and not return to this question ever again.

Evildoor: Fair points.

I looked through AMI docs and various other sources of info, found no mentions of specific, AMI-related meaning of scope, in every case scope (apart from that letter from AMI team) pretty much boils down to what's implemented in our function you've linked. Furthermore, `GetPhysicsParamsForDataset` accepts "full" scopes (`mc16_13TeV`) just as fine as it does "short" ones (`mc16`).

Altogether, this means that we should probably move the function out of 091 and into a library of some sort (`Utils/Dataflow/lib/atlas.py`, for example), from which both 091 and 095 would be able to utilize it. Doing so (together with some minor improvements) would deal with most of the magic, but we'll still have to keep `['mc15', 'mc16']` in some form, because they come directly from the letter.

Thoughts? Objections?

mgolosova: > I looked through AMI docs and <scope definition> pretty much boils down to what's implemented in our function...

That's a good news, thank you.

> we should probably move the function out of 091 and into a library of some sort (`Utils/Dataflow/lib/atlas.py`, for example), from which both 091 and 095 would be able to utilize it. 

If you check #371 ( *which, by the way, you have reviewed ;)* )  -- the suggested location for this functionality was in `pyDKB.atlas` module. But if you think it'd be better (*how?*), we can move it to a separate library.

> still have to keep `['mc15', 'mc16']` in some form

Of course, it is just the same situation as in the mentioned above #371 with "data formats we should ignore" (a0adae12). My point is that this list of values needs a name.

Evildoor:
Had to do an additional force-push due to accidental mistake with newlines.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: All done. Please, take a look again.

Evildoor:
> 1. **Formal moment**: since the library code is updated, its version should be updated as well.

Will do.

> 2. **Function naming.**
>    I think it'd be better to rename `extract_scope` to someting more self-explaining. Within Stage 091 it _might_ be obvious that `extract_scope` is about datasets, but... seing `pyDKB.atlas.misc.extract_scope` -- I wouldn't know what to think about. What scope? Where is it extracted from?..
>    Yes, the suggested with `dataset_data_format` naming is not perfect -- the better solution would be to introduce a submodule 'dataset' with these functions; or, even better, a class named `Dataset` with these functions as methods. It was not done initially and, maybe, someone will has to do it later; and I do not suggest to make things the best way possible and fix my mistakes/lack of patience/concentration/time right here and now. But at least let's not make things worse than they already are ;)

Will do - I think we can rename this `misc.py` into `dataset_name.py`, both scope extraction and format extraction deal exactly with these names.

> 3. **Function content.**
>    In the context of [640d3da](https://github.com/PanDAWMS/dkb/commit/640d3da9acc333c6f71922ac932cd7944abe71d0) it is understood why to move the function from Stage 091 to the library as is: the only thing _necessary_ is to make the previously implemented functionality accessible from another stage. But as a result -- in the common library we have some "hybrid" function, doing more actions than one might expect: not only gets scope name from the dataset name, but also "purifies" the dataset name.
>    If it is a _stage_ function, it is (more or less) fine to group actions the way author likes; but for a _library_ function it looks wrong.
> ...
> 1. **Option 1.**
>    Freeze this PR, create a new one with refactoring (split the 091's `extract_scope` into two functions, move one or both of them to the `pyDKB`) and rebase this PR.

#382, but I decided to leave moving into library out of it, to put it into next PR which will also rename `misc.py` and change the latter accordingly.

***

Note to self: since the remaining comments are not about the resulting logic of the code, data samples can (and should) be
updated.

mgolosova:
@Evildoor, 
> > I think it'd be better to rename `extract_scope` to someting more self-explaining. <...> Seeing `pyDKB.atlas.misc.extract_scope` -- I wouldn't know what to think about. What scope? Where is it extracted from?..
> 
> Will do - I think we can rename this `misc.py` into `dataset_name.py`, both scope extraction and format extraction deal exactly with these names.

I don't know if it's a good thing to do. I mean, it looks like a partial solution leading to a wrong direction.
Semantically, this situation much better fits a paradigm of `Dataset` object, initialized with `name` parameter and with property methods like `name`, `scope`, `container` etc. It's a bit more of coding and more changes in stages, and we usually do not tend to operate with such "custom" structured objects instead of basic ones like string/dict/list so it's a bit more of thinking.
But accepting now this variant of "let's collect functions related to the same type of objects into a module" instead of "into a class" also looks for me as some additional work that:
* will take some time now;
* will make it harder to switch to the object-related paradigm later;
* is not required for the current task of finilizing this PR.

Thinking about sorting out "misc" functions into separate modules, classes etc does not make much sence for me as long as we have only... one, now two or maybe even three functions there. Yes, these functions look like they call for being collected into some bigger structure element. But no, I am not ready to make decisions about the whole `pyDKB.atlas` module structure now, and unfortunately I still do not feel like relying on someone in this question without thorough supervising. So I suggest to keep the current naming: `pyDKB.atlas.misc.any_atlas_related_function_with_self_explaining_name`. We can return to this question later, but its priority does not look very high...

> > 2. **Option 1.**
> >    Freeze this PR, create a new one with refactoring (split the 091's `extract_scope` into two functions, move one or both of them to the `pyDKB`) and rebase this PR.
> 
> #382, but I decided to leave "moving function into library" part out of it, to put it into next PR which will also rename `misc.py` and change the latter accordingly.

The "next PR" requires global thinking I mentioned above, so I think it'd be better to finish this one before diving into this.

Evildoor:
@mgolosova,

> So I suggest to keep the current naming: `pyDKB.atlas.misc.any_atlas_related_function_with_self_explaining_name`. We can get back to this question later, but its priority does not look very high...

Done, so there is no need...
 
> > > 1. **Option 1.**
> > >    Freeze this PR, create a new one with refactoring (split the 091's `extract_scope` into two functions, move one or both of them to the `pyDKB`) and rebase this PR.
> > 
> > 
> > #382, but I decided to leave "moving function into library" part out of it, to put it into next PR which will also rename `misc.py` and change the latter accordingly.
> 
> The "next PR" requires global thinking I mentioned above, so I think it'd be better to finish this one before diving into this.

... To do another PR.

Per discussion in #382, name normalization is now also applied in this stage.

Version and samples were updated.

Please, take a look again.

mgolosova (commented):
I have a couple of suggestions (see below).

And a comment to 87e7b99 *(no changes needed, unless you decide to rewrite the history for some other reason)*:
Maybe it's just me (and just today ;) ), but it seems to be more useful to see in commit log "what (and why) changed", not "what didn't change" (although you're right: next question for this commit would exactly be "why only one set of samples changed"). If the commit is taken by itself, it is not clear what's going on and why presence (or absence) of "mc16" data has some special meaning. A link to the commit that changed code that produces samples (in which one can read about what changed), no description of what changed (like "added metadata from AMI for mc16 datasets").
On the other hand, brief "Update data samples." with no details would also look fine: samples were out of date and this commit updates them, nothing special.

Utils/Dataflow/pyDKB/atlas/misc.py
```diff
@@ -34,3 +34,49 @@ def dataset_data_format(datasetname):
         elif N > 5:
             ds_format = splitted[4]
     return ds_format
+
+
+def extract_scope_from_dataset_name(dsn):
```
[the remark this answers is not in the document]
Evildoor: Done.

Utils/Dataflow/data4es/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -182,9 +186,14 @@ def amiPhysValues(data):
...
-                              '--logicalDatasetName=%s' % container],
-                             format='json')
+    exec_args = ['GetPhysicsParamsForDataset',
+                 '--logicalDatasetName=%s' % container]
+    scope = pyDKB.atlas.misc.extract_scope_from_dataset_name(container)
+    if scope.startswith(SCOPES):
+        exec_args.append('-scope=%s' % scope)
+    else:
+        return True
+    res = ami_client.execute(exec_args, format='json')
```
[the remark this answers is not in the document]
Evildoor: Done. Part about `get_ami_client()` also done.

Evildoor:
@mgolosova, comments were addressed. Should I update the version again? The library code was changed again, but maybe you have more suggestions that will lead to even more changes?

mgolosova:
@Evildoor,
> Should I update the version again?
> The library code was changed again...

**Yes...**
*...if you want the last change to "belong" to the version that will be merged to `master` with this PR, not to the one that will be specified in some other PR.*

But you knew this, right?

> ...but maybe you have more suggestions that will lead to even more changes?

**No, I do not...**
*...but it does not matter if I have any other brilliant ideas; if ***you*** believe that the current version is ready to be merged to the `master`, and (supposedly) want it to be merged ASAP -- then don't ask, just prepare it to be merged (the way ***you*** believe to be a right way) [1].*

---

Since it's not our first discussion about "when and how update the version" (https://github.com/PanDAWMS/dkb/pull/307#discussion_r377159897), I decided to formalize and document this question: please take a look at the new Wiki page ([How to: versioning](https://github.com/PanDAWMS/dkb/wiki/How-to:-versioning)).

For this PR, please, try to follow the current version of the documented workflow; if you have any suggestions on its improvement, please use the [dedicated Trello card](https://trello.com/c/FyFoGWST).

---

***Comment:*** earlier I wouldn't take moving of a function form an individual stage to the library as something worth increasing the minor version: in fact, the minor version was previously increased after quite extensive changes that might worth even the major version change, while this change is just a refactoring. But since this moment needs to be formalized -- then form a formal point of view it *is* a new functionality for the library. So do not analyze "how it was used before" and just follow the new, "formal" way.

---

[1] I mean, you definitely knew that changing the version is a better way than not changing the version; you either wasn't sure how to deal with more than one "version update" commit in PR (but for some reason asked about something else), or *did* make some decision (but for some reason decided not to let anyone know about it and just ask for instructions -- or a permission to follow the decision that wasn't announced).

What should we learn from here? 
1. Your reviewer thinks too much.
2. Item 1 is a very good reason to ask exactly what you *want* to ask, not some "guiding questions" -- 'cause who knows, where they will guide...

Evildoor:
@mgolosova, updated the version.

mgolosova (approved):

mgolosova:
@Evildoor,
thank you *(although you seem to forget about [this](https://github.com/PanDAWMS/dkb/pull/284#pullrequestreview-475739425) comment of mine (now related to 143b76e))*.

The PR is checked, approved, and ready to be merged .

PR #285, opened 2019-09-04
ES scheme update

mgolosova:
ES scheme update, intended to bring some order to the existing structure without significant changes affecting the `data4es` process.

What is changed:
* formatting;
* fields types added with copy-and-paste (or simply without too careful consideration).

### DIFF
--- a/Utils/Elasticsearch/mapping/README
+++ b/Utils/Elasticsearch/mapping/README
@@ -1,0 +1,19 @@
+==============
+* ES mapping *
+==============
+
+1. Field descriptions
+---------------------
+
+toths06: CPU resources used by the task. Origins from ChicagoES:
+         jobs_archive_YYYY-MM-DD -> jobs_data -> doc['hs06sec'], total
+         value for the task jobs with status 'finished' or 'failed'.
+
+toths06_finished: CPU resources the task would use in the perfect world
+                  (see description for `toths06`).
+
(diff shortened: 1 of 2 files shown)


mgolosova:
@Evildoor,
Since I just stumbled over the `energy_gev` field (where `short` type should be good enough, but can not be used due to incorrect values in the original data source), I guess I will wait for all the data to be reindexed with the new mapping before merge -- just in case that "theoretically" picked types might not be appliable to the real data.

> 9ffa221, e06dc7a, 684ace9 - these commits change field types to integer, their descriptions are similar and consist of 3 paragraphs each. However, the order of the paragraphs is different...

Right, I might started to use copy-and-paste not soon enough ) Will update (with next fix or after the reindex is finished).

Thank you!

mgolosova:
Reindexing in the ES@bamboo with new field types finished successfully, so I presume this mapping is viable.

> @mgolosova [force-pushed](https://github.com/PanDAWMS/dkb/compare/afe80c9b5fc7eafcb728beceff8f035f5fdee586..2ea060cd138eaf67f165f1f7fee91a6f5a0b8894) the `es-scheme-update`  branch from afe80c9 to 2ea060c

@Evildoor, I reworded commits related to `pr_id`, `taskid`, `run_number`; no changes in the mapping itself or commits content. Please, re-review.

mgolosova:
[ES mapping](https://trello.com/c/JZY4zkjP/73-es-mapping)

PR #292, opened 2019-10-28
019 minor fixes

Evildoor:
Fix various minor problems with the stage.

- Add/improve help messages and options to use them properly.
- Add message when a default config is used.
- Add comments describing script logic to esFormat.php.
- Fix errors discovered during work on the PR (eb7759a, ae23121).
- Correct typos, lack of numeration in README, and other minor readability issues.

https://trello.com/c/ZUvvKjKC
https://trello.com/c/wjTOlVWt


Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -5,14 +5,60 @@ function exception_error_handler($errno, $errstr, $errfile, $errline ) {
...
+$EOP_DEFAULTS = Array("stream" => chr(0), "file" => "");
+# End-of-message marker, depending on mode.
+$EOM_DEFAULTS = Array("stream" => chr(30), "file" => chr(30));
+
+# Variable initialization.
+# Action. Possible values are 'index' and 'update'. Note: this variable is not
+# a constant and can be changed with command line argument. The "DEFAULT" in
+# its name refers to action not being specified in the data being processed,
+# not in the script's configuration (like the default values above).
 $DEFAULT_ACTION = 'index';
```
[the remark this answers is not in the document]
Evildoor: > The `$DEFAULT_ACTION` description is not very clear to me.
> 
> First of all, I'd recommend to start every section of the description with a new line:
> 
> ```
> # Action
> # Possible values: ...
> # Note: ...
> # ...
> ```

Looks reasonable to me, will do.

> Then, what is unclear to me:
> 
> > this variable is not a constant...
> 
> Pretty obvious: variables usually are _not_ constants. What does it mean?
> 
> > ...and can be changed with command line argument.
> 
> Er... I mean right, it can be changed, but is it important at this point?
> 
> > action not being specified in the data being processed, not in the script's configuration (like the default values above).
> 
> Too many "not"s: while reading, I tend to be treat them as more or less similar "not"s, but they are not... and even though I know it for now, I totally fail to read this phrase properly.
> 
> But before we start looking for better wording, I have a question: is it necessary to add this note here? I mean, whatever global variable it is, here all that you want is to understand its meaning; how it is used and where its value is altered is the second question, which is more likely to be investigated by looking into the code itself, not via reading the explaination of all occurences of the variable name gathered in one place... since to understand the latter you still need to know something about the code.

All these concerns make sense, and can be resolved by changing the name of the variable. The name - or, more precisely, the presence of "DEFAULT" in it - is what made me write this lengthy comment.  There are 6 variables initialized here, ones from the first group are used as constants (their values are set here, and not changed later), and the other ones are used normally. Only "constants" have the word "DEFAULT" in their names, hence I decided to point out that `$DEFAULT_ACTION` is different.

Actually, it seems that I was wrong in the groups' construction, because `$UPDATE_RETRIES` is also a "constant". Therefore, such division of "DEFAULT/no DEFAULT" makes even less sense. Maybe the comment can be shortened without renaming the variable...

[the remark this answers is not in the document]
mgolosova: Things may be unified even better, if we introduce another variable `$ACTION`:
```
$DEFAULT_ACTION = 'index'

<...>

function getAction($row) {
  global $ACTION;
  if (isset($row['_update']) and $row['_update'] === true) {
    $action = 'update';
  } else {
    $action = $ACTION;
  }
  return $action;
}

<...>
    case "update":
      $ACTION = "update";
      break;
<...>

if (!isset($ACTION)) $ACTION = $DEFAULT_ACTION;
```

...and remove `$ES_INDEX` initialization with `null` from top of the file -- anyway it is set to the env variable value or `$DEFAULT_INDEX` later, when all other variables like `$EOM_MARKER` get their values.

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -102,7 +189,8 @@ function decode_escaped($string) {
                       'stripcslashes("$0")', $string);
 }
 
-$opts = getopt("e:E:", Array("end-of-message:", "end-of-process:",
+# Process command line arguments.
+$opts = getopt("e:E:h", Array("help", "end-of-message:", "end-of-process:",
```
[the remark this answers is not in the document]
Evildoor: True, done.

Utils/Dataflow/019_esFormat/run.sh
```diff
@@ -1,23 +1,27 @@
...
-  echo "USAGE:
-$(basename "$0") [-c CONFIG] [--] [ARGS]
+  echo "usage: $(basename "$0") [-c CONFIG] [--] [ARGS]
   
-PARAMETERS:
-  CONFIG -- configuration file
-  ARGS   -- arguments to be passed to the PHP script
+optional arguments:
+  -c CONFIG    Elasticsearch configuration file
+  ARGS         arguments to be passed to the PHP script (see below)
```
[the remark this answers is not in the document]
Evildoor: True. Added mention, will deal with excessive "help" later.

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -53,6 +104,27 @@ function getAction($row) {
 }
 
 function constructActionJson($row) {
+  /* Generate a json with ES bulk API action information for a given document.
+
+  Action json is generated for each document to be processed in a single
+  bulk action, and includes directions about how exactly the document must
```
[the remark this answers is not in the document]
Evildoor: True, simplified the description.

Utils/Dataflow/019_esFormat/run.sh
```diff
@@ -1,23 +1,27 @@
...
 "
+$base_dir/esFormat.php -h
```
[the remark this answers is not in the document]
Evildoor: Thanks for the suggestion, done.

Utils/Dataflow/019_esFormat/run.sh
```diff
@@ -7,7 +7,7 @@ usage() {
 $(basename "$0") [-c CONFIG] [--] [ARGS]
   
 PARAMETERS:
-  CONFIG -- configuration file
+  CONFIG -- Elasticsearch configuration file
```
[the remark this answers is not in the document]
Evildoor: Just a 'config file' looks too unclear to me. Reworded and added the default value.

Utils/Dataflow/019_esFormat/run.sh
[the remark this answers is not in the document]
Evildoor: Thanks for the suggestion, done.

Evildoor:
> would be nice to have one for getAction() as well

Done.

Utils/Dataflow/019_esFormat/esFormat.php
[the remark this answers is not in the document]
Evildoor: Done.

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -5,14 +5,60 @@ function exception_error_handler($errno, $errstr, $errfile, $errline ) {
...
+# Variable initialization.
+# Action. Possible values are 'index' and 'update'. Note: this variable is not
+# a constant and can be changed with command line argument. The "DEFAULT" in
+# its name refers to action not being specified in the data being processed,
+# not in the script's configuration (like the default values above).
 $DEFAULT_ACTION = 'index';
+# How many times the update should be retried in case of conflict.
 $UPDATE_RETRIES = 3;
+# ES index where the documents should be indexed/updated.
 $ES_INDEX = NULL;
```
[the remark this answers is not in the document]
Evildoor: Removed this initialization per discussion above.

Utils/Dataflow/019_esFormat/run.sh
```diff
@@ -1,23 +1,32 @@
...
+  -h, --help   show this help message and exit
 
-ES_CONFIG=$base_dir/../../Elasticsearch/config/es
+  -c CONFIG    configuration file with parameters for
+               accessing data in Elasticsearch
+               Default value: $ES_CONFIG
+
+ARGS, arguments to be passed to the PHP script:"
+# Display part of esFormat.php's help describing its arguments.
+$base_dir/esFormat.php -h 2>&1 | sed 1,5d >&2
```
mgolosova: This version relies on current formet of the `esFormat`'s help message, and should it change -- this part must be changed as well. And I bet the one changing help message won't remember to do it ;)

Current version works, so it can be merged as-is, but it's better not to leave such traps behind. If `sed` can't skip plus one line after `/--help/` (I mean, in simple way) -- maybe `awk` will do better.

---

*If it's difficult to fix right now, it's fine to leave as-is (but add ToDo note, Trello card, or something else as a reminder).*

Utils/Dataflow/019_esFormat/run.sh
```diff
@@ -1,23 +1,32 @@
...
-  ARGS   -- arguments to be passed to the PHP script
-"
-}
+optional arguments:
+  -h, --help   show this help message and exit
 
-ES_CONFIG=$base_dir/../../Elasticsearch/config/es
+  -c CONFIG    configuration file with parameters for
+               accessing data in Elasticsearch
+               Default value: $ES_CONFIG
```
mgolosova: "Accessing data in ES" sounds a bit confusing, as if the stage was accessing ES -- while it does not.
Can you try to reword it a bit?
For me, the most accurate description would be "config.file with parameters required to prepare data for indexing in ES" (and the most "as-is" -- "config.file containing ES_INDEX definition").

Utils/Dataflow/019_esFormat/run.sh
```diff
@@ -3,14 +3,14 @@
 base_dir=$(cd "$(dirname "$(readlink -f "$0")")"; pwd)
 
 usage() {
-  echo "USAGE:
-$(basename "$0") [-h] [-c CONFIG] [--] [ARGS]
+  echo "usage: $(basename "$0") [-h] [-c CONFIG] [--] [ARGS]
```
mgolosova: `--` part can not be skipped if `ARGS` is specified, so more accurate way would be something like:
```
   usage: ... [-- [ARGS]]
```

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -5,14 +5,54 @@ function exception_error_handler($errno, $errstr, $errfile, $errline ) {
...
-$UPDATE_RETRIES = 3;
-$ES_INDEX = NULL;
+# End-of-process marker, depending on mode.
 $EOP_DEFAULTS = Array("stream" => chr(0), "file" => "");
+# End-of-message marker, depending on mode.
 $EOM_DEFAULTS = Array("stream" => chr(30), "file" => chr(30));
+# How many times the update should be retried in case of conflict.
+$UPDATE_RETRIES = 3;
+# Action.
+# Possible values: 'index', 'update'.
```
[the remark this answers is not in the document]
Evildoor: Done, 29ada3a7.

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -169,6 +270,9 @@ function decode_escaped($string) {
   $ES_INDEX = $DEFAULT_INDEX;
 }
 
+if (!(isset($ACTION))) $ACTION = $DEFAULT_ACTION;
```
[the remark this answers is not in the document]
Evildoor: Done, da946de.

Utils/Dataflow/019_esFormat/run.sh
Evildoor: Added note, 5cf279d.

Utils/Dataflow/019_esFormat/run.sh
Evildoor: Done, edc034c.

Utils/Dataflow/019_esFormat/esFormat.php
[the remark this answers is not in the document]
Evildoor: Done, da946de.

Utils/Dataflow/019_esFormat/run.sh
Evildoor: I had my doubts about this "argument", but always put them off because something else had to be done. Since we got to this...

First of all, the argument `--` is not explained anywhere (I hope that we'll fix this here, after this discussion). If I understand correctly, it is meant to separate arguments intended for `run.sh` itself from those that it must pass to `esFormat.php`, aka ARGS. However, it seems to be unnecessary, since the [`case`](https://github.com/PanDAWMS/dkb/blob/019-update-action/Utils/Dataflow/019_esFormat/run.sh#L18) processing the arguments has four cases: two for `run.sh`'s arguments, and two others: `--` and `*`, where the latter catches everything that is not the first three. Only difference between `--` and `*` is `shift` for the first, so that `--` itself is not passed to `esFormat.php`.

Therefore, what's the point of the separator if the logic "we only expect `-h` and `-c`, if we encounter something else then it's already ARGS, stop" is in place? One thing I can think of is that this is made in advance, for the situation in the future where `run.sh` would accept more arguments and their amount and type would be arbitrary, hence the need for separator. Another option is possible need to pass something to `esFormat.php` that is similar to `run.sh`'s arguments. Then yes, separator will bring sense into something like `./run.sh -c elastic-search-config -c php-handling-specific-config`.

To summarize, I see 3 options:
1) Implement your suggestion, but also update the logic to demand `--` and document it in `usage()`.
2) Discard your suggestion and simply document `--`, emphasizing the fact that it should only be used to separate aforementioned `-c` and `-c` (in other cases `*` will do just fine).
3) Drop `--` altogether and rely on `*` until (if) the need for the separator arises.

Am I missing something?

mgolosova:
[Stage 019: мелкие исправления](https://trello.com/c/ZUvvKjKC/93-stage-019-%D0%BC%D0%B5%D0%BB%D0%BA%D0%B8%D0%B5-%D0%B8%D1%81%D0%BF%D1%80%D0%B0%D0%B2%D0%BB%D0%B5%D0%BD%D0%B8%D1%8F)

mgolosova:
[Stage 019: мелкие исправления](https://trello.com/c/ZUvvKjKC/93-stage-019-%D0%BC%D0%B5%D0%BB%D0%BA%D0%B8%D0%B5-%D0%B8%D1%81%D0%BF%D1%80%D0%B0%D0%B2%D0%BB%D0%B5%D0%BD%D0%B8%D1%8F)

Utils/Dataflow/019_esFormat/run.sh
mgolosova: @Evildoor,
> the argument `--` is not explained anywhere 

Double dash (`--`) is a standard separator for command line arguments (see discussion [here](https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean), for example). Not sure if we need to add some explicit explanation.

>  right now it seems to be unnecessary, since the case processing the arguments has four cases: two for run.sh's arguments, and two others: `--` and `*`, where the latter catches everything that is not the first three. Only difference between `--` and `*` is shift for the first, so that `--` itself is not passed to `esFormat.php`.
>  what's the point of the separator if the logic "we only expect -h and -c, if we encounter something else then it's already ARGS, stop" is in place?

Right. Almost. And no one said that current logic is impeccable, right? ;)
Here's the answer:
1. For now, without `--` you already have no possibility to pass `-h` to `esFormat.php` directly (without running the PHP script itself). No one said you would ever want to, but...
2. Instead of "if it's something unknown (`*`), just go on" there probably should be "if you don't know what it is (`*`), ignore it (do not pass it to `esFormat.php`)".
3. It may also be reasonable to add check for `-*` after `--` to signalize about "unknown option" and exit.

So...
> To summarize, I see 3 options:
> ...
> Am I missing something?

...yes, you're kind of missing the general idea: it's not about "implement/ignore someone's suggestion", it's  more about what do we want to achieve. I'd love to believe the aims here are:
* to finish this PR;
* to make things right.

For the sake of the first aim -- fixing right here every bug detected is not really best way to achieve it (unless the PR was initially about fixing everything).

Talking about the second: `--` separator is required to make sure that user knows what he/she does. In other words, it's here to tell user: "you want to pass some arguments to `esFormat.php` -- say it explicitly". The fact that without `--` it works just the same is nothing but a neglect (mine, actually) that should be questioned and fixed eventually. Now or some time later -- is up to you. I mean, what's gonna take more time: fix it right now or describe the situation in Trello for someone else (or you yourself) who'll have required time later?

---

There's also another thing: why the `run.sh` is used in the first place. Just a thought, but I'd like you to keep it in your mind before deciding what to do next.
The `run.sh` was added, actually, to pass configuration parameters from `es` config file (shaped as Shell variables declaration) to the `PHP` script since it looked easier than do this directly from the PHP (just source the file, no need to parse it and check syntax and all these things required). Should this decision lead to way too many questions now, it may be reasonable to reconsider this design and teach PHP script to read the config directly, without this Shell wrapper. Or not: evaluate the complexity and make your decision :) It's easier to go on the way someone else has chosen, yet there are always reasons to make choice back then and it's better to be aware of the fact that with time and under different circumstances these reasons may look not so persuasive.

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -41,18 +93,50 @@ function convertIndexToLowerCase(&$a) {
...
+  :return: action
+  :rtype: str
+  */
+  global $ACTION;
 
   if (isset($row['_update']) and $row['_update'] === true) {
     $action = 'update';
   } else {
-    $action = $DEFAULT_ACTION;
+    $action = $ACTION;
```
mgolosova: Somehow this part makes me a bit nervous.

`$ACTION` *is* defined somewhere in the "main" program code, but it is not obvious that this function will only be used after it happens.

What do you think?

Maybe add some kind of "safety wire" here: if `$ACTION` is not set, fall back to the `$DEFAULT_ACTION`?
I know it is exactly what happens [here](https://github.com/PanDAWMS/dkb/pull/292/files#diff-c5f3f98d4013266cbee7633d5d2f7ec8R252):
```
if (!(isset($ACTION))) $ACTION = $DEFAULT_ACTION;
```
..., and that situation with `$ACTION` is just the same as with `$ES_INDEX`, used in `constructActionJson($row)`; but still...

Utils/Dataflow/019_esFormat/run.sh
Evildoor: > > the argument `--` is not explained anywhere
> 
> Double dash (`--`) is a standard separator for command line arguments (see discussion [here](https://unix.stackexchange.com/questions/11376/what-does-double-dash-mean), for example). Not sure if we need to add some explicit explanation.

I see. Still, I think that adding an explanation won't hurt - someone (such as myself at the beginning of this discussion) may be unaware of this standard, and I doubt that 1-2 strings that state the obvious would be much of a problem for the others. Will do.

> ... I'd love to believe the aims here are:
> 
> * to finish this PR;
> * to make things right.
> 
> For the sake of the first aim -- fixing right here every bug detected is not really best way to achieve it (unless the PR was initially about fixing everything).
> 
> Talking about the second: `--` separator is required to make sure that user knows what he/she does. In other words, it's here to tell user: "you want to pass some arguments to `esFormat.php` -- say it explicitly". The fact that without `--` it works just the same is nothing but a neglect (mine, actually) that should be questioned and fixed eventually. Now or some time later -- is up to you. I mean, what's gonna take more time: fix it right now or describe the situation in Trello for someone else (or you yourself) who'll have required time later?

Will look into it.

Utils/Dataflow/019_esFormat/esFormat.php
Evildoor: As you correctly point out, this situation - "we use global variable in a function without checking whether it was correctly initialized" - is present in other places of this script. Furthermore, it also exists in [at least one other place](https://github.com/PanDAWMS/dkb/blob/019-minor-fixes/Utils/Dataflow/071_esConsistency/consistency.py#L107) in this repository (by the way, there seems to be an error in documentation in that place, will look into it). In light of such situation, I'm not sure that adding a safety measure in this single place is a way to go.
Will make a Trello card about the situation in general, or maybe address the situation's instances in this stage in this PR, since it fits into "fix various minor problems" of the description.

Utils/Dataflow/019_esFormat/run.sh
Evildoor: Added explanation for `--`.

Added card for issues with the separator: https://trello.com/c/6vEydzfY. While it is possible that, after some consideration and discussion, the problem would be solved by some small fix, it is also possible that the stage would be redesigned, by removing `run.sh` and moving everything into PHP script, for example. Such (possible) changes do not fit into the idea of "minor fixes" in this PR and, should be left out of it.

Evildoor:
Rebased the branch on master to remove the commit that was already merged separately (#322).

Utils/Dataflow/019_esFormat/esFormat.php
Evildoor: > Will make a Trello card about the situation in general

https://trello.com/c/ZUZvlz3n

> and maybe address the situation's instances in this stage in this PR

Decided not to, since this looks like a general issue that warrants some thinking.

Utils/Dataflow/019_esFormat/run.sh
mgolosova: Ok, thank you!

Utils/Dataflow/019_esFormat/esFormat.php
mgolosova: OK then, I'll mark this one as resolved.

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -41,18 +93,50 @@ function convertIndexToLowerCase(&$a) {
...
+  "update":true in the document. The value is an array that contains the
+  remaining information:
+  - index name
+  - number of retries if the action is update
+  - document id
+  - document type
+  - document's parent, if it is specified.
+
+  The latter three values are taken from the document.
+
```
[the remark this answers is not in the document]
Evildoor: Done, 334ab5c.

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -78,6 +162,21 @@ function constructActionJson($row) {
 }
 
 function constructDataJson($row) {
+  /* Prepare the document for bulk operation.
+
+  - Remove fields starting with an underscore. These are service fields
+    that are processed separately and should not be included into the
+    resulting data.
+  - Set "doc_as_upsert"=true in case of update. This means that the same
```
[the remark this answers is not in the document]
Evildoor: Updated, 3b222301.

Utils/Dataflow/019_esFormat/run.sh
```diff
@@ -1,23 +1,36 @@
...
+$base_dir/esFormat.php -h 2>&1 | sed 1,5d >&2
+}
 
 while [ -n "$1" ]; do
   case "$1" in
+    --help|-h)
+      usage >&2; exit 1;;
     --config|-c)
-      [ -n "$2" ] && ES_CONFIG="$2" || { usage >&2 && exit 1; }
+      [ -n "$2" ] && { ES_CONFIG="$2" && CONFIG_DEFAULT=""; } \
```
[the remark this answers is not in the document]
Evildoor: True. Done, 8a47f98.

Utils/Dataflow/019_esFormat/run.sh
Evildoor: Done, a1eb61b.

Utils/Dataflow/019_esFormat/esFormat.php
```diff
@@ -41,18 +93,53 @@ function convertIndexToLowerCase(&$a) {
 }
 
 function getAction($row) {
-  global $DEFAULT_ACTION;
+  /* Determine action for the document.
+
+  Action is 'update' if {'_update': true} key-value pair is 
+  present in the document. Otherwise, default value is returned.
```
mgolosova, line 99: I'm afraid that after 67a80d8 (#320) this description is incomplete.
I can merge this PR to immediately fix it in the next one, or you can merge `master` here and update the description accordingly.

Evildoor: Done, 3a078d4.

Evildoor:
Comment about "error status" is valid, I probably made a mistake while typing. Fixed.

mgolosova:
@Evildoor, thank you!
Approved and merged now. If you do not need this branch for the further work, please remove it.

PR #299, opened 2019-11-12
API: setup fix

mgolosova:
Fix issues preventing regular user from installation of own development instanse of API server.

ToDo:
- [ ] final checks **(@Evildoor: bamboo, @mgolosova: aiatlas172)**;
- [x] move commits with additional config parameters to a separate PR (do not belong here).

### DIFF
--- a/Utils/API/server/setup.sh
+++ b/Utils/API/server/setup.sh
@@ -419,7 +419,7 @@ uninstall_www() {
     # Remove directory if it still exists...
     if [ -d "$d" ]; then
       # ...and does not contain files.
-      nfiles=`find $d -type f ! -name "*.pyc" | wc -l`
+      nfiles=`find $d -type f ! -name "*.pyc" ! -name "*.fcgic" | wc -l`
       [ $nfiles -eq 0 ] \
         && echo "> $d" >&2 \
         && rm -rf "$d" >/dev/null \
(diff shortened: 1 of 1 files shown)


Utils/API/server/setup.sh
```diff
@@ -19,6 +19,11 @@ NGINX_GROUP=
 MANAGE_SERVICE=
 MANAGE_SEL=
 
+# Change with caution, for this directory must have
+# specific permissions that allow nginx user
+# access to the socket files
+SOCK_DIR="$RUN_DIR"
```
Evildoor: RUN_DIR is default here. If RUN_DIR is changed via command line argument later, SOCK_DIR remains the same. Is this OK?

mgolosova: Yes, that's the exact point: `SOCK_DIR` must be readable and writable for Nginx user, so changing it makes sence only if you're sure that you have privileges to allow it to the desired directory. Of course, no one will stop you from making some directory read/writable for everyone, but it won't be very good idea -- since it means that anyone can (even erroneously) remove your app socket file, for example. Or write to the socket something, or read, whatever.

But you have a point: if we want specific value for this variable, it is a bit wierd to define it with value of another one. I think I better set `SOCK_DIR` to the desired value and `RUN_DIR` to `$SOCK_DIR` -- just to make sure that it won't be changed accidentally.

mgolosova:
[Rebased](https://github.com/PanDAWMS/dkb/compare/dd11e12214670fdbcf2ffc771c2e3a925214d11b..345e9abbaf3b6938200f472651b7e4475a2491a0) the branch on the `api` with force-push for more consistent tests on CERN node.

PR #304, opened 2019-12-02
DF/095: typo fix (missed whitespace).

mgolosova:


Evildoor:
In numerous similar cases in the script the space is added at the end of the line rather than the start of the next one. This change does otherwise - while it has a reason, namely the line's length of 80 symbols, maybe the word "create" should be moved to the next line to overcome this and uphold the style?

mgolosova:
@Evildoor 
> In numerous similar cases in the script the space is added at the end of the line rather than the start of the next one. This change does otherwise - while it has a reason, namely the line's length of 80 symbols, maybe the word "create" should be moved to the next line to overcome this and uphold the style?

I have reasons to think it's better (and at least not worse) to leave things as they are.

The beginning-spaced version is used because:
* here: yes, I didn't want to hit the 80-chars limit (maybe I wouldn't anyway, but it required an extra action to make sure ;) );
* in general: when you have a multi-line text, it is easier not to forget this space between lines if it is always in the beginning:
  ```
  some_var = "My very long text "
             "that takes multiple lines. "
             "Have you noticed the line"
             "where the space-between-lines "
             "is missed?"

  some_var = "My very long text"
             " that takes multiple lines."
             " Have you noticed the line"
             "where the space-between-lines"
             " is missed?"
  ```

Talking about statistics:
```
[root@aiatlas171 Dataflow]# git status -u no
# On branch 095-minor-fix
nothing to commit, working directory clean
[root@aiatlas171 Dataflow]# echo "End-spaced:" $(git grep ' ["'"'"']$' | grep '\.py' | wc -l)
End-spaced: 55
[root@aiatlas171 Dataflow]# echo "Beginnig-spaced:" $(git grep '^ *["'"'"'] [a-zA-Z]' | grep '\.py' | wc -l)
Beginning-spaced: 54
```

So, in the whole repository the picture is quite balanced: if we substract that one line in question, it will be **53** beginning-spaced vs. **55** end-spaced, so it is not about the global codestyle.

In this exact directory:
```
[root@aiatlas171 Dataflow]#
[root@aiatlas171 095_datasetInfoAMI]# git grep '^ *["'"'"'] [a-zA-Z]' | grep '\.py'
amiDatasets.py:            " Are you sure you have a valid certificate?\n")
amiDatasets.py:                         " proxy.\n")
[root@aiatlas171 095_datasetInfoAMI]# git grep ' ["'"'"']$' | grep '\.py'
amiDatasets.py:        sys.stderr.write("(FATAL) Failed to initialise AMI client: "
amiDatasets.py:        sys.stderr.write("(ERROR) Login authentication mode is not "
amiDatasets.py:        raise DataflowException("Failed to initialise AMI client: certificate "
```

You're right, it is **3** vs. **1** in favor of end-spaced (before this change). But still it's not *only* end-spaced, and statistics is quite shallow -- so should it really be turned into a file-specific codestyle recommendation? I mean, if there were **10** vs. **1** -- that would be a stronger point and I might even want to change that already existing one line as well. But in our case (and taking into account what is said above about why I think that beginning-spaced lines are better in general) -- the question looks disputable to me.

PR #307, opened 2019-12-05
REST-API derivation statistics.

Evildoor:
Implement an API method to get derivation statistics.
https://trello.com/c/TknSBKGY

To do:
- [x] Check format.
- [x] Write documentation.


Utils/API/server/lib/dkb/api/storages/es.py
```diff
@@ -487,3 +490,51 @@ def task_kwsearch(**kwargs):
...
+        total = 0
+        ratio = 0
+        events_ratio = 0
+    return {'total': total, 'ratio': ratio, 'events_ratio': events_ratio}
+
+
+def task_derivation_statistics(**kwargs):
+    init()
+    project = kwargs.get('project').lower()
+    tags = [tag for tag in kwargs.get('amitag').split(',') if tag]
```
mgolosova: Would be better to have `amitag`s passed as array initially: `...?amitag=TAG1&amitag=TAG2&...`; in this case `kwargs['amitag']` will have value `['TAG1', 'TAG2']`. It is not difficult to parse comma-separated string, of course, and AMI tags are not supposed to have comma within themselves -- but strict version is just more strict and leaves less space for mis-interpreting and mis-use. Also, it is the way keywords search is implemented (`...?kw=KEYWORD1&kw=KEYWORD2&...`), and I think we should keep same style of parameters handling in all methods.

Evildoor: Good idea, done.

Evildoor:
@mgolosova - please, take a look. One point of interest is the format of the results, that is illustrated in Utils/API/server/lib/dkb/api/storages/__init__.py: it can, possibly, change, depending on the components that will use this API.

Utils/API/server/lib/dkb/api/handlers.py
```diff
@@ -253,3 +253,29 @@ def task_kwsearch(path, **kwargs):
...
+    :return: calculated statistics
+    :rtype: dict
+    """
+    method_name = '/task/deriv'
+    project = kwargs.get('project')
+    if project is None:
+        raise MissedArgument(method_name, 'project')
+    tags = kwargs.get('amitag')
+    if tags is None:
+        raise MissedArgument(method_name, 'amitag')
```
mgolosova: Just a comment: it is OK to just check if given key is in hash `kwargs`, like `if 'project' not in kwargs: ...`. Taking exact value is useful when you need to check the value itself; if not -- then it is an excessive action.

Utils/API/server/lib/dkb/api/storages/__init__.py
```diff
@@ -77,3 +77,30 @@ def task_kwsearch(**kwargs):
...
+
+    Resulting statistics has the following structure:
+    {
+      'data': {
+        ...
+        'some_output_format_name': {
+          'total': 123,
+          'ratio': 0.456,
+          'events_ratio': 0.789
+        },
```
mgolosova: 'total' is a number of tasks, right?

Evildoor: Yes. It is ES's term which is used in Big PanDA Monitoring and here - in the former it is replaced with "tasks" before returning the response. I've made no such replacement here because, as I've said in a comment, such matters depend on where and how this data will be handled next.

mgolosova: Used in ProdTask interface, you meant to say (it is built upon the core of BigPanDA Monitor, but it is a different instance and different team (M.Borodin exclusively, if I get it correctly)).

Anyway: ES says "total", meaning "total number of matching documents" (since it doesn't know if these documents are "tasks" or what). We, on the other hand, are fully aware of the meaning of the data we are providing -- so why not make things a bit more user-friendly?

Utils/API/server/lib/dkb/api/storages/es.py
```diff
@@ -487,3 +490,91 @@ def task_kwsearch(**kwargs):
...
+
+    :return: output formats
+    :rtype: list
+    """
+    formats = []
+    for tag in tags:
+        f = OUTPUT_FORMATS.get(tag)
+        if f:
+            formats += f
+    return formats
```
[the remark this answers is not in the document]
Evildoor: Done.

Utils/API/server/lib/dkb/api/handlers.py
Evildoor: True. Done.

Utils/API/server/lib/dkb/api/storages/es/query/deriv
```diff
@@ -0,0 +1,54 @@
...
+        {"terms": {"ctag": %(ctag)s}},
+        {"term": {"status": "done"}},
+        {"range": {"input_bytes": {"gt": 0}}},
+        {"has_child": {
+            "type": "output_dataset",
+            "query" : {
+              "bool": {
+                "must": [
+                  {"term": {"data_format": %(output)s}},
+                  {"range": {"bytes": {"gt": 0}}}
```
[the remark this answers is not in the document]
Evildoor: Seems so. For example, task with id `14120127` has a single deleted dataset `data18_13TeV.00349646.physics_Main.deriv.DAOD_BPHY7.f933_m1960_p3553_tid14120127_00` - it is not found by the query, but IS found by one with this restriction removed.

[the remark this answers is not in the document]
mgolosova: Sorry, what I meant to ask was: is it the expected behaviour?
Those input datasets, filtered by this condition, are excluded from the resulting `"input_{events|bytes}"` -- did you think about it and decided that it its correct (then I'm asking about motivation for this decision), or was this effect unintended (then I'm asking you to take a closer look at it and make a deliberate decision)?

Utils/API/server/lib/dkb/api/storages/__init__.py
Evildoor: > @Evildoor,
> The idea is that REST API method returns data in the form ProdTask expects. For now, ProdTask constructs `result` as `List` of `Hash`es (check hash format [here](https://github.com/PanDAWMS/panda-bigmon-atlas/blob/052aebea3dcb3f1a5f4e9e7249c2ffc87f2bcb92/atlas/dkb/views.py#L931-L932)), sorted by `"output"` value.
> 
> Why have you decided to return different format?

Because I mistakenly thought that the resulting format is yet to be decided, while it is decided by ProdTask expectations. My bad, changed the format.

> Also, note the `"task_ids"` key, value of which is empty in ProdTask version for now. Could you provide value for it as well?

Done.

Utils/API/server/lib/dkb/api/storages/es/query/deriv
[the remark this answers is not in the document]
Evildoor: Hm. This condition was, like the rest of the query, taken from ProdTask's code (so the answer to your question is "no, I was not specifically evaluating this condition"). Should we change it?

While the input datasets are, as you correctly say, excluded from resulting `input` values, the output datasets of the corresponding tasks are similarly excluded from the `result` ones. Therefore, there seems to be no error in this regard.

Another point can be "why should both input and result values of these tasks be excluded from the statistics?", but, if I understand correctly, the answer is "because without input/output/both datasets no values - or only one of each pair - can be obtained, making the resulting calculation (for this particular task) impossible".

Utils/API/server/lib/dkb/api/storages/__init__.py
Evildoor: Update: while 8cf6a28 allows `task_ids` to be retrieved, it only gets them for 10 tasks or less, because it's the default behavior for searches in ES, to get one "page" of results. Currently, each format has less than 50 tasks, so this can be "fixed" by changing the size of the "page" from 10 to 100-500-etc.

However, in arbitrary case we don't know the amount of tasks, and to properly get all the ids of all the tasks (in request) it is necessary, if I understand correctly, to use [scrolling](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-scroll), which "... is not intended for real time user requests".

What should we do?

Utils/API/server/lib/dkb/api/storages/es.py
```diff
@@ -28,6 +28,9 @@
 # Default datetime format
 DATE_FORMAT = '%d-%m-%Y %H:%M:%S'
 
+with open('ctag_formats.json') as f:
+    OUTPUT_FORMATS = json.load(f)
```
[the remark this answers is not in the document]
Evildoor: Done.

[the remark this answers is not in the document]
Evildoor: Note: usage of additional query to ES instead of json file representing Oracle leads to a single difference in results for values used for testing (project `data18_13TeV`, list of tags consisting of one tag `p3553`): ES contains tasks with `p3553` tag and output datasets with `DAOD` format, while Oracle's data yields no such output corresponding to this tag. I doubt that this is a problem, since Oracle produces this format when presented with some other tags (`p3166`, for example), hence it does not looks to me as some special format that should be avoided.

Utils/API/server/lib/dkb/api/storages/es/query/deriv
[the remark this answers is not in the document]
mgolosova: > This condition was, like the rest of the query, taken from ProdTask's code

My point here is that even if you have someone else's code to get the general idea from, it is usually a good idea to dive into it and understand what the copied pieces of code do and why/for what reason. Right, Mikhail knows better what he needs to see in the result -- but everyone can make a mistake, so if you have no idea why something is done this way and not another, it may as well mean that it *is* a mistake and here's a chance to fix it. Or not, but in order to find out you have to ask the question.

> (so the answer to your question is "no, I was not specifically evaluating this condition"). Should we change it?

And that's the thing that should definitely be changed -- the belief that others can do no mistake ;)

---

Now, to the question itself...

> While the input datasets are, as you correctly say, excluded from resulting input values, the output datasets of the corresponding tasks are similarly excluded from the result ones. Therefore, there seems to be no error in this regard.

Sounds reasonable for me: if you have excluded some task from the calculation, it equally affects `sum(input)` and `sum(oputput)`, so that the result ratio should be correct.

> "why should both input and result values of these tasks be excluded from the statistics?" -- "because without input/output/both datasets no values - or only one of each pair - can be obtained, making the resulting calculation (for this particular task) impossible"

But we do not perform "calculation for particular task" -- it's two total sums and then the ratio of these values. Yet if there was only one task matching the `project-tag-format` combination -- that'd be a problem indeed...

I'd say that the original idea might have been as follows: if we know values only for output and add them to the total sum, the `total(output)` would contain more tasks than `total(input)` and dividing one by another would simply be incorrect (and vice versa). Does it sound logical, or is a bit of a stretch?

Utils/API/server/lib/dkb/api/storages/__init__.py
mgolosova: For now we can provide accurate value for the number of tasks (which is already there in `"tasks"` field) and, say, first 100 IDs. This will allow Mikhail to test usability of this data, output some notification if there's some "hidden" data -- and see if users ever need to see more than these 100 tasks.

I believe it shouldn't significantly affect the statictics calculation time, but should you see the opposit -- we can reduce the value to 50 or even get back to default 10.

Utils/API/server/lib/dkb/api/storages/es.py
```diff
@@ -502,10 +499,15 @@ def get_output_formats(tags):
...
-            formats += f
+    query = dict(TASK_KWARGS)
+    kwargs = {'tags': tags}
+    query['body'] = get_query('output_formats', **kwargs)
+    query['_source'] = ['data_format']
+    query['doc_type'] = 'output_dataset'
+    query['size'] = 500
+    r = client().search(**query)
+    for hit in r['hits']['hits']:
+        formats += hit['_source']['data_format']
```
mgolosova: This function takes from ES first 500 output datasets that match the [query](https://github.com/PanDAWMS/dkb/blob/c5326669ce489478a35d8c096152446ba37281a2/Utils/API/server/lib/dkb/api/storages/es/query/output_formats) condition ("dataset has parent with given `ctag`"), and then aggregates received data to get the list. Right?

My questions:
1. Wouldn't it be better to allow ES to make the aggregation?
Why:
    * for now there's a possibility to get incorrect result *(**Q**: when?)*
    * ES can do it faster *(**Q**: why?)*

2. Why only `ctag`?
(In this case) we seem to be interested only in formats for given `ctag-project` combination, and if this list is shorter than that of all formats for given `ctag` -- we're sending extra queries to ES that return no data, yet take some time to send request and get the response. It's not necessary that ES is locally available, so there's gonna be some additional cost for the network operations.

Utils/API/server/lib/dkb/api/storages/es.py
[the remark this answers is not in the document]
mgolosova: It must be due to the way `"data_format"` field is filled: it contains list of both "real" format ("DAOD_SMTH") and "main" format ("DAOD") (see [here](https://github.com/PanDAWMS/dkb/blob/c5326669ce489478a35d8c096152446ba37281a2/Utils/Dataflow/093_datasetsFormat/datasets_format.py#L48) for details).

Not sure how to handle it for now, will take a closer look ASAP.

Utils/API/server/lib/dkb/api/storages/__init__.py
Evildoor: > > > The idea is that REST API method returns data in the form ProdTask expects. For now, ProdTask constructs result as List of Hashes (check hash format [here](https://github.com/PanDAWMS/panda-bigmon-atlas/blob/052aebea3dcb3f1a5f4e9e7249c2ffc87f2bcb92/atlas/dkb/views.py#L931-L932)), sorted by "output" value.
> > 
> > 
> > changed the format.
> 
> Have you noticed the "sorted by "output" value" part (which was added due to this [line](https://github.com/PanDAWMS/panda-bigmon-atlas/blob/052aebea3dcb3f1a5f4e9e7249c2ffc87f2bcb92/atlas/dkb/views.py#L973))? I couldn't find how it was implemented...

Missed it, my bad. Updated 735bf7d.

> 
> > Update: while [8cf6a28](https://github.com/PanDAWMS/dkb/commit/8cf6a28268eaa18181d4445bec14a47beff9fa8c) allows task_ids to be retrieved, it only gets them for 10 tasks or less, because it's the default behavior for searches in ES, to get one "page" of results. Currently, each format has less than 50 tasks, so this can be "fixed" by changing the size of the "page" from 10 to 100-500-etc.
> > However, in arbitrary case we don't know the amount of tasks, and to properly get all the ids of all the tasks (in request) it is necessary, if I understand correctly, to use scrolling, which "... is not intended for real time user requests".
> > What should we do?
> 
> For now we can provide accurate value for the number of tasks (which is already there in `"tasks"` field) and, say, first 100 IDs. This will allow Mikhail to test usability of this data, output some notification if there's some "hidden" data -- and see if users ever need to see more than these 100 tasks.
> 
> I believe it shouldn't significantly affect the statictics calculation time, but should you see the opposit -- we can reduce the value to 50 or even get back to default 10.

Good. Updated 578e3ce.

Utils/API/server/lib/dkb/api/storages/es/query/deriv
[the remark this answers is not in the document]
Evildoor: Aha, it seems that I've missed your point. While there are "incorrect" tasks with 0 input and/or output bytes and these tasks are discarded because their contribution cannot be calculated correctly, there is (by logic) a possibility of a "correct" task with positive input but zero output [1]. To give an abstract example, if input is the list `[1, 2, 3]`, then "incorrect" task A with algorithm "find all even values" (or any other one) can have output `[]` instead of expected `[2]` due to some error/output being lost/whatever, but a "correct" task B with algorithm "find all negative values" can produce the same output `[]`, and there is nothing wrong with it. In such case, input events/bytes of B should be added to input total, which currently is not done. Is this what you've been trying to tell me?

If the answer is "yes", then the next question is "how can we distinguish between tasks A and B?". I don't have an answer, so we should probably discuss this with Mikhail, or someone else who is better informed about the ATLAS' data.

Some comments on other points of interest:

> > This condition was, like the rest of the query, taken from ProdTask's code
> 
> My point here is that even if you have someone else's code to get the general idea from, it is usually a good idea to dive into it and understand what the copied pieces of code do and why/for what reason. Right, Mikhail knows better what he needs to see in the result -- but everyone can make a mistake, so if you have no idea why something is done this way and not another, it may as well mean that it _is_ a mistake and here's a chance to fix it. Or not, but in order to find out you have to ask the question.
> 
> > (so the answer to your question is "no, I was not specifically evaluating this condition"). Should we change it?
> 
> And that's the thing that should definitely be changed -- the belief that others can do no mistake ;)

Sure, I agree with these two points. My "should we change it?" was meant as "So, you have an idea of some sort that suggests that we should change it? Let's hear it then." (if my guess above is correct, then you've already voiced the idea, but I missed it), not as "Why should we bother changing it if it was already decided by someone else?".

> > "why should both input and result values of these tasks be excluded from the statistics?" -- "because without input/output/both datasets no values - or only one of each pair - can be obtained, making the resulting calculation (for this particular task) impossible"
> 
> But we do not perform "calculation for particular task" -- it's two total sums and then the ratio of these values.

True, I've misworded my statement.

> Yet if there was only one task matching the `project-tag-format` combination -- that'd be a problem indeed...

Cases of having zero input/no tasks at all/etc are taken care of in the code, with `try/except` and `if`s. This can be improved, because the current version simply assigns 0 to some/all values, but at least in some form this protective mechanism does exist.

> I'd say that the original idea might have been as follows: if we know values only for output and add them to the total sum, the `total(output)` would contain more tasks than `total(input)` and dividing one by another would simply be incorrect (and vice versa). Does it sound logical, or is a bit of a stretch?

This is more or less what I've wanted to say above, with "particular task". Except that my logic was "we can't add unknown value to one of the two sums, which is wrong" instead of "if we add value only to one of the two sums, we upset the input-output equality, which is wrong", which is more or less the same. This returns us to the beginning, problem of distinguishing between two types of tasks and their values, a.k.a. how to separate "zero which is zero" from "zero which is unknown value".

[1] By abstract logic, it is also possible to have a task producing output from "nothing", with zero input but positive output. This case, however, shouldn't exist in here, in ATLAS data.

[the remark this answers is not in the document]
mgolosova: Fine, now I guess we get to the bottom of my original question.

---

>next question is "how can we distinguish between tasks A and B?"

where
>"incorrect" task A <...> can have output `[]` instead of expected `[2]` due to some error <...>, but a "correct" task B <...> can produce the same output `[]`, and there is nothing wrong with it.

My guess (and nothing but a guess):
* if dataset exists -- its files contain some metadata, so size on disks will not be zero;
* if task derives "empty set of data" from the input, it may not produce any dataset at all.

For now I suggest to leave this one aside to discuss it with Mikhail on occasion (at the meeting or by e-mail).

If you're OK with this, please add a note/card in Trello (to not forget about your question) and resolve this conversation.

mgolosova:
[REST API: статистика по эффективности деривации](https://trello.com/c/TknSBKGY/129-rest-api-%D1%81%D1%82%D0%B0%D1%82%D0%B8%D1%81%D1%82%D0%B8%D0%BA%D0%B0-%D0%BF%D0%BE-%D1%8D%D1%84%D1%84%D0%B5%D0%BA%D1%82%D0%B8%D0%B2%D0%BD%D0%BE%D1%81%D1%82%D0%B8-%D0%B4%D0%B5%D1%80%D0%B8%D0%B2%D0%B0%D1%86%D0%B8%D0%B8)

Utils/API/server/lib/dkb/api/storages/es/query/deriv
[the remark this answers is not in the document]
Evildoor: https://trello.com/c/bMcR3PkQ

Utils/API/server/lib/dkb/api/storages/es.py
Evildoor: > This function takes from ES first 500 output datasets that match the [query](https://github.com/PanDAWMS/dkb/blob/c5326669ce489478a35d8c096152446ba37281a2/Utils/API/server/lib/dkb/api/storages/es/query/output_formats) condition ("dataset has parent with given `ctag`"), and then aggregates received data to get the list. Right?
>

Yes. 

> My questions:
> 
> 1. Wouldn't it be better to allow ES to make the aggregation?

It would, will look into it.

>    Why:
>    
>    * for now there's a possibility to get incorrect result _(**Q**: when?)_

When there is more than 500 documents. As I've said in some other discussion, this may be enough for now, but is not enough in arbitrary case. ES aggregation, on the other hand, goes over all documents.

>    * ES can do it faster _(**Q**: why?)_

Because ES was designed to handle large amounts of data and do various things (including aggregations like this one) with them, while current mechanism is a simple Python code - it is not bad, but it is not specialized as well. The difference would be even more obvious if we were to upgrade the code to deal with the arbitrary case, where all the data had to be retrieved by pages, these pages stored somewhere, then the calculations performed over them...

> 2. Why only `ctag`?
>    (In this case) we seem to be interested only in formats for given `ctag-project` combination, and if this list is shorter than that of all formats for given `ctag` -- we're sending extra queries to ES that return no data, yet take some time to send request and get the response. It's not necessary that ES is locally available, so there's gonna be some additional cost for the network operations.

Also makes sense, will do.

Evildoor: It seems that I've overestimated the value generated by the use of aggregation - since we are still retrieving values (output formats) instead of statistics, 'size' limit remains in play. Still, it would take longer to reach 500 different output formats than 500 different datasets with parents with given ctags, and other advantages of the change remain.

Addition of `project` yields no visible result for now (for given `project-ctag` combination).

Utils/API/server/lib/dkb/api/__init__.py
```diff
@@ -10,7 +10,7 @@
 CONFIG_DIR = '%%CFG_DIR%%'
 
 
-__version__ = '0.2.dev20191203'
+__version__ = '0.2.dev20200129'
```
mgolosova: @Evildoor, just out of curiosity: why not today, but 0129? :)

Evildoor: Because 29 of January is the date of 82f4970, the last change (that is not update of the version). "Version = date of last real change" sounds better to me than "version = date of last real change + some time has passed before reviewer had an opportunity to review the changes and point out that version should be changed + some time has passed before developer could see the comment and make the change". Not that much better - after all, these "real" changes are also dependent on "when I have time to get to reviewing/fixing/discussing/etc." - but still. 

Also, in this way life would probably be simpler in a hypothetical case when someone looks through old commits (and wonders why there was a pause of N days between some activity and version update otherwise).

mgolosova: > "Version = date of last real change" sounds better to me than "version = date of last real change + some time has passed before reviewer had an opportunity to review the changes and point out that version should be changed + some time has passed before developer could see the comment and make the change".

It makes sense, so why not.

My thoughts on this idea, although, were almost opposite: "version = something close to the moment when the changes are getting to the production". Why: if we have multiple changes done in parallel branches, it'd be better to have version in production saying "when the last change was applied", not "when the last change was suggested" -- since the one suggested later may for some reason be applied earlier, and the next applied change may set the version to an earlier date... which is ridiculous.

Although in such case there will be conflicts, right? And if we solve them setting version to the date of 'conflict solving' moment, it'll do the job.

Evildoor: Aha, I see your point. Yes, I think that suggested procedure with different branches will do. And if it won't, then we would probably need to start thinking about more thoroughly planned version system.

PR #309, opened 2020-01-09
095 update, part 1.1: readability

Evildoor:
Part of the series of pull requests intended to replace #282.

Implement changes that are dealing with readability and do not affect the stage logic: descriptions, style consistency, etc.


Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -23,18 +23,23 @@
...
 
 ami_client = None
-PHYS_VALUES = [{"ami": "genFiltEff", "es": "gen_filt_eff"},
-               {"ami": "crossSection", "es": "cross_section"},
-               {"ami": "crossSectionRef", "es": "cross_section_ref"},
-               {"ami": "kFactor", "es": "k_factor"},
-               {"ami": "processGroup", "es": "process_group"},
-               {"ami": "mePDF", "es": "me_pdf"},
+# Field names in terms of AMI and ES schemes.
+PHYS_VALUES = [{'ami': 'genFiltEff', 'es': 'gen_filt_eff'},
```
mgolosova: Better to keep empty line between `ami_client` and `PHYS_VALUES`, for without one it looks like piece of code with  logically linked lines -- or group of independant, but similar one-line statements. Which it is not: both `PHYS_VALUES` and `ami_client` are, of course, global variables, but they have different purpose:
* `PHYS_VALUES` contains some reference data, required for specific logical operation. Its purpose id to move this data from code, making the latter more compact and avoiding the variable re-definition every time the function is called;
* `ami_client` stores a commonly used object. It is defined dynamically during the program execution and was introduced to avoid passing the object to every other function as parameter.

Also, these statemets are not one-liners -- so to group them as "similar independant statements", there should be a bit more than absence of blank line: e.g. common header ("Global variables definition") and unified descriptions for all variables defined in this block. Still I believe blank line would be good for readability, preventing things from bluring together. Example below is just to demonstrate what I'm talking about -- it is a bit tooo massive for a couple of variables, so the code becomes much less compact and may lose more, than gain.
```
# Global variables --->
# ---

# AMI client
ami_client = None

# Field names in terms of AMI and ES schemes
PHYS_VALUES = [{'ami': ..., 'es': ...},
               {'ami': ..., 'es': ...},
               {'ami': ..., 'es': ...},
               {'ami': ..., 'es': ...},
               {'ami': ..., 'es': ...}
               ]

# Allowed dataset formats, for which AMI should be queried
FILTER = ['AOD', 'EVNT', 'HITS']

# <--- Global variables
```

---

*Started to look into the PR a bit too early, so maybe it was gonna be changed later. But this part looks the same way in #282, so I decided not to remove the comment (since it was already written when I realized that I am too hasty ;) ).*

Evildoor: Good point, done.

> _Started to look into the PR a bit too early, so maybe it was gonna be changed later. But this part looks the same way in #282, so I decided not to remove the comment (since it was already written when I realized that I am too hasty ;) )._

Not, it was not planned. It's ok, thanks for the suggestion.

Evildoor:
Took a look at the code and the PR:

- Fixed a minor issue in README.
- Moved changes to `amiPhysValues()`' return here from #312. They are not really affecting the logic (see commit message), and moving them allows to pull over the description-related commit as well.
- Left 4bbd4dd in #312. While it's also a "cosmetic" commit, it is related to logic changes introduced in there, so it should be left as-is, or cut apart and amended to corresponding commits.

The only remaining problem with readability in here is `remove_tid()`'s description, but it gets handled in #312's as well(df80612), and there are logic changes in the function, so it should be left alone as well. Therefore, PR seems to be completed. Will remove [WIP] and request a review when reassession of #312 will be complete as well.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -23,18 +23,24 @@
...
+               {'ami': 'mePDF', 'es': 'me_pdf'},
                ]
 FILTER = ['AOD', 'EVNT', 'HITS']
 
 
 def main(argv):
-    """ Main program body. """
+    """ Main program body.
+
+    :param argv: arguments
```
mgolosova, line 41: Not very explicative description.. mb "command line arguments"? I mean, it's pretty obvious that function arguments are arguments, so this line looks like "this line must be here due to some stupid requirements, so it's here -- but no one said it must make any sense: as I said, the requirements are stupid."

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -59,43 +65,63 @@ def main(argv):
 
 
 def init_ami_client(userkey='', usercert=''):
-    """ Initialisation of AMI client into the global variable
+    """ Establish a connection to AMI.
+
+    Initialize the global variable ami_client with the resulting
+    client object.
```
mgolosova, line 71: Maybe add to this description `:raises:` section, as a clue to "under what circumstances it is expected to fail and how it will look"? I know we do not usually add this part, so "why here?"... but we had some discussions about "when to fail" (about this function and the similar one for Rucio, etc), and I remember myself extracting this summary from the code once or twice already, and I have a feeling it's not over yet.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: Done.
Note to self - if necessary, update #312 when this point is concluded: IIRC, there is at least one additional case for exception here.

Evildoor:
Travis complains about problems with variable names in code that is not touched in this PR. Weird, probably some update in Travis. Will look into said problems, but not in this PR.

mgolosova:
@Evildoor,
> Travis complains about problems with variable names in code that is not touched in this PR. Weird, probably some update in Travis. Will look into said problems, but not in this PR.

Not in Travis, in `pycodestyle`:
[Job #3013.1](https://travis-ci.com/github/PanDAWMS/dkb/jobs/323978925):
```
$ pip install pycodestyle
<...>
Successfully installed pycodestyle-2.5.0
```

[Job #3030.1](https://travis-ci.com/github/PanDAWMS/dkb/jobs/335571385)
```
$ pip install pycodestyle
<...>
Successfully installed pycodestyle-2.6.0
```

So I suggest to merge #356 ASAP and then merge `master` here -- since there's no other (good) way to merge this PR but to fix things, and #356 is much better than softening codestyle requrements by adding `E741` to ignore list or something like this.

Evildoor:
> @Evildoor,
> 
> > Travis complains about problems with variable names in code that is not touched in this PR. Weird, probably some update in Travis. Will look into said problems, but not in this PR.
> 
> Not in Travis, in `pycodestyle`:
> [Job #3013.1](https://travis-ci.com/github/PanDAWMS/dkb/jobs/323978925):
> 
> ```
> $ pip install pycodestyle
> <...>
> Successfully installed pycodestyle-2.5.0
> ```
> 
> [Job #3030.1](https://travis-ci.com/github/PanDAWMS/dkb/jobs/335571385)
> 
> ```
> $ pip install pycodestyle
> <...>
> Successfully installed pycodestyle-2.6.0
> ```
> 

Aha, I see, thank you.

> So I suggest to merge #356 ASAP and then merge `master` here

Done.

> -- since there's no other (good) way to merge this PR but to fix things, and #356 is much better than softening codestyle requrements by adding `E741` to ignore list or something like this.

True.

mgolosova:
@Evildoor, merged, thank you!

*Please remove the branch if not needed.*

PR #312, opened 2020-01-13
095 update, part 1.2: logic

Evildoor:
Part of the series of pull requests intended to replace #282.

Implement changes that affect and improve the stage logic, but aren't directly related to AMI team's feedback.

### DIFF
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -184,7 +184,10 @@ def amiPhysValues(data):
                               '--logicalDatasetName=%s' % container],
                              format='json')
     json_str = json.loads(res)
-    rowset = json_str['AMIMessage'][0]['Result'][0]['rowset']
+    try:
+        rowset = json_str['AMIMessage'][0]['Result'][0]['rowset']
+    except KeyError:
+        raise KeyError("Unexpected AMI response: %s" % json_str)
     if not rowset:
         sys.stderr.write("(WARN) No values found in AMI for dataset '%s'\n"
                          % data['datasetname'])
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -154,7 +154,13 @@ def process(stage, message):
     # 'data_format' list contains one of the allowed formats
     # or not set at all.
     if update or not formats:
-        amiPhysValues(data)
+        try:
+            amiPhysValues(data)
+        except DataflowException:
+            raise
+        except Exception:
+            stage.output_error("Failed to process dataset '%s'"
+                               % data['datasetname'], sys.exc_info())
         change_key_names(data)
     stage.output(pyDKB.dataflow.communication.messages.JSONMessage(data))
 
@@ -174,31 +180,26 @@ def amiPhysValues(data):
     if not container:
         return False
     ami_client = get_ami_client()
-    try:
-        res = ami_client.execute(['GetPhysicsParamsForDataset',
-                                  '--logicalDatasetName=%s' % container],
-                                 format='json')
-        json_str = json.loads(res)
-        rowset = json_str['AMIMessage'][0]['Result'][0]['rowset']
-        if not rowset:
-            sys.stderr.write("(WARN) No values found in AMI for dataset '%s'\n"
-                             % data['datasetname'])
-            return False
-        for row in rowset[0]['row']:
-            p_name, p_val = None, None
-            for field in row['field']:
-                if field['@name'] == 'paramName':
-                    p_name = field['$']
-                elif field['@name'] == 'paramValue':
-                    p_val = field['$']
-                if p_name and p_val:
-                    data[p_name] = p_val
-                    break
-        return True
-    except Exception as e:
-        sys.stderr.write("(ERROR) Failed to process dataset '%s': "
-                         "%r\n" % (data['datasetname'], e))
+    res = ami_client.execute(['GetPhysicsParamsForDataset',
+                              '--logicalDatasetName=%s' % container],
+                             format='json')
+    json_str = json.loads(res)
+    rowset = json_str['AMIMessage'][0]['Result'][0]['rowset']
+    if not rowset:
+        sys.stderr.write("(WARN) No values found in AMI for dataset '%s'\n"
+                         % data['datasetname'])
         return False
+    for row in rowset[0]['row']:
+        p_name, p_val = None, None
+        for field in row['field']:
+            if field['@name'] == 'paramName':
+                p_name = field['$']
+            elif field['@name'] == 'paramValue':
+                p_val = field['$']
+            if p_name and p_val:
+                data[p_name] = p_val
+                break
+    return True
 
 
 def change_key_names(data):
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -206,12 +206,14 @@ def change_key_names(data):
 def container_name(data):
     """ Retrieve container name from information about dataset.
 
-    As AMI GetPhysicsParamsForDataset works with containers,
-    we construct the container name from each dataset name,
-    removing the _tid{...} part
+    The container name is extracted from dataset name by removing
+    the '_tid...' part.
 
     :param data: dataset information, must contain 'datasetname' field
-    :return: dataset name without _tid => container name
+    :type data: dict
+
+    :return: container name if it was determined successfully, False otherwise
+    :rtype: str or bool
     """
     return re.sub('_tid(.)+', '', data['datasetname'])
 
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -157,8 +157,7 @@ def amiPhysValues(data):
     :return: True (update was successful) or False (otherwise)
     :rtype: bool
     """
-    dataset = data['datasetname']
-    container = remove_tid(dataset)
+    container = container_name(data)
     ami_client = get_ami_client()
     try:
         res = ami_client.execute(['GetPhysicsParamsForDataset',
@@ -204,17 +203,17 @@ def change_key_names(data):
     return data
 
 
-def remove_tid(dataset):
-    """ Remove TaskID (_tidXX) part from dataset name.
+def container_name(data):
+    """ Retrieve container name from information about dataset.
 
     As AMI GetPhysicsParamsForDataset works with containers,
     we construct the container name from each dataset name,
     removing the _tid{...} part
 
-    :param dataset: dataset name
+    :param data: dataset information, must contain 'datasetname' field
     :return: dataset name without _tid => container name
     """
-    return re.sub('_tid(.)+', '', dataset)
+    return re.sub('_tid(.)+', '', data['datasetname'])
 
 
 if __name__ == '__main__':
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -128,6 +128,9 @@ def process(stage, message):
     :rtype: bool
     """
     data = message.content()
+    if not isinstance(data, dict):
+        sys.stderr.write("(WARN) Cannot update non-dict data: %r\n" % data)
+        return False
     # 'data_format' field contains a list of strings,
     # e.g. ['DAOD_SUSY5', 'DAOD']
     formats = data.get('data_format', [])
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -161,6 +161,8 @@ def process(stage, message):
         except Exception:
             stage.output_error("Failed to process dataset '%s'"
                                % data['datasetname'], sys.exc_info())
+        # Do not put this into try/except above - any exception produced by it
+        # indicates a problem with the stage code that demands a full stop.
         change_key_names(data)
     stage.output(pyDKB.dataflow.communication.messages.JSONMessage(data))
 
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -181,8 +181,7 @@ def amiPhysValues(data):
                     p_val = field['$']
                 if p_name and p_val:
                     data[p_name] = p_val
-                    p_name, p_val = None, None
-                    continue
+                    break
         return True
     except Exception as e:
         sys.stderr.write("(ERROR) Failed to process dataset '%s': "
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -143,6 +143,7 @@ def process(stage, message):
     # or not set at all.
     if update or not formats:
         amiPhysValues(data)
+        change_key_names(data)
     stage.output(pyDKB.dataflow.communication.messages.JSONMessage(data))
 
     return True
@@ -182,7 +183,6 @@ def amiPhysValues(data):
                     data[p_name] = p_val
                     p_name, p_val = None, None
                     continue
-        change_key_names(data)
         return True
     except Exception as e:
         sys.stderr.write("(ERROR) Failed to process dataset '%s': "
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -186,8 +186,8 @@ def amiPhysValues(data):
     json_str = json.loads(res)
     try:
         rowset = json_str['AMIMessage'][0]['Result'][0]['rowset']
-    except KeyError:
-        raise KeyError("Unexpected AMI response: %s" % json_str)
+    except Exception:
+        raise Exception("Unexpected AMI response: %s" % json_str)
     if not rowset:
         sys.stderr.write("(WARN) No values found in AMI for dataset '%s'\n"
                          % data['datasetname'])
--- a/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
+++ b/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
@@ -158,6 +158,8 @@ def amiPhysValues(data):
     :rtype: bool
     """
     container = container_name(data)
+    if not container:
+        return False
     ami_client = get_ami_client()
     try:
         res = ami_client.execute(['GetPhysicsParamsForDataset',
@@ -215,7 +217,19 @@ def container_name(data):
     :return: container name if it was determined successfully, False otherwise
     :rtype: str or bool
     """
-    return re.sub('_tid(.)+', '', data['datasetname'])
+    if 'datasetname' in data:
+        dataset = data['datasetname']
+    else:
+        sys.stderr.write("(WARN) Required field 'datasetname' not found"
+                         " in data: %r\n" % data)
+        return False
+    if not isinstance(dataset, (str, unicode)):
+        sys.stderr.write("(WARN) Invalid type of 'datasetname' field:"
+                         " expected string, got %s.\n"
+                         "(==) Data:"
+                         " %r\n" % (dataset.__class__.__name__, data))
+        return False
+    return re.sub('_tid(.)+', '', dataset)
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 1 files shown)


Evildoor:
Rebased the branch to bring it up to date.

Evildoor:
Rebased again since some commits were moved into #309(see comment there).

Evildoor:
Split and merged commit about empty spaces into previous ones introducing the strings with said spaces.

Evildoor:
Merged two commits (minor text update and decision to stop processing) into test command commit.

Evildoor:
Added `raise` of `DataflowException` to 9741256. This commit was the one that caused problems that [this discussion](https://github.com/PanDAWMS/dkb/pull/282#discussion_r356632250) is about: before it, `amiPhysValues()` was called without try/except, called `init_ami_client()` that raised exception, it was caught by `ProcessorStage.run()`, and processing stopped while proper information about the problem was displayed. Addition of try/except caused the `DataflowException` to be caught as a problem with the first dataset, which resulted in stage still running and various unexpected messages. 

`DataflowException` is our own invention, and there are 4 cases of it in this stage. All of them are in `init_ami_client()`, produce reasonable debug messages, and are serious enough to stop the stage. Therefore, a simple `raise` seems to be enough for now.

Dropped the commit about trying to init client without credentials (ce3841d in old PR) as it was, indeed, an incorrect attempt to solve the issue.

Evildoor:
Rebased the branch to fetch new changes, updated test command commit according to [this discussion](https://github.com/PanDAWMS/dkb/pull/309#discussion_r425746706).

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -100,6 +102,14 @@ def init_ami_client(userkey='', usercert=''):
...
         raise DataflowException("Failed to initialise AMI client: certificate"
                                 " not provided or not found.")
+    try:
+        ami_client.execute('ListCommands')
+    except Exception as e:
+        sys.stderr.write("(ERROR) Failed to perform test command"
+                         " ListCommands. Are you sure you have"
+                         " a valid certificate?\nException: %s\n" % str(e))
+        raise DataflowException("Failed to perform test command"
+                                " ListCommands.")
```
mgolosova, line 112: When the stage reaches this point, the log looks this way:
```
[dkb@aiatlas171 095_datasetInfoAMI]$ ./amiDatasets.py -m s
2020-05-21 14:25:14 (INFO) (ProcessorStage) Configuration parameters:
<...>
(ERROR) Failed to perform test command ListCommands. Are you sure you have a valid certificate?
Exception: [SSL] PEM lib (_ssl.c:2597)
  File "./amiDatasets.py", line 253, in <module>
    main(sys.argv[1:])
  File "./amiDatasets.py", line 58, in main
    stage.CONFIG['ami']['usercert'])
  File "./amiDatasets.py", line 111, in init_ami_client
    raise DataflowException("Failed to perform test command"
pyDKB.dataflow.exceptions.DataflowException: Failed to perform test command ListCommands.
```

It does not look very friendly (for me, at least):
1. `Exception: [SSL] ...`  is not prepended with any prefix, which makes it look like a part of Python-generated log (just like the traceback info following it), not a "handled" log message.
1. The `(ERROR)` log line, even with the `Exception: ...`, is confusing: *whose* command is this test one? what was tested with it? Traceback gives a clue -- the exception was thrown in `init_ami_client`, but it would be better if one could get the meaning of the error without diving into it.
  I mean -- even thought the message *does* give a hint that one might want to look at the certificate files, it does *not* explain what happened (and what to check about these files except for their existence).

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -196,16 +219,33 @@ def change_key_names(data):
...
+    """ Retrieve container name from information about dataset.
+
+    The container name is extracted from dataset name by removing
+    the '_tid...' part.
 
-    As AMI GetPhysicsParamsForDataset works with containers,
-    we construct the container name from each dataset name,
-    removing the _tid{...} part
+    :param data: dataset information
+    :type data: dict
```
mgolosova, line 229: All we know about `data` after this description is that it is a dict; also, according to the line above:
```
    The container name is extracted from dataset name by removing
    the '_tid...' part.
```
-- we can assume that this dict must contain the dataset name; but there's no clue to how exactly this DS name can be provided.

I suggest to specify that there is at least one requirement to the `data`: DS name is supposed to be provided in the `'datasetname'` field -- as a part of `data` description, or where the "dataset name" is mentioned.

mgolosova, line 229: All we know about `data` after this description is that it is a dict; also, according to the line above:
```
    The container name is extracted from dataset name by removing
    the '_tid...' part.
```
-- we can assume that this dict must contain the dataset name; but there's no clue to how exactly this DS name can be provided.

I suggest to specify that there is at least one requirement to the `data`: DS name is supposed to be provided in the `'datasetname'` field -- as a part of `data` description, or where the "dataset name" is mentioned.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -139,7 +152,18 @@ def process(stage, message):
...
+            amiPhysValues(data)
+        except DataflowException:
+            raise
+        except SSLError:
+            stage.output_error("Failed to process dataset '%s'"
+                               % data['datasetname'], sys.exc_info())
+        except Exception:
+            e_type, e_value, e_trace = sys.exc_info()
+            sys.stderr.write("(ERROR) Failed to process dataset '%s': %s\n"
+                             % (data['datasetname'], e_value))
```
mgolosova, line 165: Why handle `SSLError`s and other `Exception`s differently?
As far as I can understand, the difference here is in traceback -- which is provided for `SSLError`, but not for other exceptions.

I couldn't invent a regular way to get to this point, so for testing I hardcoded invalid parameters to be passed to `ami_client.execute(...)` ([here](https://github.com/PanDAWMS/dkb/blob/933747d6eb6f80be59e2342949739dbbe1e85a20/Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py#L185)), and get the following log:
```
[dkb@aiatlas171 095_datasetInfoAMI]$ vi amiDatasets.py
# Alter ami_client.execute() parameters...
[dkb@aiatlas171 095_datasetInfoAMI]$ echo '{"datasetname": "ttt"}' | ./amiDatasets.py -m s
<...>
2020-05-21 15:48:26 (INFO) (ProcessorStage) Starting stage execution.
(ERROR) Failed to process dataset 'ttt': 'Result'
{"datasetname": "ttt"}
2020-05-21 15:48:26 (INFO) (ProcessorStage) Stopping stage.
[dkb@aiatlas171 095_datasetInfoAMI]$ vi amiDatasets.py
# Alter again...
[dkb@aiatlas171 095_datasetInfoAMI]$ echo '{"datasetname": "ttt"}' | ./amiDatasets.py -m s
<...>
2020-05-21 15:48:58 (INFO) (ProcessorStage) Starting stage execution.
(ERROR) Failed to process dataset 'ttt': No JSON object could be decoded
{"datasetname": "ttt"}
2020-05-21 15:48:58 (INFO) (ProcessorStage) Stopping stage.
```

Below I provide the traceback that (could) be output in these cases (and values of the variables that provoked the exceptions):
(1)
```
2020-05-21 15:51:27 (DEBUG) (ProcessorStage) Traceback (most recent call last):
(==)   File "./amiDatasets.py", line 157, in process
(==)     amiPhysValues(data)
(==)   File "./amiDatasets.py", line 192, in amiPhysValues
(==)     rowset = json_str['AMIMessage'][0]['Result'][0]['rowset']
(==) KeyError: 'Result'
```
`json_str` variable value:
```
{"AMIMessage" : [{"error" : [{"$" :"Command not found `AMIGetPhysicsParamsForDatase`."}],"executionTime" : [{"$" :"1.590068902701E12 "}]}]}
```

(2)
```
2020-05-21 15:53:49 (DEBUG) (ProcessorStage) Traceback (most recent call last):
(==)   File "./amiDatasets.py", line 157, in process
(==)     amiPhysValues(data)
(==)   File "./amiDatasets.py", line 191, in amiPhysValues
(==)     json_str = json.loads(res)
(==)   File "/usr/lib64/python2.7/json/__init__.py", line 338, in loads
(==)     return _default_decoder.decode(s)
(==)   File "/usr/lib64/python2.7/json/decoder.py", line 366, in decode
(==)     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
(==)   File "/usr/lib64/python2.7/json/decoder.py", line 384, in raw_decode
(==)     raise ValueError("No JSON object could be decoded")
(==) ValueError: No JSON object could be decoded
```
`res` variable value:
```
u'AMI#\nerror   : Command not found `AMIGetPhysicsParamsForDatase`.\n#'
```

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: Done.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: Fair points, added a prefix and a message about the test command being executed.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: > Why handle `SSLError`s and other `Exception`s differently?
> As far as I can understand, the difference here is in traceback -- which is provided for `SSLError`, but not for other exceptions.

`SSLError` was important because it allowed us to catch problems with authentication. Now, when a test command is performed before processing, it's not that important, so a special handling can be removed. On the other hand, normal `Exception`s may need more information: as can be seen in your example, `(ERROR) Failed to process dataset 'ttt': 'Result'` is not very informative.

Removed `SSLError` and moved its handling to `Exception`, to provide traceback and use `stage.output_error()`. Also changed the way said lack of `'Result'` is treated to display the (rightfully) suggested `json_str`.

Note - all these things really call for rebasing. Why add `SSLError` in one commit and remove it later in the other, in the same PR? And some other questionable points.
However, the rebase will be pretty deep, so I decided to do it later, when we'll conclude the discussion of this point.

> (2)
> 
> ```
> 2020-05-21 15:53:49 (DEBUG) (ProcessorStage) Traceback (most recent call last):
> (==)   File "./amiDatasets.py", line 157, in process
> (==)     amiPhysValues(data)
> (==)   File "./amiDatasets.py", line 191, in amiPhysValues
> (==)     json_str = json.loads(res)
> (==)   File "/usr/lib64/python2.7/json/__init__.py", line 338, in loads
> (==)     return _default_decoder.decode(s)
> (==)   File "/usr/lib64/python2.7/json/decoder.py", line 366, in decode
> (==)     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
> (==)   File "/usr/lib64/python2.7/json/decoder.py", line 384, in raw_decode
> (==)     raise ValueError("No JSON object could be decoded")
> (==) ValueError: No JSON object could be decoded
> ```
> 
> `res` variable value:
> 
> ```
> u'AMI#\nerror   : Command not found `AMIGetPhysicsParamsForDatase`.\n#'
> ```

This can also be caught, by checking whether the JSON was decoded correctly, but... You got this by changing the requested format in `res = ami_client.execute(...)`, right? If yes, then such check does not look necessary - we already check that AMI client works (by running test command), so we can expect its further responses to be coherent (in JSON if we request JSON). Unless something wrong happens in AMI after the test command, but before the request about a dataset, which looks like a far-fetched situation to me.

mgolosova: > Note - all these things (as well as some other ones that were done/discussed previously) really call for rebasing. Why add SSLError in one commit and remove it later in the other, in the same PR? And some other questionable points.
> However, the rebase will be pretty deep, so I decided to do it later, when we'll conclude the discussion of this point.

OK.

>> (2)
>> ```
>> 2020-05-21 15:53:49 (DEBUG) (ProcessorStage) Traceback (most recent call last):
>> <...>
>> (==) ValueError: No JSON object could be decoded
>> ```
>> `res` variable value:
>> ```
>> u'AMI#\nerror   : Command not found `AMIGetPhysicsParamsForDatase`.\n#'
>> ```
>
> This can also be caught, by checking whether the JSON was decoded correctly, but... <...> such check does not look necessary - we already check that AMI client works (by running test command), so we can expect its further responses to be coherent (in JSON if we request JSON). Unless something wrong happens in AMI after the test command, but before the request about a dataset, which looks like a far-fetched situation to me.

Right, the provided example situation was just to illustrate how some random error handling is gonna look like. As for the response format -- well... even if the AMI server sends a valid JSON document, I guess on its way to our client it still may be broken by some misfunction of the network, local resources etc. So no, I would not bet that we are guaranteed to get a valid JSON even if we asked for one and have previously received one.

I am not saying, however, that this situation must be specially handled here -- it is still a (supposedly) rare and accidental situation, so the traceback output from the outermost exceptions handling of the `ProcessorStage` object should be enough to say that we got something unexpected in the response. Should it become a regular problem one day -- then it will require a more accurate handling, but now it is just as optional as an accurate handling of dozens of other possible accidents.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -102,6 +102,9 @@ def init_ami_client(userkey='', usercert=''):
                          " create proxy.\n")
         raise DataflowException("Failed to initialise AMI client: certificate"
                                 " not provided or not found.")
+    sys.stderr.write("(INFO) Executing test command ListCommands to check"
```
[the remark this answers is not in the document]

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
Evildoor: > > > (2)
> > > ```
> > > 2020-05-21 15:53:49 (DEBUG) (ProcessorStage) Traceback (most recent call last):
> > > <...>
> > > (==) ValueError: No JSON object could be decoded
> > > ```
> > > 
> > > 
> > > `res` variable value:
> > > ```
> > > u'AMI#\nerror   : Command not found `AMIGetPhysicsParamsForDatase`.\n#'
> > > ```
> > 
> > 
> > This can also be caught, by checking whether the JSON was decoded correctly, but... <...> such check does not look necessary - we already check that AMI client works (by running test command), so we can expect its further responses to be coherent (in JSON if we request JSON). Unless something wrong happens in AMI after the test command, but before the request about a dataset, which looks like a far-fetched situation to me.
> 
> Right, the provided example situation was just to illustrate how some random error handling is gonna look like. As for the response format -- well... even if the AMI server sends a valid JSON document, I guess on its way to our client it still may be broken by some misfunction of the network, local resources etc. So no, I would not bet that we are guaranteed to get a valid JSON even if we asked for one and have previously received one.
> 
> I am not saying, however, that this situation must be specially handled here -- it is still a (supposedly) rare and accidental situation, so the traceback output from the outermost exceptions handling of the `ProcessorStage` object should be enough to say that we got something unexpected in the response. Should it become a regular problem one day -- then it will require a more accurate handling, but now it is just as optional as an accurate handling of dozens of other possible accidents.

Fair points. We leave it be for now then.

Evildoor:
Did the rebase to deal with SSL and pointless `try:except` by squashing together 4 commits into 6d45dcc:

- SSL handling added.
- Error processing (including SSL and Exception) is moved from `amiPhysValues()` into `process()` to use `stage.output_error()` for SSL handling, bare `try:except:raise` added to `amiPhysValues()`.
- Bare `try:except:raise` removed from `amiPhysValues()`.
- SSL handling is removed, `stage.output_error()` is now used for Exception handling.

Also, since we are rewriting history, squashed together some other commits in two situations where some logic was changed and then some "cosmetic" things were adjusted per discussion:

- Mention of 'datasetname' requirement into 2961640, where it becomes a requirement to a function argument.
- Changes to test command messages into the commit introducing test command (e9a907a).

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -184,7 +184,10 @@ def amiPhysValues(data):
                               '--logicalDatasetName=%s' % container],
                              format='json')
     json_str = json.loads(res)
-    rowset = json_str['AMIMessage'][0]['Result'][0]['rowset']
+    try:
+        rowset = json_str['AMIMessage'][0]['Result'][0]['rowset']
+    except KeyError:
```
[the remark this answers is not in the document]
Evildoor: Done, bbde8bc.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -184,7 +184,10 @@ def amiPhysValues(data):
...
+        raise KeyError("Unexpected AMI response: %s" % json_str)
```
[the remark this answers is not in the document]
Evildoor: Done, bbde8bc.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
```diff
@@ -139,7 +154,14 @@ def process(stage, message):
...
     if update or not formats:
-        amiPhysValues(data)
+        try:
+            amiPhysValues(data)
+        except DataflowException:
+            raise
+        except Exception:
+            stage.output_error("Failed to process dataset '%s'"
+                               % data['datasetname'], sys.exc_info())
+        change_key_names(data)
```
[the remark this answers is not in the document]
Evildoor: This is, indeed, the result of

> "as things were before"

But now, when you brought this point up... What can cause exception in `change_key_names()`?

1. `PHYS_VALUES` is not defined.
2. There is no `'ami'` in some `item` - which means that `PHYS_VALUES` was defined incorrectly.
3. Same with `'es'`.
4. `data` is not a `dict`.

First three points do not depend on `data` - if we run into them, then there is a problem with the stage, which means that it is logical to stop after the first instance [1], not process each dataset (while putting pressure on AMI) and then throw it away with "I can't process this!". The fourth one depends on `data`, but there is (as you pointed out) a check in `process()`, and if `process()` makes the check and then transforms `data` into something different... Again, looks like a problem with the stage that calls for stop.

So, unless I am missing something, the function turned out to be in the right place, even if it wasn't really intentional. It is possible for the function to become more complex in the future, where maybe there would be actual problems with given `data`, and then moving it into `try...except` would be reasonable... But not now.

[1] If we really care about this, then it would be better to check that `PHYS_VALUES` is defined, and defined properly, somewhere in the beginning, not during the processing of each dataset. But we do not.

Evildoor:
@mgolosova,
> I also have some doubts about [2961640](https://github.com/PanDAWMS/dkb/commit/29616409575ee92eb3a8667f7434c529033a91b8) (`Rework container name determination`). The function that returns the container name does not _need_ the rest of the dataset metadata -- in only needs the dataset name; and, for me -- more "atomic" functions look better. I mean, if one day someone needs to get container name for another data structure -- (s)he will have to transform the structure to the expected one and then transform it back to the desired one. But, on the other hand -- if there were more than one place in the code where we need to get the container name from the whole message -- this new version would look more logical... And since both "if"s are equally hypothetical, I'm not going to say that preparing code for one of them is better than for another; and preparing for both of them (by making some `get_container_name_from_ds_metadata` call the previously existing `remove_tid`, renamed to `get_container_name_from_ds_name`) will take more efforts, while there's no actual reason to think it's gonna help one day :) So I'm just saying -- leaving things as they are now is totaly OK.

These doubts sound reasonable, but I also have to agree that we lack data to properly determine the possibilities of need to extract name out of different structure and need to extract name out of message in a different place in the code. Hence - let's leave it be.

Utils/Dataflow/095_datasetInfoAMI/amiDatasets.py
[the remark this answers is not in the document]
mgolosova: I made this digging into the `change_key_names()` as well, and -- technically -- you're right.

My point here is that for now it is not clear from the code if author(s) thought about this moment, or it just "historically happened" -- which makes others like me (who always have doubts even about self-written code, not to mention someone else's one) stumble here and try to find out if there's a possible misfunction, or everything's (accidentally ;) ) fine.

It could be done more clear by something like this:

```
try:
    change_key_names(data)
except Exception, err:
    raise DataflowException("Stage internal failure.", reason=err)
```

However (I didn't check) most likely it will mask the original `err` and makes it harder to find out the reason... so maybe just point out that it is not an accident:

```
try:
   ...
except:
   ...

# Do not try/except this one -- any exception would indicate an internal
# functionality problem, meaning that the stage cannot be used until it's fixed
change_key_names(data)

stage.output(...)
return True
```

My version may be too wordy (even being much shorter than the first version ;) ), but it's just an idea -- if it makes any sense, it can be shortened/refined to your liking.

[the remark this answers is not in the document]
Evildoor: Added a comment.

mgolosova:
@Evildoor, thanjk you, approved and merged.

*Please remove the branch if no longer needed.*

PR #313, opened 2020-02-10
DF/025: fix ES queries (doc_type changed).

mgolosova:
At some point doc_type in the UC ES indices `jobs_archive_*` has changed, and now it is "_doc" (the default one, like "no specific type").
Same seem to come for the `tasks_archive_*` at some point.

There's no documents with `toths06_failed` field (based on the jobs data) in the DKB ES after `03-02-20 12:31:40`; and in UC ES for now there's available only jobs since `2019-11-13T00:00:01.000Z` (at least via Search API).

After this fix we should re-integrate missed data, but should be careful not to remove what we can't recover.


mgolosova:
@anastasiakaida, thank you for fast response with this review.
Merging now.

PR #318, opened 2020-02-12
DF/025: add possibility to pass ES auth parameters via config file.

mgolosova:
Main changes:
1. Stage 025 (Chicago ES):
  * stage can use a configuration file to get authentication parameters for the Chicago ES access;
  * default config file name is set to `/Utils/Dataflow/config/025.cfg`.
2. pyDKB library (`pyDKB.dataflow.stages`):
  * command line arguments default values can be reset during stage creation/configuration process;
  * these new default values can be ignored in case of `--skip` scenario (to avoid stage failure when it is not even supposed to do anything that would require those stage-aware default values).


Utils/Dataflow/025_chicagoES/stage.py
```diff
@@ -353,12 +362,16 @@ def main(args):
...
+    stage.set_default_arguments(config=os.path.join(base_dir, os.pardir,
+                                                    'config', '025.cfg'))
+
     stage.process = process
 
     exit_code = 0
     exc_info = None
     try:
         stage.configure(args)
+        init_es_client(stage.CONFIG['ChicagoES'])
```
Evildoor: Commit message states that this commit merely "allows" something - however, this line means that the ES client is now initialized before starting stage, rather than when `get_es_client()` is called for the first time. This explicit initialization has nothing to do with "allowing" and changes in `init_es_client()`, right?

mgolosova: @Evildoor, thank you for pointing to it: this part of the changes was not only unexplained, but also lead to some problems with the stage `--skip` mode.

I have fixed those issues (0f70241...bc8645d) and reworded the initial commit (0f70241) -- please take a look.

Evildoor: This particular issue - `init_es_client()` not mentioned in commit logs - was fixed. Will comment on other commits and close this conversation as resolved some time later.

Evildoor:
I have my doubts about 5a17523. The description is rather long, but it only explains the logic behind the change declared in the first line. This means that "why?" is more or less covered, but then comes "how?".  Changes to the code are quite extensive:

- `AbstractStage` now stores default arguments values.
- `AbstractStage.add_argument()` now includes a relatively long hand-made procedure of determining the name of the argument, and storage of the argument into `_default_args`.
- Due to previous point `add_argument()` is now separated into three parts, each one having a comment describing it (on first glance the comment about help formatting looks like something that does not belong in the commit).
- `AbstractStage` now has a function for resetting arguments' values to defaults.
- `--skip` is now non-hardcoded value in `ProcessorStage`.
- `ProcessorStage` now has a function to invoke the same one from its `super`.
- `ProcessorStage` now resets arguments to defaults when in `skip` mode.
- Something else that I've missed?

I believe that such changes should be documented more thoroughly, and maybe even split into separate commits (non-hardcoded `--skip` is the prime candidate here). Some of them can, possibly, be skipped because they are strongly implied by the existing description (well, yeah, we implement reset mechanic, this means we implement a method which does exactly that), but not the other ones.

Other commits look fine.

mgolosova:
@Evildoor 
> I have my doubts about [5a17523](https://github.com/PanDAWMS/dkb/commit/5a17523e3d344186fee80c358d13f8d336daf562). The description is rather long, but it only explains the logic behind the change declared in the first line. This means that "why?" is more or less covered, but then comes "how?". Changes to the code are quite extensive...

...and they are clearly seen in the commit diff. So why one would prefer to read about them in some human-oriented description insead of looking into real changes? I mean, if the one's concern is *how* something was done, diff is the most reliable source of information. What diff *can't* do is to explain why it was done -- which is, as you said, completely covered by the commit log message.

> I believe that such changes should be documented more thoroughly, and maybe even split into separate commits (non-hardcoded `--skip` is the prime candidate here).

Non-hardcoded `--skip` option is required only to have possibility to check if this option is passed to the stage prior to `ArgumentParser.parse_args()` call. By itself it has no sense, has it?

As for the other changes -- they also look completely interlinked for me, and moving any of them outside of this commit would produce a lot of questions like "why are you doing this, it's not used anywhere, and I don't even see how it can be used?"

However I may be wrong: the commit *is*, for sure, much more complicated than some single-line "typo fix" one, and I might just not see a way to simplify it. I think I'd ask a third opinion on this question, just to make sure I'm not "juggle the facts" to keep the commit untouched :) 

Maybe @IATkachenko could say something (or anyone else at your choise)?

IATkachenko:
@Evildoor, commit log should not answer to "How?" question. Code do this.
GitHub [Best Practice](https://github.com/trein/dev-best-practices/wiki/Git-Commit-Best-Practices), for example, recommends (in section about commit message):

> The body of your message should provide detailed answers to the following questions: – What was the motivation for the change? – How does it differ from the previous implementation?

And this is really useful by following reasons: 
1. If you are using revision system for debugging, you goes from code and MUST know motivations of changes. If you are changing code that was changed by someone else, motivation part of commit log will answer to question "What the hell is going on?!", "What result author tries to reach".  And with this knowledge you may apply own changes.
1. If you are using revision system as a source of information about changes in system, than technical details is not interested for you.

If code changes is not clear, than you should focus attention on this parts of code, keeping in mind commit motivation, to make changes in commit clean and clear.

Few steps away for better understanding.
You have asked to add information "ProcessorStage now resets arguments to defaults when in skip mode." in commit log.  Why this information should be in commit log, from your point of view, and how this information in commit log will be used? (My suggestion is: you are going to use commit log as a "Developer manual", with description of all methods in code, but commit log evolution log, not manual. It is nod designed to be manual). How do you use commits log?
For me scenarios is following: if we have some unexpected behaviour, after changes, we check when and what was changed, found this commit by changes in code (at this step we see that ProcessorStage now resets arguments to defaults when in skip mode) and get from commit log motivation why this changes were made. After that we may make a decision about fixes.

@mgolosova, for me changes in add_argument(AbstractStage.py) is not clear: you collecting default values for arguments and must know argument name. It is clear. But at same time you are changing kwargs['dest'] if argument not start with '-'. This is new code and it seem not related to ability to skip some arguments for "--skip" stages. From the other hand, if 'dest' in kwargs, no defaults for other arguments will be setted.
Other functions is clean enough for me, but this part makes me doubt.

IATkachenko:
@mgolosova, OK, kwargs['dest'] is the "user-specified argument name". After this all clear.
So... 'dest' is not good name for this parameter in general (but should not be changed in this set of changes).

mgolosova:
@IATkachenko,
> OK, kwargs['dest'] is the "user-specified argument name". After this all clear.

OK, thank you!

> So... 'dest' is not good name for this parameter in general (but should not be changed in this set of changes).

Unfortunately, it is not up to us: https://docs.python.org/2.7/library/argparse.html#dest

mgolosova:
> @mgolosova [force-pushed](https://github.com/PanDAWMS/dkb/compare/bb411ef17edd38acbb8449f7bdcd1fe4e380b066..7084f7076381e65e0ac01df9aecea30a1647fa4f) the [025-auth-config](https://github.com/PanDAWMS/dkb/tree/025-auth-config)  branch from bb411ef to 7084f70 [now](https://github.com/PanDAWMS/dkb/pull/318/#event-3137082141)

Tried to solve conflicts via GUI, but ended up with missed newline at the end of file. Fixed it by (re-)merging master manually and force-puch.

Evildoor:
@mgolosova 

> @Evildoor
> 
> > I have my doubts about [5a17523](https://github.com/PanDAWMS/dkb/commit/5a17523e3d344186fee80c358d13f8d336daf562). The description is rather long, but it only explains the logic behind the change declared in the first line. This means that "why?" is more or less covered, but then comes "how?". Changes to the code are quite extensive...
> 
> ...and they are clearly seen in the commit diff. So why one would prefer to read about them in some human-oriented description insead of looking into real changes? I mean, if the one's concern is _how_ something was done, diff is the most reliable source of information. What diff _can't_ do is to explain why it was done -- which is, as you said, completely covered by the commit log message.

In many cases the code, indeed, is the place to look if one wants to understand the changes' logic. The question is, is the code simple and straightforward enough, and how much time and effort is needed to look through and understand it. I spent a considerable amount of both on this commit, hence my comment. Obviously, there is a possibility that this "considerable amount spent" is stemming from me being slow-witted rather than commit being too complicated, but I'm not sure if that's the case.

> > (non-hardcoded `--skip` is the prime candidate here).
> 
> Non-hardcoded `--skip` option is required only to have possibility to check if this option is passed to the stage prior to `ArgumentParser.parse_args()` call. By itself it has no sense, has it?

I don't know. It's a possibility to change the `--skip` option, its introduction raises questions:

- Is the change necessary for achieving the goal of checking for option's presence?
- Is it the best way to do so?
- Is the possibility usable for anything else? 

The change's existence and form means that you did your research and arrived at "yes", "yes" and "no", respectively. I would need to do some research of my own to support or contest these answers, I will do so if necessary. And anyone (possibly including you or me several years later) looking through the commit where the `--skip` change is not even mentioned would have to stumble upon it, understand what's going on (maybe stop there, but maybe not, it depends), ask these questions, and try to answer them. Is this OK? I think we can save them some work by mentioning the change and saying a few words about why it is done this way, and not the other.

> As for the other changes -- they also look completely interlinked for me, and moving any of them outside of this commit would produce a lot of questions like "why are you doing this, it's not used anywhere, and I don't even see how it can be used?"

We had a discussion about a large, interlinked commit at least once (email, "PR#255 (Stage 017)", point 5). An important thing from it, in my opinion, is "Главное -- чтобы было наглядно". Since I've started this discussion, this point is, most likely, not completely upheld here.

> However I may be wrong: the commit _is_, for sure, much more complicated than some single-line "typo fix" one, and I might just not see a way to simplify it.

I already made a list to illustrate my point, I can try to make it into an addition to commit message or a plan for separation of the commit into multiple ones, if that would be of use. But this, I think, should wait until we complete this discussion.

Evildoor:
@IATkachenko 

> @Evildoor, commit log should not answer to "How?" question. Code do this.

See my answer to @mgolosova above. Yes, the commit message should not answer to "how does the code work" - for this, indeed, we have code itself, comments inside code, documentation, and other things. It can, however, answer to "how does the commit change the code": what changes it consists of, why these changes are necessary, why the changes are done in this particular way, etc. Obviously, a one-liner can usually do just fine with "Do X" message, but more complex commits may be better with more clarification.

> GitHub [Best Practice](https://github.com/trein/dev-best-practices/wiki/Git-Commit-Best-Practices), for example, recommends (in section about commit message):
> 
> > The body of your message should provide detailed answers to the following questions: – What was the motivation for the change? – How does it differ from the previous implementation?

The second question seems to include my point, at least to me. So does example 3 of good practice - if I understand correctly, it is not entirely about the motivation behind the commit: it also underlines a few key points of how the new feature works, points out that it is not turned on by default, and even states that "something will be done later in another commit". 

> And this is really useful by following reasons:
> 
> 1. If you are using revision system for debugging, you goes from code and MUST know motivations of changes. If you are changing code that was changed by someone else, motivation part of commit log will answer to question "What the hell is going on?!", "What result author tries to reach".  And with this knowledge you may apply own changes.
> 2. If you are using revision system as a source of information about changes in system, than technical details is not interested for you.

No objections here.

> If code changes is not clear, than you should focus attention on this parts of code, keeping in mind commit motivation, to make changes in commit clean and clear.

If you are talking about changes not being clear enough when the developer is making them, then this is exactly what I'm suggesting - to "make changes clean and clear" by either adding some explanation of their structure or by splitting them into several smaller commits.

If you are talking about changes not being clear to someone who is reading them after they've already been made by the developer - in theory, we can forget about github's purpose and guidelines, bunch all eight (at this moment) commits of this PR into a single one and leave only the PR title as the description. Technically, it would still work, because the changes are more or less connected, they are working towards a single goal, and the reader would be able, with enough time and effort, understand them by looking at the code. But the current situation, with separate commits containing smaller logical changes and some explanations, is much better, right? I'm merely asking to take it a step further and improve/break down one of these commits.

> Few steps away for better understanding.
> You have asked to add information "ProcessorStage now resets arguments to defaults when in skip mode." in commit log. Why this information should be in commit log, from your point of view, and how this information in commit log will be used?

Uh, no. I made a list of descriptions of changes, including this one, to illustrate my point. Then I said that such extensive changes should be documented more thoroughly or even split into separate commits, and I stand by these words. But I'm not saying that this line must be added into commit message as-is.

> (My suggestion is: you are going to use commit log as a "Developer manual", with description of all methods in code, but commit log evolution log, not manual. It is nod designed to be manual).

It should be clear from my reasoning above that I'm not going to do this. On this point I agree with you - commit log is not a manual for the code.

> How do you use commits log?

Rather broad question, but... When I look at a commit, I want to understand what it does and why (for what reason, or what goal) it was done. Sometimes, if there are problems with the commit and not everything is clear, I would also like to know, why it was done in exactly this way. And in such situation, replacing "Do X" with more elaborate "Do X, do it using Y because Z, also pay attention to ABC" would help me to either find the problem faster, or say "right, this makes sense to me" and decide that the problem is, more likely, somewhere else, and go looking for it.

> For me scenarios is following: if we have some unexpected behaviour, after changes, we check when and what was changed, found this commit by changes in code (at this step we see that ProcessorStage now resets arguments to defaults when in skip mode) and get from commit log motivation why this changes were made. After that we may make a decision about fixes.

No objections here either.

IATkachenko:
> The second question seems to include my point, at least to me. So does example 3 of good practice - if I understand correctly, it is not entirely about the motivation behind the commit: it also underlines a few key points of how the new feature works, declares that it is not turned on by default, and even states that "something will be done later in another commit".

I guess that second part of 3rd example describes architecture changes: before this commit libvirt did not query the guest capabilities and it is important outside the commit. And so on. Just because there is no reason to put it in commit log according to guide.

>  But I'm not saying that this line must be added into commit message as-is.

Of course not as-is. I've read all you suggestions as: "Information about this should be documented in commit log in some way". And it is strange for me, because I can't see any reasons why it may be useful in commit log after merge.
 
>  But the current situation, with separate commits containing smaller logical changes and some explanations, is much better, right?  I'm merely asking to take it a step further and improve/break down one of these commits.

Could not agree with you in this case. Small changes is good, but all related changes must go together. Check my example in [IATkachenko/dontSplit](https://github.com/IATkachenko/dontSplit/commits/master). This repository was created as you wish: every commit contain isolated changes, but as a result we got final commit that broke everything and looks like correct commit that should not break code. 
You may object that review of c21198b will find mistake, but it does not work in real life: nothing was broken, you can't  see "red flags" and have to very carefully check code to find logical mistake.

GitHub is allowing you to see squashed changes in PR, but it goes as separate commits to the repository, all information about connections between commits is only at GitHub and hidden in logs (you may find it, but it require some additional efforts). So every single commit must be self-sufficient and PR must link commits that are solving one problem.

> > How do you use commits log?
> Rather broad question, but... When I look at a commit, I want to understand what it does and why (for what reason, or what goal) it was done. 

So, original commit log should answer to this questions:
  - what: "allow resetting custom default values to 'safe' ones on skip."
  - why: "If stage is skipped, some things that are supposed to be properly configured for real processing, may not be configured..." and so on

And then, as a reviewer, you are starting interesting "How?" and need more explanation (sorry, I have to add "after merge" to my original question):
>Sometimes, if there are problems with the commit and not everything is clear, I would also like to know, why it was done in exactly this way. And in such situation, replacing "Do X" with more elaborate "Do X, do it using Y because Z, also pay attention to ABC" would help me to either find the problem faster, or say "right, this makes sense to me" and decide that the problem is, more likely, somewhere else, and go looking for it.

This information is important at review stage, but useless after merge: if we will find a bug in code, then we should know why code was added (commit log) and understand how it works (code itself, not diff from commit, because code may be changed several times in different commits, but have a bug now, after all changes), not why it works in this way (just because "this" is wrong due to bug). Developer should not explain why he/she solve the problem in this way. Just because every one have own "best way to solve this". You may like suggested way, or not. If you don't like solution, but it works -- offer own, as a refactoring part.

The best way to get more information about "Why this changes were done in this way" -- inline comments in code due to review: it will gave all answers that you need, but will not overload commit log for future reading.

mgolosova:
@Evildoor,

I have a really long comment below, but as we talk about such philosophical matters as "art of good commits" -- I guess it was inevitable.

> In many cases the code, indeed, is the place to look if one wants to understand the changes' logic. The question is, is the code simple and straightforward enough, and how much time and effort is needed to look through and understand it.

Unfortunately, some code is just complex, because it takes care of complex matters. And it will be voluminous and interlinked, no matter what: I don't know how to describe a complex object in simple statements, not referring to each other.

Say, it is not about this commit, but about the `pyDKB.dataflow.stage` submodule as a whole. Is it "simple and strightforward enough"? I'm afraid not. And I am not sure if it can be done (much) more "simple and strightforward": there are just a lot of things one should grasp. Concept of the "stage" itself and how is it used; why does it have to do something about command line parameters; *what* it does about them and *when* it happens; which parameters are "native" for the stage and how their values affect the stage's behaviour; etc. And the best way to drag through all of this (and the only one I know) -- is to read the code. Read it "far and wide", following chains of function calls to see how something works -- or looking up for mentioning of variable/function to see where and how something is used to answer "what is it for and why does it looks like this".

Yes, it is a difficult reading. But would it be any better if every method was wrapped into a separate file, to make the piece of code one is reading at one time smaller? My answer is: no. The reader would have to jump form file to file far too many times, which is really, really annoying.

I'm not saying *all* the code should be gathered in a single file -- of course not. But as for the heavily interlinked pieces -- it's better to have them all together.

Back to commits: splitting this commit into multiple smaller pieces, for me, is very similar to placing pieces of code into different files. It would be easier to read each one of then -- but much harder to understand how all of them work together and why in the end it gives the claimed result.

> I spent a considerable amount of both *(time and efforts)* on this commit, hence my initial comment. Obviously, there is a possibility that this "considerable amount spent" is stemming from me being slow-witted rather than commit being too complicated, but I'm not sure if that's the case.

Huh, now you're saying things like "c'mon, tell me I'm too dumb to understand your code" ;) But as far as I can tell, you managed through it -- so no, obviously you're not.

I see two possible reasons why this commit took so much time and efforts to get on with:
* it is complex;
* the code is not documented well enough.

I can do nothing about the first item: it is the best solution I ended up with, that gave me the result I needed (which was to "allow resetting custom default values to 'safe' ones on skip"). I don't have any better, simpler and more elegant way to achieve this result -- which does not, of course, mean that my solution *has* any of these "good, simple and elegant" properties. Any suggestions that will make the code better are welcome; as well as any comments that point to author's mistake, carelessness or possible side effects that author might not seen.

As for the second -- I am sorry. I really tried to do my best but it might not be enough. E.g., as @IATkachenko pointed out after he got through this commit -- things would be much more readable if I put [this part](https://github.com/PanDAWMS/dkb/commit/5a17523e3d344186fee80c358d13f8d336daf562#diff-f30fbc13e9f48a63fd86bb3b51c82ae1R152-R171) into a separate function. Which is a fair point -- and worth doing as a code refactoring within this PR, for it makes the code better.
There could be other parts that are not properly documented, or have not very self-explaining names so you had to guess what I meant to say: please let me know and I'll add comments or rename items.

But what you are asking about -- to chop the commit into smaller pieces -- in my point of view, will not make a complex thing simpler. It will only complicate its comprehension as a whole.

You also said:
> I believe that such changes should be documented more thoroughly

This almost sounds in the spirit of "the code is not documented well enough" (except that originally you're suggesting to "document" things in the commit log message). So we can proceed from this point: but, again, I need you to tell me what was not clear enough in the code, so I could try to clarify things.

> > Non-hardcoded `--skip` option is required only to have possibility to check if this option is passed to the stage prior to `ArgumentParser.parse_args()` call. By itself it has no sense, has it?
> 
> I don't know. It's a possibility to change the `--skip` option...

Well, if you *really* needed to change it -- you could do it even without moving it to a separate variable. Now it's easier -- but the leading underscore (`weak "internal use" indicator` (c) [PEP-8](https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles)) in the variable name is kind of telling: "you know that you are not supposed to change this value, right?"

> ...its introduction raises questions:
> 
> * Is the change necessary for achieving the goal of checking for option's presence?

Check all occurences of  `_skip_option` name in the commit log: the place at which it's used will most probably answer you "why it was moved to a separate variable". Maybe it was too early, and it would be better to check for `"--skip"` in the list of the command line arguments and in the next commit -- replace both occurrences of `"--skip"` with the variable; but, on the other hand -- it would be the only occurence of `"--skip"` in the first commit. Leading to even more questions: "what is it? is it a newly invented for this exact purpose option? or what?" -- and necessity to look into code outside this commit's diff.

> * Is it the best way to do so?

That's the way I've chosen -- which automatically means that I believe it to be. I may be wrong: but however many words I write about my choice it won't make me see it. The only way is if you as a reviewer tell me it's not and explain why you think so.

> * Is the possibility usable for anything else?

I don't get the question, actually... I mean -- what does it have to do with this exact change? It can be used for something else -- fine; when some one needs to -- he/she will use. It can not -- fine; I don't need anything else here.

I would understand a question like "wouldn't it introduce any problems?" -- which, in fact, gets us back to choosing the best way to achieve the goal. If it introduces problems -- it is not the best way and I wouldn't choose it if I knew. Maybe I just didn't see the problems -- well... this code review is our last hope.

> The change's existence and form means that you did your research and arrived at "yes", "yes" and "no", respectively. I would need to do some research of my own to support or contest these answers, I will do so if necessary.

Wait, are you telling that author must write down and explain all the solutions he/she made writing the code? Er... content of this exact commit was re-written during the development more than once: I started to write one way, realized it's not a good idea for this and that reason, took another approach, after few steps saw it also leads to poor results -- so I stepped back and looked for better ideas. That's the development process. If I was a better developer, I might choose this way without second thought and wouldn't even think about other possibilities that are "obviously" worse than this one.

You have your doubts and question my decision -- that's absolutely right thing to do; if you didn't you wouldn't be able to do the review. But instead of asking "why did you do it this way?" -- please, ask "why didn't you do it this or that way (which I believe would be better for this and that reason)?" Just asking "why" and getting the answer won't make much difference: whatever happened in the author's head, it won't tell you if the result is good enough to go on with.

Yes, it means that you have to look for different solutions of the already solved problem -- or at least for issues that the author's choise might introduce. It's not easy. But it's what reviewers do.

> And anyone (possibly including you or me several years later) looking through the commit where the `--skip` change is not even mentioned would have to stumble upon it, understand what's going on (maybe stop there, but maybe not, it depends), ask these questions, and try to answer them. Is this OK? I think we can save them some work by mentioning the change and saying a few words about why it is done this way, and not the other.

This commit will be viewed for a specific reason. If the reason is related to this `--skip` option -- than no explaination in the commit log would work better than checking what really happaned to `--skip` here. For whatever is said -- it is words; they may be correct, but still are not as accurate as the code *and* one will still have to look into the code. And if the words are saying right things, while some of these things are somehow missed in the code -- it will take even longer to understand "but why, it says everything should be fine, but it isn't!" I mean -- there will be two sources of just the same information about "what happened to `--skip`". But one of them is 100% accurate, no matter was in the author's head at the moment (as it shows you what *exactly* happened), while another is a "loose adaptation" (which can be really loose, with a souce of "probabbly useful for...", "I also thought to do it $this_way, but...").

If the reason, why the one is checking this commit, is in something else -- than it does not actually matter, why `--skip` was moved to a separate variable. And reading about it in commit log message might even have the opposite effect: if it is mentioned, than it is important to understand the commit as a whole, so it worse some extra attention. Which it may not deserve.

> > As for the other changes -- they also look completely interlinked for me, and moving any of them outside of this commit would produce a lot of questions like "why are you doing this, it's not used anywhere, and I don't even see how it can be used?"
> 
> We had a discussion about a large, interlinked commit at least once (email, "PR#255 (Stage 017)", point 5). An important thing from it, in my opinion, is "Главное -- чтобы было наглядно". Since I've started this discussion, this point is, most likely, not completely upheld here.

Your quote from my e-mail is taken out of the context. Still a bit shortened, but more complete version:

ср, 19 июн. 2019 г. в 14:53, Marina Golosova <golosova.marina@gmail.com>:
  > Если логическое изменение слишком большое для одного коммита, ничто не
  > запрещает оформить его и как несколько коммитов. Но по их **summary** должно быть
  > понятно, что все они -- **последовательные этапы одного изменения**. Как это
  > **обозначить** -- дело вкуса. **Главное -- чтобы было наглядно.** Чтобы
  > вылетев по истории изменений на один из коммитов, можно было предположить,
  > что для полной картины может быть нужно заглянуть в соседние коммиты.

In other words: what you're quoting was said about commit summary, the first line of a commit log; and what I emphasised as "important to make clear" -- was about marking interlinked commits as "Do X: pt.1", "Do X: pt.2", "Do X: pt.3", etc. And, let's be honest: the changes in PR #55 we were discussing -- they contained hundreds of lines of added and removed code. Here we are talking about (I counted) 76 changed lines, of which only 50 are "pure code changes", without comments (note: I counted empty lines as "code" ones as well, for without them it is not readable at all).

> > However I may be wrong: the commit _is_, for sure, much more complicated than some single-line "typo fix" one, and I might just not see a way to simplify it.
> 
> I already made a list to illustrate my point, I can try to make it into an addition to commit message or a plan for separation of the commit into multiple ones, if that would be of use.

My summary:
* I believe that splitting this commit into pieces will make more damage than good;
* I am willing to fill blanks in the code where it is not clear enough what's going on -- but I need to know where is this "where";
* I *may* add to commit log message a brief explaination of what happened [1], if you have a suggestion -- but I would prefer *not* to retell the commit's diff content there;
* I will do some refactoring as additional commits (unless there's a really good reason to make force-push that will leave this discussion "without a leg to stand on") -- at least the part with that too-long-code-for-a-single-action will be moved to a separate function.


[1] For example:
```
Now `*Stage` keeps all pre-defined default values for command line
parameters, specified during the `argparse.ArgumentParser`
configuration. These values may be used as 'safe' ones in case of
skip-stage scenario instead of the user-defined defaults provided with
`ignore_on_skip` option set to `True`.
```

Evildoor:
@mgolosova, @IATkachenko 
While I still have some doubts about the commit in question's current state, your logic also makes sense, so I think that we can conclude this discussion by proceeding with these changes:

> My summary:
> 
> - I believe that splitting this commit into pieces will make more damage than good;
> - I am willing to fill blanks in the code where it is not clear enough what's going on -- but I need to know where is this "where";
> - I may add to commit log message a brief explaination of what happened [1], if you have a suggestion -- but I would prefer not to retell the commit's diff content there;
> - I will do some refactoring as additional commits (unless there's a really good reason to make force-push that will leave this discussion "without a leg to stand on") -- at least the part with that too-long-code-for-a-single-action will be moved to a separate function.

Should be enough to deal with those remaining doubts of mine.

There are some points in your last comments that I'd like to answer, but only for sake of understanding each other, not deciding the code's further changes. Will do so later.

@mgolosova, I'll provide (before doing the previous line, since this is more important) suggestions for points 2 and 3 of the list.

mgolosova:
[Stage 025 (ChicagoES): использование конфиг-файла с параметрами аутентификации](https://trello.com/c/K2NLWKWV/167-stage-025-chicagoes-%D0%B8%D1%81%D0%BF%D0%BE%D0%BB%D1%8C%D0%B7%D0%BE%D0%B2%D0%B0%D0%BD%D0%B8%D0%B5-%D0%BA%D0%BE%D0%BD%D1%84%D0%B8%D0%B3-%D1%84%D0%B0%D0%B9%D0%BB%D0%B0-%D1%81-%D0%BF%D0%B0%D1%80%D0%B0%D0%BC%D0%B5%D1%82%D1%80%D0%B0%D0%BC%D0%B8-%D0%B0%D1%83%D1%82%D0%B5%D0%BD%D1%82%D0%B8%D1%84%D0%B8%D0%BA%D0%B0%D1%86%D0%B8%D0%B8)

Evildoor:
> > My summary:
> > 
> > * I believe that splitting this commit into pieces will make more damage than good;
> > * I am willing to fill blanks in the code where it is not clear enough what's going on -- but I need to know where is this "where";
> > * I may add to commit log message a brief explaination of what happened [1], if you have a suggestion -- but I would prefer not to retell the commit's diff content there;
> > * I will do some refactoring as additional commits (unless there's a really good reason to make force-push that will leave this discussion "without a leg to stand on") -- at least the part with that too-long-code-for-a-single-action will be moved to a separate function.
>
> ...
>
> @mgolosova, I'll provide (before doing the previous line, since this is more important) suggestions for points 2 and 3 of the list.

- Add a code comment explaining non-hardcoded skip.
- Code uses term "safe default value" several times, and it might be unclear. Add something like your example from [1], but into code rather than commit message. Also, expand the example a bit to illustrate their purpose - "for example, it is possible to specify a default path to config file, but when the stage is in skip mode, this default value may cause problems by demanding this config to be present, while it's actually unnecessary because no actual processing will happen", or something like that. Current commit message contains some words about this - if this suggestion is accepted, they can be cut out (or moved into code) to avoid duplication.

mgolosova:
> @mgolosova [force-pushed](https://github.com/PanDAWMS/dkb/compare/65ac720ca24fb80378899d05ecb681a8706e16be..7c090964125f284e420511cb166f8688cdb5d0df) the 025-auth-config branch from 65ac720 to 7c09096

Rebased the PR branch on master to resolve conflicts with the `pyDKB` version number.

mgolosova:
@Evildoor,
> Add a code comment explaining non-hardcoded skip.

It will look (and already looks) this way:
```
+    # Name of the option that switches stage into 'skip' mode
     _skip_option = '--skip'
```

I mean, I can't add here any explaination about "why did author need this value in a variable" (which was the main point of our discussion). If I do, it will look wierd: no other variable is explained in terms of "why is it a variable", right?

> Code uses term "safe default value" several times, and it might be unclear.

Fair point: these values are "safe" by designation, not by their origin, the real meaning and content.
I reworded it to "original" values, which looks less confucing to me.

> Add something like your example from [1], but into code rather than commit message. Also, expand the example a bit to illustrate their purpose - "for example, it is possible to specify a default path to config file, but when the stage is in skip mode, this default value may cause problems by demanding this config to be present, while it's actually unnecessary because no actual processing will happen", or something like that. Current commit message contains some words about this - if this suggestion is accepted, they can be cut out (or moved into code) to avoid duplication.

Not sure that adding examples of usage and purpose into the code would be a good idea. First of all, it is about "how author uses it", not about "what is it" -- while the latter is much more important for the one reading the code. After all, an example of "how it is used" is usually known (or why else someone would dig into it) -- and one do not need to know *all* the usage examples to understand one of them. While "what is it" reqiuires careful tracking of all changes of the variable value, and having this information stated in a plain text can really help.

Second -- one day the added functionality and variables may become useful for some other purposes, and in this case the detailed description of a specific usage example might mask all the others, so that one might think that there are *no* (and may not be) any other usage scenarios at all.

---

From the library user's point of view (the one who uses the library), the option [`ignore_on_skip`](https://github.com/PanDAWMS/dkb/pull/318/files#diff-40439c64c338c1790d2233e1f5ea1c41R194-R195) looks (for me) explained clear enough.
Docstrings of other newly introduced methods/functions also do explain "what does this function do"; for what purpose it will be used -- is up to the one who uses it. All the other business (how these methods interact, how this or that information is stored, etc) is not really important.

From the stage developer's point of view (the one who developes the stages) -- the most popular use-cases will, of course, be described as "best practices" (like the one related to the stage config file in the [Installation/Development Guide](https://docs.google.com/document/d/1I8LlE3CmpA7_vfAdJcGuGKwbPmcU8XOE3JEH8XEuEU4/edit#)). So here comes the usage examples and purpose illustrations you're talking about.

From the library developer's point of view... well, the library developer should study the code itself before making any changes. In this case it doesn't matter "for what it is supposed to be used"; it is more important to understand "what it does in fact". *(Of corse it would be super-nice to have module tests, that would give the developed something but the understanding to control if new changes break something or not, but.....)*

Utils/Dataflow/pyDKB/dataflow/stage/AbstractStage.py
```diff
@@ -153,6 +166,28 @@ def set_default_arguments(self, **kwargs):
         """
         self.__parser.set_defaults(**kwargs)
 
+    def reset_default_arguments(self, args=None):
+        """ Reset default argument values to the original ones.
+
+        Original default value is a value, passed to the ``add_argument()``
```
Evildoor, line 172:
```suggestion
        Original default value is a value passed to the ``add_argument()``
```

Evildoor:
> @Evildoor,
> 
> > Add a code comment explaining non-hardcoded skip.
> 
> It will look (and already looks) this way:
> 
> ```
> +    # Name of the option that switches stage into 'skip' mode
>      _skip_option = '--skip'
> ```

Maybe just "Option that ...", without "name"? It's an actual string, including two hyphens.

> > Code uses term "safe default value" several times, and it might be unclear.
> 
> Fair point: these values are "safe" by designation, not by their origin, the real meaning and content.
> I reworded it to "original" values, which looks less confucing to me.

Looks good, thank you.

> 
> > Add something like your example from [1], but into code rather than commit message. Also, expand the example a bit to illustrate their purpose - "for example, it is possible to specify a default path to config file, but when the stage is in skip mode, this default value may cause problems by demanding this config to be present, while it's actually unnecessary because no actual processing will happen", or something like that. Current commit message contains some words about this - if this suggestion is accepted, they can be cut out (or moved into code) to avoid duplication.
> 
> Not sure that adding examples of usage and purpose into the code would be a good idea. 
>
> ...

Alright.

mgolosova:
@Evildoor,
> > ```
> > +    # Name of the option that switches stage into 'skip' mode
> >      _skip_option = '--skip'
> > ```
> 
> Maybe just "Option that ...", without "name"? It's an actual string, including two hyphens.

It is a string, indeed: and that's why it is a "name", not "option". I mean, "option" is something to be chosen/changed/switched/... -- and "Option that switches..." would call for values `True`/`False`. For me, at least.

Also: this comment is immediately followed by the value assignment, which gives the reader a clue to what author means by 'option name'; and should it be changed, it is supposed to pass through the review process. And change like this:
```
-    _skip_option = '--skip'
+    _skip_option = 'idle'
```
would look suspicios, wouldn't it? If it's consistent with the rest of the changes -- I see no problem here. If it's not -- I want to believe that if not the author, then at least a reviewer will detect it.

Evildoor:
> > > ```
> > > +    # Name of the option that switches stage into 'skip' mode
> > >      _skip_option = '--skip'
> > > ```
> > 
> > 
> > Maybe just "Option that ...", without "name"? It's an actual string, including two hyphens.
> 
> It is a string, indeed: and that's why it is a "name", not "option". I mean, "option" is something to be chosen/changed/switched/... -- and "Option that switches..." would call for values `True`/`False`. For me, at least.
> 
> Also: this comment is immediately followed by the value assignment, which gives the reader a clue to what author means by 'option name'; and should it be changed, it is supposed to pass through the review process. And change like this:
> 
> ```
> -    _skip_option = '--skip'
> +    _skip_option = 'idle'
> ```
> 
> would look suspicios, wouldn't it? If it's consistent with the rest of the changes -- I see no problem here. If it's not -- I want to believe that if not the author, then at least a reviewer will detect it.

If I understand you correctly, you mean that one can mistake "(command line) option (that is used for certain purpose and shouldn't be touched without need, and without review)" for "(script) option (set here, in the code, that can be changed to better adapt the script to given situation / hardware / user preferences / etc)", like

    debug=False

that can be set to `True` to test something, and then set back to `False`.

This is a fair point, but my suggestion was coming from different logic: "name of command line option", in my understanding, differs from "command line option" itself. So, if option is `--skip`, then its name would be `skip`, without hyphens. [1]
It can be solved by something like

    # Name of the option that switches stage into 'skip' mode
    _skip_option_name = 'skip'
    _skip_option = '--' + _skip_option_name
    ...

but changing the comment looks simpler and more logical. Since we have this possible confusion that you pointed out, maybe change it to something like "Command line option that ..." or "Command line argument that ..."?

[1] I may be wrong on this point, and then my suggestion can be disregarded. But [here](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html), for example, we have: "Long options consist of ‘--’ followed by a name made of alphanumeric characters and dashes."

mgolosova:
@Evildoor, 
> > > > ```
> > > > +    # Name of the option that switches stage into 'skip' mode
> > > >      _skip_option = '--skip'
> > > > ```
> > > 
> > > Maybe just "Option that ...", without "name"? It's an actual string, including two hyphens.
> > 
> > "Option that switches..." would call for values `True`/`False`. For me, at least.
> > Also: this comment is immediately followed by the value assignment, which gives the reader a clue to what author means by 'option name'.
> > <...>
> > If it's consistent with the rest of the changes -- I see no problem here. If it's not -- I want to believe that if not the author, then at least a reviewer will detect it.
> 
> If I understand you correctly, you mean that one can mistake "(command line) option (that is used for certain purpose and shouldn't be touched without need, and without review)" for "(script) option (set here, in the code, that can be changed to better adapt the script to given situation / hardware / user preferences / etc)", like
> 
> ```
> debug=False
> ```
> 
> that can be set to `True` to test something, and then set back to `False`.

No, what I meant is that "option" may be taken as something that deserves changing in code, like:
```
my_stage._skip_option = True
```

As for review: no changes in code, generally, should be done without it (except for local usage and/or tests, in which case the one making changes is the only person who's responsible for any consequences).

> my suggestion was coming from different logic: "name of command line option", in my understanding, differs from "command line option" itself. So, if option is `--skip`, then its name would be `skip`, without hyphens. [1]
> It can be solved by something like
> 
> ```
> # Name of the option that switches stage into 'skip' mode
> _skip_option_name = 'skip'
> _skip_option = '--' + _skip_option_name
> ...
> ```

...which, I'm sorry, for me looks ugly and pointless -- unless, of course, somewhere in the code we actually *need* to add somewhere this string, `"skip"` without hyphens, dynamically (and for some reason can not just strip hyphens in place). Which is not the case.

> changing the comment looks simpler and more logical. Since we have this possible confusion that you pointed out, maybe change it to something like "Command line option that ..." or "Command line argument that ..."?
> 
> [1] I may be wrong on this point, and then my suggestion can be disregarded. But [here](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html), for example, we have: "Long options consist of ‘--’ followed by a name made of alphanumeric characters and dashes."

You are right -- conceptually. But still I do not see how any of the suggested (by both of us) variants is any better -- less confusing and not less readable -- than the currently used one. So I suggest us to leave things as they are for now, and if any of us wants to change it -- to do it in a separate single-commit PR, providing us with a more appropriate space for further discussion than this PR. Which, if I may remind you, is one of reasons to keep #348 postponed.

Evildoor:
> > my suggestion was coming from different logic: "name of command line option", in my understanding, differs from "command line option" itself. So, if option is `--skip`, then its name would be `skip`, without hyphens. [1]
> > It can be solved by something like
> > ```
> > # Name of the option that switches stage into 'skip' mode
> > _skip_option_name = 'skip'
> > _skip_option = '--' + _skip_option_name
> > ...
> > ```
> 
> ...which, I'm sorry, for me looks ugly and pointless -- unless, of course, somewhere in the code we actually _need_ to add somewhere this string, `"skip"` without hyphens, dynamically (and for some reason can not just strip hyphens in place). Which is not the case.

This wasn't meant as a real suggestion worth considering, more like an illustration of "what if we really want to define an option __name__ here while disregarding common sense and other things" - sorry, probably should've underlined this better. So yes, it is indeed ugly and pointless.

> 
> > changing the comment looks simpler and more logical. Since we have this possible confusion that you pointed out, maybe change it to something like "Command line option that ..." or "Command line argument that ..."?
> > [1] I may be wrong on this point, and then my suggestion can be disregarded. But [here](https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html), for example, we have: "Long options consist of ‘--’ followed by a name made of alphanumeric characters and dashes."
> 
> You are right -- conceptually. But still I do not see how any of the suggested (by both of us) variants is any better -- less confusing and not less readable -- than the currently used one. So I suggest us to leave things as they are for now, and if any of us wants to change it -- to do it in a separate single-commit PR, providing us with a more appropriate space for further discussion than this PR. Which, if I may remind you, is one of reasons to keep #348 postponed.

Ok.

mgolosova:
@Evildoor,
> Update: oh, there is still [a small change suggested](https://github.com/PanDAWMS/dkb/pull/318#discussion_r410958111). Will reapprove after it's handled then.

Yeah, totally missed it behind our discussion. Fixed now, thank you!

mgolosova:
@Evildoor, thank you! Merged.

PR #329, opened 2020-02-27
Fix a typo.

Evildoor:
This mapping is the only place where the field name lacks "s" at the end.
Everywhere else - in Stage 009, data samples, DKB's ES - the other version
is used. I presume that this is a typo, and the actual mapping used in our
ES was "fine-tuned by a file" (or not touched at all, given how
ES's dynamic mapping works).

### DIFF
--- a/Utils/Elasticsearch/mapping/tasks.mapping
+++ b/Utils/Elasticsearch/mapping/tasks.mapping
@@ -130,7 +130,7 @@
         "core_count": {
           "type": "short"
         },
-        "conditions_tag": {
+        "conditions_tags": {
           "type": "keyword",
           "ignore_above": 2048
         },


mgolosova:
@Evildoor, it is a typo in fact, except that the field *must* have been named `conditions_tag`, without `s` in the end (after the JEDI task parameter `conditionsTag`).

I'm gonna approve this PR for it makes the mapping consistent with what is in ES and it is a good thing to do.
But also I added a task in [Trello](https://trello.com/c/7jM2SPNU) to think about renaming it back to `tag` when we have this a chance, during reindexing of all our data due to the global changes in the ES scheme.

mgolosova:
@IATkachenko, could you please take a look at this PR's Travis check status?
I found something here: https://github.com/travis-ci/travis-ci/issues/10204 -- but I can't see what they are talking about in the branch protection settings...

IATkachenko:
Our "continuous-integration/travis-ci"  is now deprecated. I'll check how we may replace it.

mgolosova:
@IATkachenko, somehow it looks like tho problem was temporary -- now I turned the `continuous-integration/travis-ci` on back, and found a button to restart Travis build in its interface. We'll continue to use this version until figure out how to manage the Travis CI app version.

Thank you!

PR #343, opened 2020-03-19
pyDKB/df/messages: fix issue with changeable message content.

mgolosova:
If message internals ("original", "decoded" or "encoded" data) are
represented by changeable objects (like `dict` or `list`), one might do
something like this (for JSON messages):

```
data = msg.content()
data['new_field'] = 'some_value'
some_function(msg)
```

...and, unintentionally changing the original message content, pass to
the `some_function()` the changed message.

To avoid this situation, `Message` methods (`getOriginal()`, `content()`,
`decode()`, `encode()`) now return copies of the internals. In some cases it
may be too paranoic (like `content()` creating a copy of a copy of
`self.decoded`), but... within the `AbstractMessage` we have no
guarantee that a specific message class has a correct `decode()` method
implementation.


mgolosova:
[pyDKB...message: исправить работу с телом сообщения](https://trello.com/c/JVcdkaH3/111-pydkbmessage-%D0%B8%D1%81%D0%BF%D1%80%D0%B0%D0%B2%D0%B8%D1%82%D1%8C-%D1%80%D0%B0%D0%B1%D0%BE%D1%82%D1%83-%D1%81-%D1%82%D0%B5%D0%BB%D0%BE%D0%BC-%D1%81%D0%BE%D0%BE%D0%B1%D1%89%D0%B5%D0%BD%D0%B8%D1%8F)

PR #320, opened 2020-02-17
DF/091: improve error handling

mgolosova:
Changes in error handling.
1. If Rucio returns `DatasetIdentifierNotFound`, set `deleted` to
   `True` (previously was treated as any other `RucioException`).
2. In case of `RucioException` mark message as "incomplete" -- this
   record will be written to ES in "update" mode, not "insert"
   (previously `deleted` and `bytes` would be set to `True` and `-1`).
3. If some of required fields can not be extracted from Rucio (or have
   value `null`), they are left unset (previously would be set to
   `None`.
4. If in the result message some fields of those that were supposed to
   be added at this stage are missed, the message is marked as
   `_incomplete`.

Additional changes:
* Stage 019 is set to always use `update` action for incomplete messages [1];
* fixed minor bug in logic: when output datasets being processed, all datasets were dropped as soon as single one had problems with ES indexing service fields definition;
* Stage 091: refactoring;
* Stage 091: samples update.

[1] It was not done initially since we wanted to first check how many "incomplete" records we're gonna have before make a decision: should records be "updated" on the fly, or maybe run some separate process that would update records marked with `_update_required` once a week or so.
But now the goal is to make data integration more reliable: as for statistics, we may teach 069 to count "update" records... anyway it should start counting records sooner or later, so why not.

---

What to check:
- general logic: things should really work as it is said below:
  - [x] @Evildoor;
  - [x] @mgolosova;
- behaviour when there's no Rucio errors:
  - [x] records are generated without `_incomplete` marker;
  - [x] -//-
- behaviour in case of Rucio errors:
  - **`rucio.client` failed to load:** *(new item)*
    - [ ] **stage stops with error message indicating problems with `rucio.client` import;**
    - [x] -//-
  - invalid credentials [2]:
    - [x] proper error message;
    - [x] -//-
    - [x] ~output message is marked as `_incomplete` ***([To Be Discussed](https://github.com/PanDAWMS/dkb/pull/320#issuecomment-606628096))***~ *(replaced with the one below)*;
    - [x] interrupt stage execution;
    - [x] -//-
  - dataset not found:
    - [ ] ~proper error message~ *(not found == removed, nothing to worry about)*;
    - [x] output message is marked as `_incomplete`;
    - [x] -//-
- interaction with Stage 019 [3]:
  - [x] not "incomplete" messages are treated as usual (decorated with "insert" action JSON);
  - [x] -//-
  - [x] "incomplete" messages are decorated with "update" action JSON;
  - [x] -//-
[...]

PR #345, opened 2020-03-19
pyDKB: improve message incomplete marker handling.

mgolosova:
Changes in messages "incompleteness" handling [1]:
* for complete messages we'll never see `"_incomplete": False` as a part of a message, however many times it was reset to True/False;
* no service field `_incomplete` in the message `content()` (for it is a *service* one, which is not a part of the content);
* if message content is not a dict, but some other JSON object, we don't know how to mark it as "incomplete" -- so an error will be produced;
* messages are "decoded" now as soon as they are created [2] -- so if an input message contained the `_incomplete` marker, it will be detected, used to set the corresponding property of the message and removed from the message's content (to reappear there after the "encoding" operation) [3].

[1] Mostly applicable to JSON messages, for TTL ones has no special way to propagate information about their incompleteness further. And perhaps they don't need to, as they are used to convert JSON ones into Virtuoso-friendly format; and by that moment we already should get all the metadata and information about their "incompletenets", which, at max, will be turned into an additional TTL message that sets some corresponding property of an ontological object... or something like this.

[2] But it does not mean that test cases 39-40 are useless: who knows what will change next time? So it's better to make sure that explicit decoding still works as it is supposed to. After all, under some circumstances this "double decoding" may as well lead to an incorrect/unexpected behaviour...

[3] This part made me think of #320 (and any other code that makes use of "incompleteness" mechanism): whatever is done with messages, **no stage should set the marker to `False`** -- to avoid erasing "incompleteness" information from previous stages. Unless, of course, we have an additional stage that knows how to check if message is complete or not (e.g. by presence of all fields that are supposed to be there). But currently we don't, so -- for now -- there's no exceptions fron this rule.
BTW, the #320 is fine (in this regard).


mgolosova:
Resolve `[WIP]` status for all related PRs are successfully merged into `master`.
Current branch was reset on the actual master to get rid of octopus merge in its history (and bring to it all minor fixes that might be done during the review process of the mentioned PRs).

@Evildoor, please take a look at the updated version and (re-)approve it if everythin's fine.

PR #353, opened 2020-03-27
DF: readme update

mgolosova:
1. Add introductory section with some general info about what's in the directory.
2. Update sub-headers design (originally they were not unified, and the one of presented designs I decided to follow appeared to be too heavy).

### DIFF
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -81,7 +81,7 @@ output one according to a set of rules common for all the stages.
 
 ./XXX_StageName/
           -- Dataflow stage (see below).
-             Stages are named with three leading digits (XXX), usualy referred
+             Stages are named with three leading digits (XXX), usually referred
              to as stage ID. But the number assignment logic is long gone, so
              it's safe to think about them simply as unique IDs.
 
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -80,9 +80,9 @@ for all the stages.
   utils/    -- misc tools
 
 ./XXX_StageName/
-	  -- Dataflow stage (see below).
-	     Stages are named with three leading digits (XXX), usualy referred
-	     to as stage ID. But the number assignment logic is long gone, so
+          -- Dataflow stage (see below).
+             Stages are named with three leading digits (XXX), usualy referred
+             to as stage ID. But the number assignment logic is long gone, so
              it's safe to think about them simply as unique IDs.
 
 
--- a/Utils/Dataflow/README
+++ b/Utils/Dataflow/README
@@ -6,9 +6,9 @@
 * General information *
 =======================
 
-====================
-* 1. Main concepts *
-====================
+
+1. Main concepts
+----------------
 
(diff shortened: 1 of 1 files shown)


Evildoor:
In general, I approve these changes (in non-github sense, though we'll probably get to that soon too). Some comments:

- In main concepts, "Dataflow" and "stages" are more or less covered, but "messages" is on the "less" side. Maybe replace "Messages" with "data", and add more text in the end, something like 'Data is organized into "Messages" - individual portions encoded in one of the formats: JSON, TTL.'?
- "... set of rules common for all the stages", without comma.

mgolosova:
@Evildoor,
> In main concepts, "Dataflow" and "stages" are more or less covered, but "messages" is on the "less" side. Maybe replace "Messages" with "data", and add more text in the end, something like 'Data is organized into "Messages" - individual portions encoded in one of the formats: JSON, TTL.'?

I just didn't wasn to dive too deep into the definitions. Making them self-consistent and complete would take quite an effort -- and poduce quite a lot of symbols that would be irrelevant here (for the section is not about full and complete conceptual description -- it is about "what's going on, in general"). So I just added a short explaination -- please take a look, if it's explicative enough (or only short).

Evildoor:
> @Evildoor,
> 
> > In main concepts, "Dataflow" and "stages" are more or less covered, but "messages" is on the "less" side. Maybe replace "Messages" with "data", and add more text in the end, something like 'Data is organized into "Messages" - individual portions encoded in one of the formats: JSON, TTL.'?
> 
> I just didn't wasn to dive too deep into the definitions. Making them self-consistent and complete would take quite an effort -- and poduce quite a lot of symbols that would be irrelevant here (for the section is not about full and complete conceptual description -- it is about "what's going on, in general"). So I just added a short explaination -- please take a look, if it's explicative enough (or only short).

Yes, I'm ok with this version.

Utils/Dataflow/README
```diff
@@ -2,7 +2,89 @@
...
-*** NOTE: stages are to be named with three leading digits (XXX_StageName). ***
+=======================
+* General information *
+=======================
+
+
+1. Main concepts
+----------------
+
+"Dataflow" is a series of "Stages", applied to a stream of data in a specific
```
Evildoor:
```suggestion
"Dataflow" is a series of "Stages" applied to a stream of data in a specific
```

 Pretty much the same as "set of rules common for all the stages".

Utils/Dataflow/README
```diff
@@ -2,7 +2,89 @@
...
+
+./run/    -- dataflow process management files (start/stop/status/...).
+
+./test/   -- misc tools for this or that functionality testing:
+  pyDKB/    -- tests for main pyDKB usage scenatios;
+  utils/    -- misc tools
+
+./XXX_StageName/
+          -- Dataflow stage (see below).
+             Stages are named with three leading digits (XXX), usualy referred
```
Evildoor:
```suggestion
             Stages are named with three leading digits (XXX), usually referred
```

PR #361, opened 2020-06-03
Change Dataflow directory layout.

mgolosova:
Stage directories are moved to dataflow-dedicated subdirectories:
*  `Dataflow/data4es`
*  `Dataflow/docs4virtuoso`

Stages configuration files (`Dataflow/config`) are moved together with stages (to `Dataflow/<df_name>/config`).

DF management scripts (`Dataflow/run/...`) are also moved (to `Dataflow/<df_name>/run`), but symlinks are created instead (for compatibility).

Common libraries are left in Dataflow/, but symlinks are created:
*  `Dataflow/<df_name>/shell_lib`
*  `Dataflow/<df_name>/pyDKB`

NOTE: after this PR any command, used to manage metadata integration process (specifically, `data4es`) keeps on working as if nothing has changed.

---

Known issues.

1. ~**Stage 009 (Oracle Connector)**~ ***(fixed)***.
  Path to SQL queries is specified via config file, and it is eighter absolute or relative to the config file location; but corresponding files are moved together with the stage, while the config is supposed to stay as-is.
  ***Update**: will be fixed in #363 (after which this PR will require rebase or merge from `master`).*
  ***Update**: done.*


mgolosova:
> mgolosova force-pushed the df-layout branch from 6c154a2 to d8b2ef6

Multiple forced pushes rebased this branch on the latest `master`.

PR #363, opened 2020-06-05
DF/009: change query configuration structure.

mgolosova:
This change makes the config file less dependent on the stage source directory location.

Queries themselves can be quite complicated, so usually they are considered to be more a part of the stage itself, rather than its configuration. So whenever the stage directory moved (e.g. the repository is checked out to different location), the configuration becomes invalid and it is not good.

Now instead of the full (or relative to the configuration file) path to the queries sources, one is supposed to configure query names and type -- and the stage will look for them in the stage source directory: `$STAGE_DIR/query[/$query_type]/$query_name`

The `$query_type` parameter is optional and is introduced for cases when the single stage is used in different scenarios (here: `data4es` integration process and process for consostency check of the integrated by this process data).

NOTE: currently the stage implementation hardly relays on the query names ('tasks' and 'datasets'), so changes in query names should be done with caution.

---

Working configurations after this change are:

* **`data4es` process**:
  ```
  [query]
  type: data4es
  names: tasks,datasets

  [process]
  mode: SQUASH
  ```
* **`data4es-consistency-check` process**:
  ```
  [query]
  type: data4es-consistency
  names: tasks

  [process]
  mode: PLAIN
  ```

*(`process.mode` value wasn't changed, but is provided here as a reminder -- it must be set to `PLAIN` when only one query is in use)*


mgolosova:
@witxka, please take a look at this PR.
It alters ths integration process configuration parameters, and if applied without configuration update -- will break things. Is it OK to apply it *and* change configs simultaneously, or shold I think about making it less stressful?

mgolosova:
[Stage 009: изменить конфигурацию SQL queries](https://trello.com/c/ttGpe8SH/175-stage-009-%D0%B8%D0%B7%D0%BC%D0%B5%D0%BD%D0%B8%D1%82%D1%8C-%D0%BA%D0%BE%D0%BD%D1%84%D0%B8%D0%B3%D1%83%D1%80%D0%B0%D1%86%D0%B8%D1%8E-sql-queries)

witxka:
> @witxka, please take a look at this PR.
> It alters ths integration process configuration parameters, and if applied without configuration update -- will break things. Is it OK to apply it _and_ change configs simultaneously, or shold I think about making it less stressful?

Dear @mgolosova I do not think that there are any problems with such things in the current deployment environment. No need to worry.

mgolosova:
@witxka, OK, thank you!

Utils/Dataflow/config/009.cfg.example
```diff
@@ -2,10 +2,16 @@
...
-# absolute (or relative to config file) path to tasks query
-tasks: %__009_query_dir__%/prodsys2ES.sql
-# absolute (or relative to config file) path to datasets query
-datasets: %__009_query_dir__%/datasets.sql
+# Query type (e.g. DF process for which it is used)
+# To be used if the stage is to be used in multiple scenarios
+# with different query sets.
+# The value will be mapped to queries subdir in the stage
+# source directory; if not specified, the queries directory
+# itself will be used
```
Evildoor, line 10:
```suggestion
# itself will be used.
```

PR #364, opened 2020-06-08
pyDKB/InputStream: update descriptions.

Evildoor:
Initial idea was to merely fix a "retrun", but since it's a typo in a description that will be, sooner or latter, rewritten in a proper form, that "sooner or later" may as well be "now".


Utils/Dataflow/pyDKB/dataflow/communication/stream/InputStream.py
```diff
@@ -38,6 +38,16 @@ def reset(self, fd, close=True, force=False):
         """ Reset current stream with new file descriptor.
 
         Overrides parent method to reset __iterator property.
+
+        :param fd: new file descriptor
+        :type fd: file
+        :param close: if True, close the old file descriptor
+        :type close: bool
+        :param force: if True, force the reset of iterator
```
mgolosova, line 46: I think it'd be nice to mention the difference between "forced" and "normal" reset in the description. Kind of this:

```
        :param force: if True, force the reset of iterator
                      (normally, it won't be reset if the passed new
                       descriptor is the same as the old one)
```

Evildoor: Done.

Evildoor:
Force-push because I forgot to rebuild the docs.

mgolosova:
@Evildoor, thank you, merged.

*Please remove the branch if no longer needed.*

PR #366, opened 2020-06-10
DF: allow single dataflow writing to two different ES indices.

mgolosova:
For now it is pretty much hardcoded: stage 019 (esFormat), that prepares data for upload, is only aware of 2 indices: "tasks" (for tasks metadata) and "progress" (for tasks processing progress metadata). If no index specified in the input data, "tasks" index is used.
There's also a "default" index -- it will be used if a desired index is not configured in the provided (or default) config file.

It may be improved later, e.g. when PHP script starts reading the configuration INI file by itself -- but for now that's enough.

ES configuration file previously contained only one index name ("ES_INDEX" variable); it is used by multiple stages so it is left in the config file to keep existing functionality. It can not be set to the
"value of ES_INDEX_TASKS" like this:

  ES_INDEX=$ES_INDEX_TASKS

Although it will be OK for shell scripts that use this file by simply sourcing it, non-shell ones have to "manually" parse it, and "$ES_INDEX_TASKS" will be considered as a plain string, not variable
value.

---

NOTE: after this PR all stages but **Stage 019** will continue to use the `ES_INDEX` parameter, while **Stage 019** will ignore it.

### DIFF
--- a/Utils/Dataflow/019_esFormat/esFormat.php
+++ b/Utils/Dataflow/019_esFormat/esFormat.php
@@ -6,7 +6,10 @@ function exception_error_handler($errno, $errstr, $errfile, $errline ) {
 set_error_handler("exception_error_handler");
 
 # Default values.
-# ES index where the documents should be indexed/updated.
+# Alias of ES index to be used when input data does not provide one
+$DEFAULT_INDEX_ALIAS = 'tasks';
+# Name of ES index where the documents should be indexed/updated,
+# if unknown alias is used
 $DEFAULT_INDEX = 'tasks_production';
 # End-of-process marker, depending on mode.
 $EOP_DEFAULTS = Array("stream" => chr(0), "file" => "");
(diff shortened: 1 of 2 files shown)


mgolosova:
[Расширение `data4es` для данных по прогрессу обработки событий](https://trello.com/c/Hk5rXhfc/178-%D1%80%D0%B0%D1%81%D1%88%D0%B8%D1%80%D0%B5%D0%BD%D0%B8%D0%B5-data4es-%D0%B4%D0%BB%D1%8F-%D0%B4%D0%B0%D0%BD%D0%BD%D1%8B%D1%85-%D0%BF%D0%BE-%D0%BF%D1%80%D0%BE%D0%B3%D1%80%D0%B5%D1%81%D1%81%D1%83-%D0%BE%D0%B1%D1%80%D0%B0%D0%B1%D0%BE%D1%82%D0%BA%D0%B8-%D1%81%D0%BE%D0%B1%D1%8B%D1%82%D0%B8%D0%B9)

PR #367, opened 2020-06-15
DF/skel: update stage example.

mgolosova:
Usage of 'traceback' and exceptions handling is no lonnger needed since it is taken care of in the pyDKB library.

### DIFF
--- a/Utils/Dataflow/skel/stage.py
+++ b/Utils/Dataflow/skel/stage.py
@@ -26,7 +26,16 @@ except Exception, err:
 
 
 def process(stage, message):
-    """ Single message processing. """
+    """ Single message processing.
+
+    :param stage: ETL processing stage
+    :type stage: pyDKB.dataflow.stage.ProcessorStage
+    :param message: input message with data to be processed
+    :type message: pyDKB.dataflow.communication.messages.JSONMessage
+
+    :returns: True or False in case of failure
+    :rtype: bool
+    """
     data = message.content()
     # Processing machinery
     out_data = {"key": "value"}
--- a/Utils/Dataflow/skel/stage.py
+++ b/Utils/Dataflow/skel/stage.py
@@ -10,7 +10,6 @@ Authors:
 
 import os
 import sys
-import traceback
 
 try:
     base_dir = os.path.dirname(__file__)
@@ -44,25 +43,12 @@ def main(args):
 
     stage.process = process
 
-    exit_code = 0
-    exc_info = None
-    try:
-        stage.configure(args)
-        stage.run()
-    except (DataflowException, RuntimeError), err:
-        if str(err):
-            sys.stderr.write("(ERROR) %s\n" % err)
-        exit_code = 2
-    except Exception:
-        exc_info = sys.exc_info()
-        exit_code = 3
-    finally:
-        stage.stop()
+    stage.configure(args)
 
-    if exc_info:
-        trace = traceback.format_exception(*exc_info)
-        for line in trace:
-            sys.stderr.write("(ERROR) %s" % line)
+    exit_code = stage.run()
+
+    if exit_code == 0:
+        stage.stop()
 
     exit(exit_code)
 


mgolosova:
[Обновить и доработать skel](https://trello.com/c/P6tpo0cp/180-%D0%BE%D0%B1%D0%BD%D0%BE%D0%B2%D0%B8%D1%82%D1%8C-%D0%B8-%D0%B4%D0%BE%D1%80%D0%B0%D0%B1%D0%BE%D1%82%D0%B0%D1%82%D1%8C-skel)

Utils/Dataflow/skel/stage.py
```diff
@@ -27,7 +26,16 @@
 
 
 def process(stage, message):
-    """ Single message processing. """
+    """ Single message processing.
+
+    :param stage: ETL processing stage
+    :type stage: pyDKB.dataflow.stage.ProcessorStage
+    :param message: input message with data to be processed
+    :type message: pyDKB.dataflow.communication.messages.JSONMessage
```
Evildoor, line 34: Not `AbstractMessage`?

Utils/Dataflow/skel/stage.py
```diff
@@ -27,7 +26,16 @@
...
+
+    :returns: True or False in case of failure
```
Evildoor, line 36: This sounds like "If there is a failure, returns True or False, I don't really know.". Maybe a comma after `True`, or something else?

Utils/Dataflow/skel/stage.py
mgolosova: No. It is an example of a *real* use-case funcion docs, and in the real use-case one should better specify the exact type of message his (her) function accepts. `AbstractMessage` should be in the placeholder method of `ProcessorStage` (and if it is not there, that's a mistake).

Utils/Dataflow/skel/stage.py
mgolosova: It's not my invention: http://doc.docs.sk/php/rararchive.setallowbroken.html
```
Return Values
===

Returns TRUE or FALSE on failure. It will only fail
if the file has already been closed.
```

Maybe not over-standard, but still looks fine to me.

---

We're talking about return values, so any "OR" I read as a separator between different cases. I mean, it makes no sense to read it as "(T or F) in case..." in connection with both return values:
* the return type is `bool`, so it's pretty obvious that the return value is True or False -- so why specify explicitly?
* if "(T or F) in case...", then what does it return in other cases?

By default, I tend to believe that documentation is correct and complete; if so, "T or F in case..." must cover all cases; and if some case is explicitly named, it must be related to a specific return value. Not to all of them.

It's subjective, of course, so not exactly conclusive, but... that's how I read docs, and I don't think I'm somehow unique in this respect.

Utils/Dataflow/skel/stage.py
Evildoor: Aha, I see.

> `AbstractMessage` should be in the placeholder method of `ProcessorStage` (and if it is not there, that's a mistake).

It's not there because there is no type specified at all - placeholder method's documentation is not completed, but that's an issue for some other PR. `skip_process()` correctly has `AbstractMessage`.

Utils/Dataflow/skel/stage.py
Evildoor: > I mean, it makes no sense to read it as "(T or F) in case..." in connection with both return values:
> 
> * the return type is `bool`, so it's pretty obvious that the return value is True or False -- so why specify explicitly?
> * if "(T or F) in case...", then what does it return in other cases?
> 
> By default, I tend to believe that documentation is correct and complete; if so, "T or F in case..." must cover all cases; and if some case is explicitly named, it must be related to a specific return value. Not to all of them.

Sure, one can (and should) apply this logic when reading, and I did it as well while reviewing, just thought that maybe we should make it a bit more obvious. But I'm not insisting, so let's leave it be.

PR #176, opened 2018-09-18
Merge REST API server branch to master

mgolosova:
This PR originally was meant to introduce some basic functionality of the API server, but now this branch (`api`) is considered to be an independent branch, where th server application is being developed.

It was never merged to `master` for two reasons:
* the development was never finished -- there's always something to do;
* there was no need to merge (no functionality from `api` is needed in `master`).

But these two branches share one thing: ES mapping. Some changes were made to it in the `api` and never get to `master` -- and it was never a problem... until today. Now I need to make some changes to the mapping in `master` that requires usage of our custom analyzer, which is `task_dataset_anamyzer` in `master` and `dsname` in `api`).

So I suggest to merge this PR now, and open a new one whenever `api` changes mapping:
* now -- to avoid conflicts in analyzers in the future (for sooner or later the mapping will be merged);
* now and later -- to have the most actual version of mapping in `master` [1].

[1] For it is more natural to take it from here when we need to load new version to ES instance, and I personally was quite surprised to realize that the mapping currently used by ES instance is not the one provided by `master`...

---

**Old description**

In this PR only the basic functionality of the server is to be added:
 - [x] skeleton of the functions and modules layout (`NotImplemented` placeholders are good enough)
 - [x] query string parsing
 - [x] error handling
 - [ ] ~default method (server info)~ *(one day it will happen, but not now)*

PR #368, opened 2020-06-16
DF/data4es: add parent task metadata (p_taskname, p_total_events).

mgolosova:
~Converted to DRAFT until #176 is finished.~ *(done)*

These values are required for proper calculation of the tasks's input events number. It is not really necessary to store these metadata in tES, but since we already have them and no one knows for sure that they won't ever be needed -- let's pass them to ES instead of filtering out.

Also, having these data in the ES will give more information for the events numbers calculation debugging.

ToDo:
 - [x] rebase on `master` after #176 and update mapping (obsolete analyzer is used).

---

NOTE: seems like data samples for output datasets processing (stages 93, 95) were not properly updated after PR #201 (559b9f1) -- when for the first time some tasks like 14259240 (previous status: "done", previous timestamp: "15-06-2018 12:13:57") appeared to change (new status:
"obsolete", new timestamp: "10-04-2019 08:53:10"). Due to this in output samples of stages 93, 95 we have 115 lines removed and only 67 added.
Some (in fact, all) changes in these samples are not related to this specific change, but it looks more or less logical to update samples whenever the dataflow process is changed, so... let's make things consistent.


mgolosova:
[Расширение `data4es` для данных по прогрессу обработки событий](https://trello.com/c/Hk5rXhfc/178-%D1%80%D0%B0%D1%81%D1%88%D0%B8%D1%80%D0%B5%D0%BD%D0%B8%D0%B5-data4es-%D0%B4%D0%BB%D1%8F-%D0%B4%D0%B0%D0%BD%D0%BD%D1%8B%D1%85-%D0%BF%D0%BE-%D0%BF%D1%80%D0%BE%D0%B3%D1%80%D0%B5%D1%81%D1%81%D1%83-%D0%BE%D0%B1%D1%80%D0%B0%D0%B1%D0%BE%D1%82%D0%BA%D0%B8-%D1%81%D0%BE%D0%B1%D1%8B%D1%82%D0%B8%D0%B9)

mgolosova:
> mgolosova [force-pushed](https://github.com/PanDAWMS/dkb/compare/5127e8ef97e0c013b19b1df373bef88067eba2c9..100bd77961c75cc47e2489033a2f61024478320f) the data4es-task-parent-info branch from 5127e8e to 100bd77

Rebased the PR branch on updated master (`parent_taskname` analyzer and sub-fields changed).

Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
```diff
@@ -170,23 +175,41 @@ def transform_chain_data(data):
...
+                         ' "p_tid:p_name:p_total_events"): %s (tid: %s).\n'
+                         % (chain_items[-2], data.get('taskid')))
+    try:
+        chain_data = [int(i) for i in chain_items]
+        data['chain_id'] = chain_data[0]
+        data['chain_data'] = chain_data
+    except ValueError, err:
+        sys.stderr.write('(WARN) Invalid chain_data item: %s (tid: %s).\n'
+                         % (err, data.get('taskid')))
+        return False
```
Evildoor, line 212: Before this PR, correct `taskid` + invalid `chain_data` caused the task to be defined as a root of its own chain.

```
>>> d = {"chain_data":"1,2B,3", "taskid":10}
>>> adjustMetadata.transform_chain_data(d)
(WARN) Task 10: cannot transform chain_data "1,2B,3", it seems to be incorrect. Setting chain_id=10, chain_data=[10].
False
>>> d
{'chain_data': [10], 'chain_id': 10, 'taskid': 10}
```

Now, since the check in former line 173 (`if not chain_data or ...`) was changed, the same situation will be handled in this exception, and `chain_id` and `chain_data` fields will not be set/updated at all.

```
>>> d = {"chain_data":"1,2B,3:NAME:55,4", "taskid":10}
>>> adjustMetadata.transform_chain_data(d)
(WARN) Invalid chain_data item: invalid literal for int() with base 10: '2B' (tid: 10).
False
>>> d
{'parent_taskname': 'NAME', 'chain_data': '1,2B,3:NAME:55,4', 'parent_total_events': '55', 'taskid': 10}
```

P.S. Yes, I know that `taskid` of `10` makes no sense since it's not present in the `chain_data`.

Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
```diff
@@ -170,23 +175,41 @@ def transform_chain_data(data):
...
-                             'received data with incorrect chain_data and '
+                             'received data with empty chain_data and '
                              'without taskid: %s. Skipping.\n' % str(data))
             return False
-        sys.stderr.write('(WARN) Task %s: cannot transform chain_data "%s", '
-                         'it seems to be incorrect. Setting chain_id=%s, '
-                         'chain_data=[%s].\n'
+        sys.stderr.write('(WARN) Task %s: chain_data field is empty.'
+                         'Setting chain_id=%s, chain_data=[%s].\n'
                          % (taskid, chain_data, taskid, taskid))
```
Evildoor, line 187: Too much arguments for formatting: `chain_data` (its `%s`, to be precise) was removed from text, but not from `% (...)`.

Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
mgolosova: Right, the behaviour has changed -- and most likely ES will be surprised by the string `chain_data`, so there're at least two reasons to do something about it.

But I have a question. I am not sure that it's *correct* to set `chain_data` and `chain_id` to `taskid` when original `chain_data` is so wierd -- I'd rather leave them empty. What do you think? Of course, in this exact PR this change is out of place -- it was unintended and I won't even try to keep it here -- but it may be forgotten for good or just moved to a separate PR, so I'd like to know your opinion.

Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
mgolosova: My bad, will be fixed.

Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
mgolosova: And, talking about it: in original logic in case of missed `taskid` the function would still leave `chain_data` as is:
```
~$ echo '{"chain_data":"1,2B,3", "taskid":null}' | ./adjustMetadata.py -m s -E ''
<...>
2020-06-24 15:29:08 (INFO) (ProcessorStage) Starting stage execution.
(WARN) Function transform_chain_data() received data with incorrect chain_data and without taskid: {u'chain_data': u'1,2B,3', 'phys_category': ['Uncategorized'], u'taskid': None}. Skipping.
{"chain_data": "1,2B,3", "taskid": null, "phys_category": ["Uncategorized"]}
2020-06-24 15:29:08 (INFO) (ProcessorStage) Stopping stage.
```
-- which is also a questionable behaviour; I mean, the stage is supposed to set `chain_data` to an expected format -- list of task IDs -- which it does not; so any post-processing relying on this will fail. Setting it to `null` would look fairer to me.
*(Also, out of this PR, but since we're here...)*

Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
mgolosova: Fixed: 147a34e

Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
mgolosova: Original logic restored: 0d9c09a
```
(dkb-dev) [dkb@aiatlas171 017_adjustMetadata]$ echo '{"chain_data":"1,2B,3:NAME:55,4", "taskid":10}' | ./adjustMetadata.py -m s -E ''
<...>
2020-06-24 15:40:09 (INFO) (ProcessorStage) Starting stage execution.
(WARN) Invalid chain_data item: invalid literal for int() with base 10: '2B' (tid: 10).
(INFO) Setting chain_id=10, chain_data=[10] (tid: 10).
{"parent_total_events": "55", "chain_data": [10], "chain_id": 10, "phys_category": ["Uncategorized"], "taskid": 10, "parent_taskname": "NAME"}
2020-06-24 15:40:09 (INFO) (ProcessorStage) Stopping stage.
```

Evildoor: > But I have a question. I am not sure that it's _correct_ to set `chain_data` and `chain_id` to `taskid` when original `chain_data` is so wierd -- I'd rather leave them empty. What do you think? Of course, in this exact PR this change is out of place -- it was unintended and I won't even try to keep it here -- but it may be forgotten for good or just moved to a separate PR, so I'd like to know your opinion.

My opinion is as following: you are asking reasonable questions, both here and in other parts of your replies, and I had similar thoughts while reviewing this code. Since possible changes are outside of this PR's code, I made a [card](https://trello.com/c/rLxtg5Mn) for discussing the issues - will write some more thoughts into it later.

As for this point here, "things are now as they were before", so we are concluding it.

Utils/Dataflow/017_adjustMetadata/adjustMetadata.py
```diff
@@ -154,9 +154,14 @@ def input_events(data):
 
 
 def transform_chain_data(data):
-    """ Transform chain_data into array of integers and get chain_id from it.
+    """ Transform chain_data into proper task metadata fields.
 
-    chain_id is the taskid of the task chain's root.
+    New/updated fields:
+      - chain_data:          task IDs of all previous tasks in chain
```
Evildoor, line 160: Sorry, got too busy with actual problems with logic and forgot to bark at trivial things: `chain_data` includes the current task. I'm not sure that this is the right place for explaining what exactly a task chain is, we have READMEs and 009 for that, so I think dropping `previous` would be enough.

mgolosova: Fixed: 6df3e94
I couldn't say "all tasks", for "all" may include tasks _after_ the current one -- so it would be not less confusing.

PR #377, opened 2020-07-20
DF/data4es: fix issue with invalid ESC characters in JSON.

mgolosova:
If JSON contains symbols like '*', in `echo $line` they will be interpreted by Shell and turned into (e.g.) a list of local files and subdirectories. Which may contain invalid (from the JSON syntax point of view) characters/ESC sequences, like '\ '.

---

The issue had effect, for example, for tasks from production request 31979 (e.g. taskid: 21835493, 21834893).

### DIFF
--- a/Utils/Dataflow/069_upload2es/load_data.sh
+++ b/Utils/Dataflow/069_upload2es/load_data.sh
@@ -18,7 +18,7 @@ ES_CONFIG="${base_dir}/../../Elasticsearch/config/es"
 
 verify_ndjson() {
   while read -r json_line; do
-    err=`echo $json_line | jq "." 2>&1 >/dev/null`
+    err=`echo "$json_line" | jq "." 2>&1 >/dev/null`
     [ -n "$err" ] && log WARN "Failed to parse input as JSON ($err): $json_line"
   done
 }
--- a/Utils/Dataflow/run/data4es-start
+++ b/Utils/Dataflow/run/data4es-start
@@ -158,7 +158,7 @@ mediator() {
   buffer=`get_buffer "$1"`
   i=0
   while read -r line; do
-    echo $line >> $buffer
+    echo "$line" >> $buffer
     let i=$i+1
     let f=$i%$BATCH_SIZE
     [ $f -eq 0 ] && flush_buffer "$1"


witxka:
Actually it is more complicated to use double or single  quotes in scripts with echo. Looks that in our case double quotes more useful than single. 
Here is the nice info https://stackoverflow.com/questions/6697753/difference-between-single-and-double-quotes-in-bash

mgolosova:
@witxka, 
> Actually it is more complicated to use double or single quotes in scripts with echo. Looks that in our case double quotes more useful than single.
> Here is the nice info https://stackoverflow.com/questions/6697753/difference-between-single-and-double-quotes-in-bash

Not sure what you meant... single quotes are not applicable here, while double quotes seem to do what's needed:
```
[dkb@aiatlas171 ~]$ line='{"key_1": "val_1", "key_2": 123, "key_3": "string `with` '"'single quotes'"' and * $(just in case)", "key_4": ["array", "of", "values"]}'
[dkb@aiatlas171 ~]$ echo $line
{"key_1": "val_1", "key_2": 123, "key_3": "string `with` 'single quotes' and 009.cons.out 016.cons.out 071.cons.out 16.out 17.out 2 25.out 69.b16.inp 69.b91.inp 91_in.out cron.20200702 crt.pem data4es-analysis.log data4es.analysis.tmp.log data4es.analysis.tmp.log.debug data4es.analysis.tmp.log.failed data4es.cfg.analysis.tmp data4es.home.analysis.tmp data4es-hourly.log data4es-start-tmp data4es-state-analysis.log data4es-state-hourly.log Dataflow dkb-dev.git dkb-dev.git.config dkb.git dkb.git.config dkb.p12 dkb-test-20200702.git dkb-test-20200721.git dkb-test.git dkb-test.git.config dkb-tmp.config etc hist20200211.txt key.pem oradiag_dkb perf_test test_prodsys_rucio_ami_datasets.json TO_DELETE upgrade vcsrepo-test.git $(just in case)", "key_4": ["array", "of", "values"]}
[dkb@aiatlas171 ~]$ echo '$line'
$line
[dkb@aiatlas171 ~]$ echo "$line"
{"key_1": "val_1", "key_2": 123, "key_3": "string `with` 'single quotes' and * $(just in case)", "key_4": ["array", "of", "values"]}
```

Did I miss something? I am not very good with "ANSI-C quoting" (like `$'this'`), history expansion (like `"!this"`), etc., so if double quoting is not safe enough and it can be done better -- please let me know where to look for details.

witxka:
Looks like you are right in the snippet _while read -r line; do echo $line;done_ there is only option to use double quotes and some hacks like _echo "`cat /etc/passwd`"_ does not work.

Issue #383, opened 2020-08-14
Add automatic docs build (PDF)


mgolosova:
It would be great to get rid of re-building docs every now and again by hands (and still have out-of-date docs).

We could do something like this:
- ~someone decides to update documentation and does so in a branch named `...-docs`~ *(called off after discussion)*;
- ~for this branch automaticaly created branch `...-docs-built` (maybe anly when it becomes a PR?), to which after each push to the original branch the built PDF version of the updated documentation is commited~ *(called off after discussion)*;
- when the PR is committed to `master`, to some special branch (e.g. `pdf-docs`) should be committed the new version of the PDF document;
- **Q**: do the same when *any* PR is commited (for it *may* add / change docstrings just as a part of development process)?

As for the HTML version -- I am not sure if we really need it (unless we know how to publish it somewhere) so let's forget about it for now.

How it works now (see #158):
* we annotate code with Sphinx-compatible docstrings;
* directory `/Docs/source` contains docs structure;
* `/Docs/Makefile` contains instructions on how to build the docs (`make pdf` will rebuild the PDF file and place it to `/Docs/build/pdf/DKB.pdf`).

Building the docs requires:
* Sphinx (`pip install sphinx`);
* pdflatex;
* (maybe) tons of TeX stuff...

IATkachenko:
Points 1 and 2 is require a lot of actions from contributors and not looks like what you want: regular code updates will require additional -docs branch plus will be created -docs-built branch. So contributor should manage three branches instead of one.
We may regenerate documentation after every push to every branch, but it may create a lot of commits "documentation update" in PR and master. And it is not good solution too.

For point 3 check my PR. Workflow will create PR for documentation update (if needed) after every push to master. If PR was not merged, then it will be updated.

mgolosova:
@IATkachenko, OK, I guess you're right about "too many actions" -- it's better to rebuild the PDF locally when it's really needed than be annoyed by some unintuitive logic. Didn't think of it this way but should have :)

> For point 3 check my PR. Workflow will create PR for documentation update (if needed) after every push to master. If PR was not merged, then it will be updated.

Thank you, looks great!

PR #386, opened 2020-08-14
Use github action for automatically generate documentation

IATkachenko:
After every push to master this workflow will create PR for
documentation update.

Fixes part 3 of #383

### DIFF
--- a/.github/workflows/doc_update_on_push2master.yml
+++ b/.github/workflows/doc_update_on_push2master.yml
@@ -33,9 +33,7 @@ jobs:
 
     - name: make
       working-directory: ./Docs
-      run: |
-        make clean
-        make pdf
+      run: make pdf
 
     #https://github.com/marketplace/actions/create-pull-request
     - name: Create Pull Request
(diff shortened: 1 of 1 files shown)


.github/workflows/doc_update_on_push2master.yml
```diff
@@ -0,0 +1,48 @@
...
+
+    - name: Install pip packages
+      uses: BSFishy/pip-action@v1
+      with:
+        packages: sphinx
+
+    - name: make
+      working-directory: ./Docs
+      run: |
+        make clean
```
mgolosova, line 37: Since every time it runs in a newly built env, `clean` is unnecessary step.

```suggestion
```

.github/workflows/doc_update_on_push2master.yml
```diff
@@ -0,0 +1,48 @@
...
+        make pdf
```
IATkachenko, line 38:
```suggestion
      run: make pdf
```
should be better, because we need just one action => we don't need multiline command

PR #396, opened 2020-08-25
DF/data4es-nested: rename 'conditions_tags' to 'conditions_tag'.

mgolosova:
Original parameter in JEDI is called 'conditionsTag'.

This change requires full data reindexing, which can only be done at the point of switching to the new storage scheme.


mgolosova:
[ES mapping: rename '`conditions_tag[s]`'](https://trello.com/c/7jM2SPNU/155-es-mapping-rename-conditionstags)

PR #400, opened 2020-08-26
DF/data4es-nested: remove 'output' field.

mgolosova:
This field is used during metadata processing, but is not supposed to be stored in ES documents (especially when output datasets are stored as nested fields and are available from the original document).

---

Closes https://trello.com/c/72OzDlGB


mgolosova:
[Поле `output` в ES](https://trello.com/c/72OzDlGB/124-%D0%BF%D0%BE%D0%BB%D0%B5-output-%D0%B2-es)

PR #402, opened 2020-08-26
ES/mapping/nested: change output dataset `data_format` mapping.

mgolosova:
Previously the field values were splitted by '_' prior to data load, and in ES document the field might have two values: "DAOD_EXOT13" and "DAOD", indexed as keyword values. As a result we could look for datasets by "DAOD" keyword -- but in `terms` aggregation had extra bucket "DAOD", which makes no sence from the users' point of view.

Now the field is stored as text field, analyzed with dsname analyzer (which splits it by '_'); and the original value is stored as 'data_format.keyword', indexed as keyword. With this mapping one can still have `terms` aggregation by formats (via 'data_format.keyword'), and can query for `"term": {"data_format": "daod"}`. The only issue is that queries like `"wildcard": {"data_format": "daod_*"}` are doomed; but `"daod*"` is still fine to use.


mgolosova:
[ES mapping: поле `data_format`](https://trello.com/c/ZX9H2mkv/156-es-mapping-%D0%BF%D0%BE%D0%BB%D0%B5-dataformat)

PR #409, opened 2020-09-17
DF/data4es-nested: port `data4es` DF changes made in PR #284.

mgolosova:
095 update: part 2.

Query improvements:
* use `-scope` parameter to get data for datasets in non-default scope;
* do not query AMI for datasets beyond AMI's scopes ('mc15', 'mc16');
* use normalized DS name in the query.


mgolosova:
@Evildoor,
As it was promised, your PR #284 is ported to the "nested" version of the dataflow (for the PR had been created before the "nested" DF landed to `master`).
From now on, please remember that all the improvements are to be done in this "nested" DF, not the original `data4es` (except for hotfixes, which are to be made in both DFs).

PR #420, opened 2020-10-02
DF/data4es-nested/091: fix 'skip' scenario behaviour.

mgolosova:
Default behaviour: pass message as is, marking it 'incomplete'.

New behaviour: replace field `output` (list of output datasets) with `output_datasets` (list of nested documents representing datasets).


mgolosova:
[DF/nested: 091 skip scenario](https://trello.com/c/JzZFHM2p/191-df-nested-091-skip-scenario)

Issue #446, opened 2023-06-02
DKB migration to python3


witxka:
DKB needs to be migrated to python3

witxka:
Tested on aiatlas174 AlmaLinux9 Python 3.9.16 with Atlas production workflow

PR #447, opened 2023-06-02
Refs: #446: DKB migration to python3

witxka:
Migration DKB to python3.

DKB run in Conda python environment independent on OS used.

Initial migration performed by running 2to3 python helper program:
  find  /opt/inMem/dkb/ -type f -name *.py   -exec 2to3 -n -w -f all -f
  idioms -f buffer -f set_literal -f ws_comma {} \;

DBK specific fixes for python3:
  * all strings are unicode
  * no more unbuffered io for strings
  * sorted dictionary in mem
  * no more encoded strings by default
  * no more decode 'string_escape'
  * changes in error output for tests
  * changes in Elasticsearch python3 library
  * add conda python environment yaml


Utils/Dataflow/test/pyDKB/test.sh
```diff
@@ -47,22 +49,27 @@ test_case() {
...
   eval "$before $cmd; $after"  2>&1 1> out.tmp | \
-    grep -a -v '(WARN) (pyDKB.dataflow.cds) Submodule failed (No module named invenio_client.contrib)' | \
+    grep -a -v "(WARN) (pyDKB.dataflow.cds) Submodule failed (No module named 'invenio_client')" | \
     sed -E -e"s#$base_dir#\$base_dir#" \
+           -e"s#$base_dir_mod##" \
+           -e's#\$base_dir/./../../##g' \
            -e"s#^[0-9]{4}-[0-9]{2}-[0-9]{2} [0-9]{2}:[0-9]{2}:[0-9]{2} ##" \
-           -e"s#, line [0-9]+#, line <NNN>#" > err.tmp
+           -e"s#, line [0-9]+#, line <NNN>#" \
+           -e"s#/.*/os.py#os.py#" | sort > err.tmp
```
mgolosova, line 65: Why the unsorted output no longer works?

mgolosova (commented):
- in a couple of places original bug became obvious so it'd be nice to fix,
- in some other places now there's too many checks against string and string (after turning `unicode` into `str`); nothing wrong, but mb simplify?
- and, a couple of questions about changes happened to test utility.

Utils/Dataflow/test/pyDKB/test.sh
witxka: As soon as there are changes in exception handling and print order is changed it was decided to use already prepared tests in python2 for python3. By using this technique  we avoided writing new tests for a python3 code.

Utils/Dataflow/pyDKB/dataflow/stage/ProcessorStage.py
```diff
@@ -249,18 +249,19 @@ def run(self):
...
                     self.flush_buffer()
                 else:
                     self.clear_buffer()
                 self.forward()
-        except BaseException, err:
+        except BaseException as err:
             # Catch everything for uniform exception handling
             # Clear buffer -- just in case someone will decide
             # to reuse the object.
+            process_err = err
```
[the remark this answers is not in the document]
witxka: Python3 more strict checks defined variables so process_err should be initialized before usage in exception blocks.

